PR #427, opened 2015-06-17
Update README.md

reduxionist:
Correct path in installation instructions since project repository renamed to srs from simple-rtmp-server

It's easy enough for everyone to see on their own, but for ease of copy-and-paste installation, it's still worth updating.


01118976:
16.16.63.0

PR #499, opened 2015-10-10
fix bug for build failure when configure with --log-info

synote:
wrong use of size in SrsStream::initialize of  srs_kernel_stream.cpp


winlinvip:
Good job~

Issue #503, opened 2015-10-16
source 对 forward/hls/hds/dvr/trancode  reload的处理有问题


zhengfl:
在对forward/hls/hds/dvr/trancode做reload操作时，需要判断源流是否在推（origin），或者拉（edge）。若源流已断开，不需要做reload操作

winlinvip:
Assignee to Li Qiang, Feilong will make the changes whenever he is free. If it is not changed before the release, I will make the changes.

`TRANS_BY_GPT3`

zhengfl:
Reproduction steps:
Streaming (forward/hls/dvr... start) --> Stop streaming (forward/hls/dvr... stop) --> Modify configuration (forward/hls/dvr...) --> Reload (forward/hls/dvr started again)

`TRANS_BY_GPT3`

winlinvip:
Replay is succeed.

PR #505, opened 2015-10-19
do not reload forward/hls/hds/dvr/trancode when stream unpublish #503

tufang14:
#503


winlinvip:
@zhengfl Feilong, take a look at this modification and see if it works.

`TRANS_BY_GPT3`

lemonbinary:
This one is closed for now, we will test overload submission properly another day.

`TRANS_BY_GPT3`

winlinvip:
:+1:

Issue #548, opened 2015-12-24
No http api token authentication support for http-api


FlowerWrong:
There is no token authentication support for http-api module.
I have found the `srs/trunk/src/app/srs_app_http_api.cpp`, but no code for this.

winlinvip:
Not support yet.

FlowerWrong:
I also found that there is only [security](https://github.com/ossrs/srs/wiki/v2_EN_Security) config for user who can push rtmp flow to srs server, but the ip of push flow client may change often. Can you support it on `http-callback`. For example, you can support it with `on_publish` event. The `on_connect` event can not do that.

winlinvip:
SRS does support `on_publish` and `on_connect` callback. The `security` is another solution, for example, some edge only allow play.

FlowerWrong:
Hi, I found that `http callback api` has a `"tcUrl": "rtmp://x/x?key=xxx"` in `on_connect`. But there is no one in `on_publish`, I want to deny someone to publish rtmp stream to srs server without username and password or named token.

winlinvip:
OK, Accept.

`TRANS_BY_GPT3`

PR #558, opened 2016-01-08
Add tcUrl to on_publish http hook for issue #548

FlowerWrong:
Only add one line, see [diff](https://github.com/FlowerWrong/srs/commit/9ddd0054a17b4dbee76294387b91bf7342530b80#diff-74b96fe66df78acf666fd22f3d16604aR141).

### DIFF
--- a/trunk/src/app/srs_app_http_hooks.cpp
+++ b/trunk/src/app/srs_app_http_hooks.cpp
@@ -60,9 +60,9 @@ SrsHttpHooks::~SrsHttpHooks()
 int SrsHttpHooks::on_connect(string url, SrsRequest* req)
 {
     int ret = ERROR_SUCCESS;
-
+    
     int client_id = _srs_context->get_id();
-
+    
     std::stringstream ss;
     ss << SRS_JOBJECT_START
         << SRS_JFIELD_STR("action", "on_connect") << SRS_JFIELD_CONT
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_http_hooks.cpp
```diff
@@ -60,9 +60,9 @@ SrsHttpHooks::~SrsHttpHooks()
 int SrsHttpHooks::on_connect(string url, SrsRequest* req)
 {
     int ret = ERROR_SUCCESS;
-    
+
```
winlinvip: Can you please revert these black lines?

FlowerWrong:
I have revert it, but I think that you should remove all blank in empty line.

winlinvip:
It make sense, we should remove the blanks in line, but it will cause the branch merge failed. So, I think let we ignore it.

winlinvip:
for #548

winlinvip:
Merged to develop: https://github.com/ossrs/srs/commit/eac54405593595dd559c887043d8b0f92290ac54

FlowerWrong:
Thanks for merge. I have edited the [wiki](https://github.com/ossrs/srs/wiki/v2_CN_HTTPCallback#http-callback-events).

winlinvip:
:+1

Issue #553, opened 2015-12-29
ingest内存泄露的问题


zdco:
使用的是develop最新分支代码，srs配置文件开启100个ingest实例，采集doc中的source.200kbps.768x320.flv文件，配置文件内容如下：
```
listen 1935;
pid ./objs/srs.pid;
chunk_size 60000;
ff_log_dir /dev/null;
srs_log_tank file;
srs_log_level trace;
srs_log_file ./objs/srs.log;
max_connections 1000;
daemon off;
utc_time off;
heartbeat {
    enabled on;
    interval 9.3;
    url http://192.168.1.56:8080/api/v1/servers;
    device_id 001;
    summaries on;
}
stats {
    network 0;
    disk sda sdb xvda xvdb;
}
http_api {
    enabled on;
    listen 1985;
    crossdomain on;
    raw_api {
        enabled on;
        allow_reload on;
        allow_query on;
        allow_update on;
    }
}
http_server {
    enabled on;
    listen 9090;
    dir ./objs/nginx/html;
}
vhost __defaultVhost__ {
    min_latency on;
    tcp_nodelay on;
    play {
        gop_cache off;
        queue_length 10;
        mw_latency 100;
    }
    publish {
        mr off;
    }
    hls {
        enabled on;
        hls_path ./objs/nginx/html;
        hls_fragment 4;
        hls_window 8;
        hls_on_error ignore;
        hls_storage ram;
        hls_mount [vhost]/[app]/[stream].m3u8;
    }
}
vhost ingest {
    ingest 1 {
        enabled on;
        input {
            type file;
            url ./doc/source.200kbps.768x320.flv;
        }
        ffmpeg ./objs/ffmpeg/bin/ffmpeg;
        engine {
            enabled on;
            vcodec copy;
            acodec an;
            output rtmp://127.0.0.1:[port]/live/1;
        }
    }
    ingest 2 {
        enabled on;
        input {
            type file;
            url ./doc/source.200kbps.768x320.flv;
        }
        ffmpeg ./objs/ffmpeg/bin/ffmpeg;
        engine {
            enabled on;
            vcodec copy;
            acodec an;
            output rtmp://127.0.0.1:[port]/live/2;
        }
    }
    ingest 3 {
        enabled on;
        input {
            type file;
            url ./doc/source.200kbps.768x320.flv;
        }
        ffmpeg ./objs/ffmpeg/bin/ffmpeg;
        engine {
            enabled on;
            vcodec copy;
            acodec an;
[...]

PR #559, opened 2016-01-09
srs reusable thread need join

tufang14:
我详细描述一下问题
在SrsThread::stop()函数中是先loop = false, 然后再执行SrsThread::dispose()函数，
在SrsThread::thread_cycle()中退出loop，变量disposed = true，
这样的话，就有可能在执行SrsThread::dispose()之前变量disposed就赋值为true，导致SrsThread::dispose()函数直接return了

还有个问题，我就注释了一行代码，上面是issue546#546的描述，我觉得#546的出现是因为在handler->on_thread_stop()只后执行了disposed的赋值，因为对于SrsConnection来说在handler->on_thread_stop()中是删除自己的操作，也就是free当前的thread，所以任何在handler->on_thread_stop()之后的任何操作都会出错

### DIFF
--- a/trunk/src/app/srs_app_thread.cpp
+++ b/trunk/src/app/srs_app_thread.cpp
@@ -235,7 +235,7 @@ namespace internal {
         // when thread terminated normally, also disposed.
         // we must set to disposed before the on_thread_stop, which may free the thread.
         // @see https://github.com/ossrs/srs/issues/546
-        disposed = true;
+        // disposed = true;
         
         handler->on_thread_stop();
         srs_info("thread %s cycle finished", _name);
--- a/trunk/src/app/srs_app_thread.cpp
+++ b/trunk/src/app/srs_app_thread.cpp
@@ -232,11 +232,6 @@ namespace internal {
         // readly terminated now.
         really_terminated = true;
         
-        // when thread terminated normally, also disposed.
-        // we must set to disposed before the on_thread_stop, which may free the thread.
-        // @see https://github.com/ossrs/srs/issues/546
-        // disposed = true;
-        
         handler->on_thread_stop();
         srs_info("thread %s cycle finished", _name);
     }


winlinvip:
This is because if the thread exits on its own, there is no need to go through the complicated process of disposing, waiting for the thread to exit, and confirming the exit, etc.
The issue you mentioned is indeed a bug. It should be changed as follows:
1. If it is an active exit, that is, calling stop to exit, it should be confirmed that the thread has exited as it is now.
2. If the thread exits on its own, as it is currently, then disposed=true should be set.
Both of these methods should call the third function: st_release, to release resources using join.

`TRANS_BY_GPT3`

winlinvip:
Or in the second case, you can actively call join:


```
        // when joinable, wait util quit.
        if (_joinable) {
            // wait the thread to exit.
            int ret = st_thread_join(tid, NULL);
            if (ret) {
                srs_warn("core: ignore join thread failed.");
            }
        }
```


Then set disposed=true.

`TRANS_BY_GPT3`

lemonbinary:
Can join be called in the current thread? It should be called in another thread, right?

`TRANS_BY_GPT3`

winlinvip:
Makes sense, I think you just removed these few lines. It will run again in the dispose, so it's not a problem.

`TRANS_BY_GPT3`

lemonbinary:
Yeah, I feel the same way. If there's no need to join, just skip it. And the while loop below will also exit directly, so there's no significant consumption.

`TRANS_BY_GPT3`

winlinvip:
:+1: 

Please update the pr.
Good job~

winlinvip:
This pr fixed the #553

winlinvip:
Merged to develop: https://github.com/ossrs/srs/commit/eac54405593595dd559c887043d8b0f92290ac54

PR #609, opened 2016-06-03
新增一些功能并对对部分代码进行了调整

aesirteam:
1. added support -rtsp_transport in srs_app_ffmpeg
2. added raw api ingest (create/update/delete/enable/disable)
3. added token authentication of http hook
4. added localhost token authentication


winlinvip:
Firstly, the changes are too significant; secondly, many unnecessary changes have also been added, such as srs.conf and some qt files.

`TRANS_BY_GPT3`

PR #749, opened 2017-01-18
bugfix atc 开启后，由于客户端重置时间撮，会导致consumer wait condition 长时间不触发.

censhanhe:
bugfix atc 开启后，由于客户端重置时间撮，会导致consumer wait condition 长时间不触发.

### DIFF
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -2202,7 +2202,10 @@ void SrsSource::on_unpublish()
     // donot clear the sequence header, for it maybe not changed,
     // when drop dup sequence header, drop the metadata also.
     gop_cache->clear();
-    
+    srs_freep(cache_metadata);
+    srs_freep(cache_sh_video);
+    srs_freep(cache_sh_audio);    
+
     srs_info("clear cache/metadata when unpublish.");
     srs_trace("cleanup when unpublish");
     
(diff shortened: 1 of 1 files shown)


censhanhe:
The condition for this bug to occur is: ATC is enabled, and the streaming end is disconnected and then reconnected. 
The reason is: ATC is enabled, and the timestamp is reset (meaning the streaming end did not continue from the previous timestamp). Because the SrsSource cache contains some SH packets with the previous streaming timestamp (which is much larger than the newly pushed timestamp), this causes the queue duration calculation to be negative. As a result, the consumer wait condition is not triggered for a long time.

`TRANS_BY_GPT3`

winlinvip:
This is indeed a problem, but the solution is not to directly free SH and metadata. SRS has a function called `reduce_sequence_header`, which means that if the encoder pushes duplicate SH to the server, the server can choose not to send it to the client. This function is designed to accommodate some clients that may crash when receiving duplicate SH [facepalm].

You can think of other ways to fix this problem, such as modifying the timestamp. In any case, SH cannot be freed, it needs to be cached in order to support SH deduplication functionality.

`TRANS_BY_GPT3`

censhanhe:
Hmm, actually there is another similar issue, which is that the timestamp of the client itself is not incremental. In the ATC mode, there may also be a problem where the condition is not triggered. 
In this case, a common solution is to modify the trigger condition of the condition. In ATC mode, if the duration is less than 0, the condition can be triggered immediately.

`TRANS_BY_GPT3`

winlinvip:
Based on your work, I made some slight modifications and updated the version and README.
👍 Thanks~

`TRANS_BY_GPT3`

winlinvip:
The purpose of this ATC mode is to expose the timestamp issue without making any modifications to the timestamp. However, with the changes you made, this goal can still be achieved, and the tolerance is even higher. Great job~ It improves robustness without compromising the original intention of ATC.

`TRANS_BY_GPT3`

PR #763, opened 2017-02-13
Replace ffmpeg version from 2.1.1 to 3.2.4

HungMingWu:
Replace FFmpeg version from 2.1.1 to 3.2.4
Cause:
The old FFMpeg transcode audio from G711 Audio to AAC, it would met memory corruption.


winlinvip:
Thanks~

PR #782, opened 2017-02-24
RTSP存在内存泄漏问题

ljx0305:
存在内存泄漏问题


winlinvip:
The `send_message` function should be modified to release within it. If it is necessary to enter the queue without sending immediately, it should not be released.
```
SrsRtspStack::send_message
```

If you feel that the function name does not reflect the purpose of this function, you can change it. Please refer to:
```
SrsProtocol::send_and_free_message
```

RTMP sends many messages.

`TRANS_BY_GPT3`

winlinvip:
PR that has not been given feedback for a long time, converted to ISSUE: #797

`TRANS_BY_GPT3`

Issue #787, opened 2017-02-28
APPLICATION: Add more parameters on HTTP publish/play callbacks


antonioparisi:
Hi @winlinvip,

thanks for your effort on this project, I think, it's one of the most stable, most performant and most well designed so far out there. Lets keep up the good work! 👍 

So, I raised this issue cause I'm having few bottlenecks on HTTP Callbacks.
I need to get the i.e. `access_token` or `token` parameter passed via the RTMP connection URL.
So far you can only kinda get it through the `on_connect` callback using the `tcUrl` parameter.

But only having that `tcUrl` param on that callback isn't enough for me since I need it also on the `on_publish` and `on_unpublish` callbacks.

So I'll try to raise a PR adding those params on those two callback if you agree. Also I noticed that the `SrsRequest` class has got also a `param` field. Would it make sense just to return that `param` on those callbacks? In theory that should include any parameter passed on the RTMP connection URL. i.e. `rtmp://localhost/app?access_token=foobar&another_key=barfoo`

Regards,
Antonio

winlinvip:
It should work. Please file PR to develop branch, for 2.0release is released, only open to bug fix PR. However, you can merge it to your 2.0release(maybe a new branch is better), and give us your branch link.

winlinvip:
Does that make sense? Let me explain more specific:

1. I think this feature is reasonable.
1. I think it should be file to develop branch, not the 2.0release branch.
1. For guys to need to use this feature on 2.0release, you can maintain your 2.0release or a new branch, which patched your PR to support this feature.

Good idea, regards.👍

antonioparisi:
Thanks @winlinvip - I'll raise a new PR to develop branch :v:

winlinvip:
👍 Great. I usually work for SRS at weekends, so if you need more quick feed back, please contact me by Skype `winterserver`.

winlinvip:
Fixed in SRS3.

PR #788, opened 2017-02-28
Return param key on on_publish and on_unpublish

antonioparisi:
- Add `param` key to the `on_publish` and `on_unpublish` HTTP callbacks
so to manage parameters like authentication token and others easily.

Reference issue: https://github.com/ossrs/srs/issues/787


winlinvip:
Please read https://github.com/ossrs/srs/issues/787#issuecomment-284133014 and file a new PR.

PR #794, opened 2017-03-03
fix bug, 修复UserControlPacket包的event data 数值

SnailTowardThesun:
the event data in User Control Packet whose type is SrsPCUCFmsEvent0(0x1a) should be 1 not 0


winlinvip:
https://github.com/ossrs/srs/pull/794/files#diff-01e14c2adcb813653137c9e61ac884ecL5301
@ObservingCloud-Research and Development-Han Kun
1. The title should not be "fix bug", but should explain what problem was fixed.
2. This doesn't seem like a bug, right? Why is the value 0x01 being written instead of writing event_data for wirte_1bytes(1)? From what I see in the logic, when it is of type SrsPCUCFmsEvent0, it is only 1 byte, while others are 4 bytes. Why does it change to writing the constant 1? Even if writing a constant, it should be done using a macro definition, otherwise, it is unclear what it means.

`TRANS_BY_GPT3`

Issue #634, opened 2016-08-25
api/v1/streams/ 对于已经停止的推流，没有及时清除数据


MaxLoveThree:
测试版本如下：
[root@localhost trunk]# ./objs/srs -V
2.0.209
测试过程：
1， 向srs推送rtmp流rtmp://172.16.198.129:1935/my_test/test_0，然后停止
2，再srs推送rtmp流rtmp://172.16.198.129:1935/my_test/test_1，然后停止
3，调用http://172.16.198.129:1936/api/v1/streams/
{"code":0,"server":5672,"streams":[{"id":5676,"name":"test_0","vhost":5673,"app":"my_test","live_ms":1472126974403,"clients":0,"send_bytes":4524,"recv_bytes":466352,"kbps":{"recv_30s":0,"send_30s":0},"publish":{"active":false,"cid":-1},"video":null,"audio":null},{"id":5677,"name":"test_1","vhost":5673,"app":"my_test","live_ms":1472126974403,"clients":1,"send_bytes":4190,"recv_bytes":3481,"kbps":{"recv_30s":0,"send_30s":0},"publish":{"active":false,"cid":-1},"video":null,"audio":null}]}

测试发现streams数据会一直增加，不会减少
然后跟踪代码，发现存在泄露问题
srs_app_statistic.hpp srs_app_statistic.cpp
SrsStatistic 类的vhosts，rvhosts，streams，rstreams的这四个成员只会不断增加，没有erase，只有在程序结束，析构的时候才会clear

winlinvip:
Not much can be leaked.

`TRANS_BY_GPT3`

hyt-hz:
The latest research is being conducted to use SRS as a replacement for nginx-rtmp. Testing is being done on version 2.0-r0 and the same issue has been discovered.

In our application scenario, we frequently create short-lived streams with randomly generated names. This results in a large number of streams being generated within a short period of time. We would like to use the SRS HTTP API to confirm the status of these streams. However, if we use the api/v1/streams/ interface, it returns a large number of streams that no longer exist, causing unnecessary pressure on both SRS and our server. Our proposed solution is to record the correspondence between the stream names on our server and the stream IDs in SRS, and then use the api/v1/streams/{stream_id} interface to query the status of the streams. Although this method solves our problem, we still hope to see this issue being fixed.

Thank you.


`TRANS_BY_GPT3`

ChandZhang:
[...]

PR #800, opened 2017-03-09
FIX：When stream stopped，Stream API does not remove it。

zachry-zhang:
HTTP API for streams is not correct. When stream stopped, API does not
remove it. 
Issue #634


winlinvip:
I think we should add a delay for cleaning up, instead of immediately clearing it when the streaming stops.

`TRANS_BY_GPT3`

henmaodun:
How to clean the stream of edge servers?

`TRANS_BY_GPT3`

PR #849, opened 2017-04-18
Update srs_core.hpp

betterjun:
workaround for compiling srslibrtmp on any other cpus, eg mips.

### DIFF
--- a/trunk/src/core/srs_core.hpp
+++ b/trunk/src/core/srs_core.hpp
@@ -133,7 +133,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  * @reamrk to patch ST for arm, read https://github.com/ossrs/state-threads/issues/1
  */
 #if !defined(__amd64__) && !defined(__x86_64__) && !defined(__i386__) && !defined(__arm__)
-    #error "only support i386/amd64/x86_64/arm cpu"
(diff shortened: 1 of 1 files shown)


trunk/src/core/srs_core.hpp
```diff
@@ -133,7 +133,12 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  * @reamrk to patch ST for arm, read https://github.com/ossrs/state-threads/issues/1
  */
 #if !defined(__amd64__) && !defined(__x86_64__) && !defined(__i386__) && !defined(__arm__)
-    #error "only support i386/amd64/x86_64/arm cpu"
+    /* for srslibrtmp, it doesn't depend on st, so no limitation here. 
+     * the below macro is defined in the auto generated srslibrtmp header file, it's not a good solution but a workaround.
+     */
+    #ifndef SRS_LIB_RTMP_HPP
```
winlinvip, line 139: This macro definition, you have chosen incorrectly, it will not work, there will still be errors. It should be `--export-librtmp-single` that specifies this compilation option, which will define a macro, rather than the macro definition of librtmp.hpp.

`TRANS_BY_GPT3`

winlinvip: It should be this macro definition: SRS_EXPORT_LIBRTMP.

`TRANS_BY_GPT3`

winlinvip:
Closed for no response.

PR #860, opened 2017-04-25
fixed compile error when add --log-info

walkermi:
fixed compile error when add --log-info


wuxianlijiang:
Happy Labor Day.

`TRANS_BY_GPT3`

PR #876, opened 2017-05-05
Fix the timebase bug, change from us to ms.

xialixin:
修改SRS_CONSTS_RTMP_PULSE_TMMS的定义 从500 改为500*1000，如果是500，在使用http-flv播放时单个cpu会占100%的问题

### DIFF
--- a/trunk/src/kernel/srs_kernel_consts.hpp
+++ b/trunk/src/kernel/srs_kernel_consts.hpp
@@ -77,7 +77,7 @@
 // if timeout, we generally ignore and send the data to client,
 // generally, it's the pulse time for data seding.
 // @remark, recomment to 500ms.
-#define SRS_CONSTS_RTMP_PULSE_TMMS (int64_t)(500*1000LL)
+#define SRS_CONSTS_RTMP_PULSE_TMMS (500)
 
 /**
  * max rtmp header size:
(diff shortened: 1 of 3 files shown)


trunk/src/kernel/srs_kernel_consts.hpp
```diff
@@ -77,7 +77,7 @@
 // if timeout, we generally ignore and send the data to client,
 // generally, it's the pulse time for data seding.
 // @remark, recomment to 500ms.
-#define SRS_CONSTS_RTMP_PULSE_TMMS (500)
+#define SRS_CONSTS_RTMP_PULSE_TMMS (int64_t)(500*1000LL)
```
winlinvip: This place should not be changed like this, it is TMMS, which stands for milliseconds. If you change it to 500*1000 milliseconds, it will be wrong.
You should multiply it by 1000 in every place where this variable is used.

`TRANS_BY_GPT3`

xialixin: Yes, that's correct. The modification above is indeed incorrect. SRS_CONSTS_RTMP_PULSE_TMMS should still be set to 500. So, all the calls to st_usleep(SRS_CONSTS_RTMP_PULSE_TMMS) should be changed to st_usleep(SRS_CONSTS_RTMP_PULSE_TMMS*1000).

`TRANS_BY_GPT3`

PR #902, opened 2017-05-23
Refine srs_app_thread code

HungMingWu:
SrsReusableThread and SrsReusableThread2 are most indetical.
ISrsReusableThreadHandler and ISrsReusableThread2Handler have same interface declaration.
Merge them into one class and one interface.


winlinvip:
I think the thread model of SRS should be refined, but maybe need more deeply work.
Thanks, I will look to it.

winlinvip:
Thread model refined.

PR #1003, opened 2017-10-29
to support flv.js(https://github.com/Bilibili/flv.js)

SnailTowardThesun:


winlinvip:
Directly setting is not the most reasonable way, especially when this header is not needed for CORS. SRS3 has already solved this problem. CORS in version 2.0 can be fixed by itself or using an HTTP proxy like https://github.com/ossrs/go-oryx/tree/develop/httpx-static.

`TRANS_BY_GPT3`

wony-zheng:
3.0 has another issue. When sending metadata, if there is no audio information specified with the hasAudio attribute set to false, then flv.js crashes.

`TRANS_BY_GPT3`

winlinvip:
3.0 does not specify hasAudio? Or does it mean that when there is no audio, SRS3 will also set hasAudio to true?

`TRANS_BY_GPT3`

PR #1039, opened 2018-01-04
update init.d file

RocFang:
Hi, @winlinvip 

there are two points may need to be update:
1. In function of start, since srs is already running and failed_msg is used, maybe it's not proper to return 0.
2. this check will always success after ret=$?

thanks.


winlinvip:
👍 I will merge to 2.0release/3.0release/develop.
I will merge to master when next version of 2.0release is released.

PR #1045, opened 2018-01-07
set tcp keepalive for tcp socket

juntaoliu:


winlinvip:
Fix #1044 Issue of abnormal disconnection of TCP connection in live streaming pull flow end cannot be released.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_listener.cpp
```diff
@@ -223,6 +223,16 @@ int SrsTcpListener::listen()
     }
     srs_verbose("setsockopt reuse-addr success. port=%d, fd=%d", port, _fd);
     
+#ifdef SO_KEEPALIVE
```
winlinvip, line 226: Why use macro definition to enable this setting? Why not always set SO_KEEPALIVE?
This macro should be undefined, so this code will not take effect, right?

`TRANS_BY_GPT3`

winlinvip: 我改成了SRS_PERF_SO_KEEPALIVE，默认不开启。SRS2就这样了。

winlinvip: This macro is system-defined and is a reasonable way to write it.

`TRANS_BY_GPT3`

juntaoliu: This is not set to open using macros, but to enhance the portability of the code. Most Unix-like systems generally define SO_KEEPALIVE in the system, so there is no problem with or without this macro judgment in such systems. However, if the code is ported to other systems, those systems may not support SO_KEEPALIVE.

`TRANS_BY_GPT3`

Issue #948, opened 2017-08-01
srs遇到一个coredump


juntaoliu:
warning: exec file is newer than core file.
[New Thread 49573]

warning: Corrupted shared library list

warning: Corrupted shared library list
Core was generated by `./objs/srs -c ./conf/srs.conf'.
Program terminated with signal 11, Segmentation fault.
#0  0x0000003e5689d258 in ?? ()
(gdb) where
#0  0x0000003e5689d258 in ?? ()
#1  0x0000000000001388 in ?? ()
#2  0x0000000000a75780 in ?? ()
#3  0x00000000004e286c in SrsConfig::get_hls_cleanup (this=0xa75798, vhost=Cannot access memory at address 0x0
) at src/app/srs_app_config.cpp:3908
#4  0x00000000004a7d11 in SrsHls::dispose (this=Cannot access memory at address 0x4e2844
) at src/app/srs_app_hls.cpp:1166
Backtrace stopped: previous frame inner to this frame (corrupt stack?)

日志已经丢失，这是coredump的信息。srs的可执行文件已经加日志重新编译过，所以一些函数入口显示不全了。

juntaoliu:
Encountered another core, I'll make a note of it here. I'll investigate the cause slowly.

warning: exec file is newer than core file.
[New Thread 90307]

warning: Corrupted shared library list

warning: Corrupted shared library list

warning: Corrupted shared library list

warning: Corrupted shared library list
Core was generated by `./objs/srs -c ./conf/srs.conf'.
Program terminated with signal 11, Segmentation fault.
#0  0x0000003e5689d258 in ?? ()
(gdb) where
#0  0x0000003e5689d258 in ?? ()
#1  0x0000000000001388 in ?? ()
#2  0x0000000001425aa0 in ?? ()
#3  0x00000000004d5918 in SrsConfig::get_vhost_hls_nb_notify (this=0x1425ab8, vhost=...) at src/app/srs_app_config.cpp:3886
#4  0x00000000004880b7 in SrsSource::~SrsSource (this=0x8993a0, __in_chrg=<value optimized out>) at src/app/srs_app_source.cpp:1000
#5  0x00000000004871c6 in SrsSource::fetch_or_create (r=0x1477920, h=0x7fffa3778740, pps=0x49bdf3) at src/app/srs_app_source.cpp:758
#6  0x0000000000473ae2 in SrsServer::dispose (this=0x14011a0) at src/app/srs_app_server.cpp:567
[...]

PR #1055, opened 2018-01-26
fix a core brought by hls->dispose()

RocFang:
当配置为rtmp边缘，且配置中有hls块时，如果先推流到这个边缘(虽然边缘推的时候不会生成hls)，之后 kill或restart srs，会coredump。

分析了一下coredump的原因，在kill srs时，会通过source的dispose执行hls的dispose,而hls中的_req成员在配置为边缘的时候，在SrsRtmpConn::acquire_publish中不会执行source->on_publish，进而不会执行hls->on_publish,进而不会去拷贝request到hls的_req成员，那么在hls的dispose中，get_hls_dispose(_req->vhost)就会core掉。

原因找到了，解决这个coredump可以在几个地方做，这个pr只是一个方法，还请
@winlinvip 帮看下，谢谢！

### DIFF
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -1018,7 +1018,9 @@ SrsSource::~SrsSource()
 void SrsSource::dispose()
 {
 #ifdef SRS_AUTO_HLS
-    hls->dispose();
+    if (!_srs_config->get_vhost_is_edge(_req->vhost)) {
+        hls->dispose();
+    }
 #endif
     
     // cleaup the cached packets.


RocFang:
I searched for issues and found the following reports related to this: 

1. https://github.com/ossrs/srs/issues/884
2. https://github.com/ossrs/srs/issues/948

`TRANS_BY_GPT3`

winlinvip:
👍

winlinvip:
Is it possible to create a copy of 'req' when creating a Source with edge triggering?

`TRANS_BY_GPT3`

winlinvip:
It means to ensure that Source always has a usable _req of its own, because req contains vhost information, and having valid vhost information is reasonable for Source.

`TRANS_BY_GPT3`

RocFang:
In fact, there is _req in the source, which is why in the dispose of source in the PR, we can check the vhost of source's _req. 
Are you trying to say that we should also ensure there is a usable _req in HLS? :)

`TRANS_BY_GPT3`

RocFang:
When creating a source, the request is originally copied to its own _req in the initialize function of the source. Then, the initialization of hls->_req is done by copying this source's _req again in hls->on_publish. Therefore, Source indeed always has its own available _req. Haha.

`TRANS_BY_GPT3`

RocFang:
Now we need to see if it is necessary to also force a copy in the initialization of HLS.

`TRANS_BY_GPT3`

RocFang:
@winlinvip After thinking about it, this pull request is too abrupt. Although it can avoid the core problem caused by hls->dispose() when pushing to the edge, hls->dispose() is needed for the Hls triggered by play and locally cut. After making this change, we can no longer dispose it.

In this case, can we move the process of copying req from SrsHls::on_publish to hls->_req to SrsHls::initialize?

After making this change, for the edge that is configured with hls, whether it is triggered by push or play, source->hls will have a copy of the _req member.

`TRANS_BY_GPT3`

PR #1080, opened 2018-03-05
fix create dir success result check

xialixin:
When creating TS slices, after the creation of the directory, the check state is not true, resulting in the failure of the publish stream.


winlinvip:
:+1

Issue #1126, opened 2018-04-24
/etc/init.d script not finishing


dayjaby:
On calling /etc/init.d/srs start, the script hangs up. Using & to detach the process solves this issue:

https://github.com/ossrs/srs/blob/master/trunk/etc/init.d/srs#L79
https://github.com/ossrs/srs/blob/master/trunk/etc/init.d/srs#L81

winlinvip:
You can enable the daemon in config.

PR #1127, opened 2018-04-24
resolves issue #1126

dayjaby:
Resolves https://github.com/ossrs/srs/issues/1126


winlinvip:
Please enable daemon in config.

PR #1146, opened 2018-05-23
relax the restriction of edge's origin config. make it not crash.

RocFang:
@winlinvip  hi，
如果edge没有配置origin配置项的话，在推流时会crash掉。这个限制似乎有点太严厉了，在现实场景中，新增一个vhost时，存在由于疏忽忘记增加origin了的情况，本来通过推流失败以及错误日志是能发现问题并修改的。

### DIFF
--- a/trunk/src/app/srs_app_edge.cpp
+++ b/trunk/src/app/srs_app_edge.cpp
@@ -600,7 +600,11 @@ int SrsEdgeForwarder::connect_server(string& ep_server, string& ep_port)
     close_underlayer_socket();
     
     SrsConfDirective* conf = _srs_config->get_vhost_edge_origin(_req->vhost);
-    srs_assert(conf);
+    if (!conf) {
+        ret = ERROR_RTMP_EDGE_ORIGIN_NOT_FOUND;
+        srs_warn("vhost %s origin not found. ret=%d", _req->vhost.c_str(), ret);
+        return ret;
+    }
     
     // select the origin.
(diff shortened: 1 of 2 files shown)


winlinvip:
Configuration errors should be reported during the configuration file check (during startup or reload) instead of adding logs when using the configuration.
When using the configuration, it is essential for the configuration to be correct; otherwise, it becomes complicated and difficult to ensure that every aspect is checked.
In general, configuration files should not be maintained by humans but should be generated and checked by programs before being used by servers.

`TRANS_BY_GPT3`

RocFang:
Thank you~ @winlinvip

The principle that the configuration file you mentioned is not maintained by humans is certain. In practice, services should be provided in this way. It's just that I accidentally discovered it during testing. The specific process is as follows: In fact, we made a small modification when obtaining the origin, based on the interface instead of the configuration file. Therefore, it is possible to omit the explicit configuration of the source IP in the configuration file. However, the version I happened to test was an old version, so this problem occurred :)

Later, I looked at the corresponding code (connect_server) of the edge ingester, where this crash prevention is implemented. Although the purpose is to prevent crashes when users delete vhosts and then reload, it also unintentionally prevents crashes when the origin IP is not configured in the settings.

I think since this is a situation where errors can be clearly indicated, it might be better to avoid the service being down. However, I also support the principle you mentioned, so I will turn it off.

`TRANS_BY_GPT3`

Issue #1235, opened 2018-09-28
_definst_ at the end of app name


MakarovYaroslav:
Is it possible to add some configuration parameter to ignore '\_definst\_' at the end of app?

I`d like to have possibility to watch stream via two URL (with '\_definst\_' and without) after setting this parameter to True.

nginx-rtmp added support for \_definst\_:
http://nginx-rtmp.blogspot.com/2013/09/support-for-definst-in-nginx-rtmp-module.html

winlinvip:
To support AMS/FMS?

MakarovYaroslav:
@winlinvip To support Wowza Media Server(Wowza Streaming Engine).
We wish to replace wowza with srs. Wowza has logic with default instance (\_definst\_) and it leads to mismatch.

MakarovYaroslav:
No, we would like to publish stream with specified app (not only \_definst\_). 
For example:
We publish the stream to the URL rtmp://testhost:8888/app1/stream1 and we can watch the published stream by both links: rtmp://testhost:8888/app1/stream1 and rtmp://testhost:8888/app1/\_definst\_/stream1
Another case: 
We publish the stream to the url rtmp://testhost:8888/app1/\_definst\_/stream1 and can watch by both links: rtmp://testhost:8888/app1/\_definst\_/stream1 and rtmp://testhost:8888/app1/stream1

PR #1261, opened 2018-11-09
#1235 add ignoring _definst_ at the end of app

MakarovYaroslav:
I added ignoring default instance (\_definst\_) at the end of app in RTMP-URL
Issue #1235


winlinvip:
👍

PR #1285, opened 2018-12-09
Change ASM compiler for x264 and ffmpeg from yasm to nasm

T-bagwell:


winlinvip:
Suggestion for using Docker to compile and run SRS in other environments: https://github.com/ossrs/srs-docker/tree/centos#usage

Recommend to use docker to build and run SRS: https://github.com/ossrs/srs-docker/tree/centos#usage

`TRANS_BY_GPT3`

Issue #1300, opened 2019-01-17
播放器反复请求不存在的http-flv流，产生大量close_wait, 并会计入clients统计中


rageJune:
播放器反复请求不存在的http-flv流，反复拉取断开，close_wait均会增长，并且会计入统计中的clients中

srs version: 3.0.45
配置文件:  conf/http.flv.live.conf

![51233031-b0ea0080-19a3-11e9-923d-55d3d8059be6](https://user-images.githubusercontent.com/30453799/51233268-661cb880-19a4-11e9-85bf-fab56d74ff65.png)
![image](https://user-images.githubusercontent.com/30453799/51233049-c3fcd080-19a3-11e9-83f9-ec0fd9fdbe53.png)
![image](https://user-images.githubusercontent.com/30453799/51233153-1d64ff80-19a4-11e9-99e2-54e346ab29a9.png)

_Originally posted by @rageJune in https://github.com/ossrs/srs/issues/858#issuecomment-454682856_

rageJune:
Also tested the situation in cluster mode:
```
## edge.conf
listen              1935;
max_connections     1000;
pid                 objs/edge.pid;
daemon              off;
srs_log_tank        console;
http_api {
    enabled         on;
    listen          1985;
}
http_server {
    enabled         on;
    listen          8080;
    dir             ./objs/nginx/html;
}
vhost __defaultVhost__ {
    cluster {
        mode            remote;
        origin          127.0.0.1:19350;
    }
    http_remux {
        enabled     on;
        mount       [app]/[stream].flv;
        hstrs       on;
    }
}
```

After pulling a non-existent FLV stream and disconnecting, the connection between origin and edge is still being repeatedly established.
![image](https://user-images.githubusercontent.com/30453799/51233847-fd364000-19a5-11e9-94b1-200b2002c4b9.png)

`TRANS_BY_GPT3`

rageJune:
After carefully reviewing the solution and code in #636, it seems that the final issue is that the error was not passed to "trd" after receiving the client close signal SrsHttpConn::pop_message(&req).


`srs_error_t SrsHttpRecvThread::cycle() 
{
    srs_error_t err = srs_success;

    while ((err = trd->pull()) == srs_success) {
        ISrsHttpMessage* req = NULL;
        SrsAutoFree(ISrsHttpMessage, req);
        
        if ((err = conn->pop_message(&req)) != srs_success) {
[...]

PR #1304, opened 2019-01-22
#1300 拉取不存在的http-flv流时fd泄漏的问题

rageJune:
#1300  基于#636的小改动

### DIFF
--- a/trunk/src/app/srs_app_recv_thread.cpp
+++ b/trunk/src/app/srs_app_recv_thread.cpp
@@ -562,6 +562,7 @@ void SrsPublishRecvThread::set_socket_buffer(int sleep_ms)
 
 SrsHttpRecvThread::SrsHttpRecvThread(SrsResponseOnlyHttpConn* c)
 {
+    trd_err = srs_success;
     conn = c;
     trd = new SrsSTCoroutine("http-receive", this, _srs_context->get_id());
 }
@@ -584,7 +585,7 @@ srs_error_t SrsHttpRecvThread::start()
 
 srs_error_t SrsHttpRecvThread::pull()
 {
-    return trd->pull();
+    return srs_error_copy(trd_err);
 }
 
 srs_error_t SrsHttpRecvThread::cycle()
@@ -596,6 +597,7 @@ srs_error_t SrsHttpRecvThread::cycle()
         SrsAutoFree(ISrsHttpMessage, req);
         
         if ((err = conn->pop_message(&req)) != srs_success) {
+	    trd_err = err;
             return srs_error_wrap(err, "pop message");
         }
     }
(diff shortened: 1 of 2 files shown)


trunk/src/app/srs_app_recv_thread.cpp
```diff
@@ -596,6 +597,7 @@ srs_error_t SrsHttpRecvThread::cycle()
         SrsAutoFree(ISrsHttpMessage, req);
         
         if ((err = conn->pop_message(&req)) != srs_success) {
+	    trd_err = err;
```
winlinvip: The err here is shared by trd_err and wrapped err, we should free the trd_err then copy err to it.

winlinvip:
First, I will set the `hstrs` of SRS2 to `on` by default. As a streaming media server, it should actively fetch content from the origin. Reference: https://github.com/ossrs/srs/commit/a8781aee3e9dc046db3dc9f9e7436f4a4e13a492

`TRANS_BY_GPT3`

winlinvip:
Analyze the code and find that it is a low-level coroutine object. After actively exiting, the pull function cannot obtain the error.

The upper-level code will start a thread to read messages from the client.

```
srs_error_t SrsHttpRecvThread::cycle()
{
    
    while ((err = trd->pull()) == srs_success) {
        ISrsHttpMessage* req = NULL;
        SrsAutoFree(ISrsHttpMessage, req);
        
        if ((err = conn->pop_message(&req)) != srs_success) {
            return srs_error_wrap(err, "pop message");
```

The `pop_message` function will return an error after the client closes the connection, and then it will be retrieved and returned by the `pfn` function.

```
void* SrsSTCoroutine::pfn(void* arg)
{
    SrsSTCoroutine* p = (SrsSTCoroutine*)arg;
    void* res = (void*)p->cycle();
    return res;
```

And this error code is stored in the return value of the `st` coroutine.

```
326	  /* Run thread main */
327	  thread->retval = (*thread->start)(thread->arg);
(gdb) p thread->retval
$1 = (void *) 0x9c87c0
```

Usually, when the stop function stops the thread, the return value of the coroutine is obtained so that `pull` can know if there is an error.

```
void SrsSTCoroutine::stop()
{
    if (!started || disposed) {
        return;
    }
    disposed = true;
    
    interrupt();
    
    void* res = NULL;
    int r0 = st_thread_join((st_thread_t)trd, &res);
    srs_assert(!r0);
    
    // Always override the error by the error from worker.
    if ((srs_error_t)res != srs_success) {
        srs_freep(trd_err);
        trd_err = (srs_error_t)res;
```

In fact, there is no place to call `stop`, so `pull` cannot obtain the error. The solution is to set `trd_err` when the coroutine naturally exits, so that `pull` can return the error.

`TRANS_BY_GPT3`

winlinvip:
SRS2, SRS3, Origin, and Edge modes all passed the test without any FD leakage issues.

Fixed.

`TRANS_BY_GPT3`

Issue #1220, opened 2018-09-07
gcc 7.3 on Ubuntu 18.04 compile http-parser-2.1 error


aggresss:
build http-parser-2.1
patching file Makefile
cc -I. -DHTTP_PARSER_STRICT=1  -Wall -Wextra -Werror -O0 -g  -c http_parser.c
http_parser.c: In function 'http_parser_parse_url':
http_parser.c:2093:18: error: this statement may fall through [-Werror=implicit-fallthrough=]
         found_at = 1;

http_parser.c:2096:7: note: here
       case s_req_server:

cc1: all warnings being treated as errors
Makefile:40: recipe for target 'http_parser.o' failed
make: *** [http_parser.o] Error 1
build http-parser-2.1 failed, ret=2

huanghe2015:
Dup for  #1159 #1065

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/a554dd85e110ca4ce9ec35423c9886edaff90371 in 2.0.263

Issue #1159, opened 2018-06-22
Compile Error


Baa14453:
Hello,
I am trying to compile on Ubuntu 18.04 by running `./configure --full && make`
I get the following error, which seems similar to [1065](https://github.com/ossrs/srs/issues/1065).
```

src/core/ngx_murmurhash.c: In function ‘ngx_murmur_hash2’:
src/core/ngx_murmurhash.c:37:11: error: this statement may fall through [-Werror=implicit-fallthrough=]
         h ^= data[2] << 16;
         ~~^~~~~~~~~~~~~~~~
src/core/ngx_murmurhash.c:38:5: note: here
     case 2:
     ^~~~
src/core/ngx_murmurhash.c:39:11: error: this statement may fall through [-Werror=implicit-fallthrough=]
         h ^= data[1] << 8;
         ~~^~~~~~~~~~~~~~~
src/core/ngx_murmurhash.c:40:5: note: here
     case 1:
     ^~~~
cc1: all warnings being treated as errors
objs/Makefile:429: recipe for target 'objs/src/core/ngx_murmurhash.o' failed
make[1]: *** [objs/src/core/ngx_murmurhash.o] Error 1
make[1]: Leaving directory '/home/baa/srs/srs/trunk/objs/nginx-1.5.7'
Makefile:8: recipe for target 'build' failed
make: *** [build] Error 2
build nginx-1.5.7 failed, ret=2

```

wnpllrzodiac:
High version gcc semantic  check is more strict？

winlinvip:
What's your SRS version? SRS2?

Baa14453:
I don't have the directory anymore, but I think it was SRS2, yes

wwng2333:
Plz try use build with ubuntu 16.04.

winlinvip:
OK, I'll have a look.

winlinvip:
Dup #1065

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/a554dd85e110ca4ce9ec35423c9886edaff90371 in 2.0.263

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/a554dd85e110ca4ce9ec35423c9886edaff90371 in 2.0.263

Issue #1154, opened 2018-06-08
configure failed when compile on Ubuntu 18.04 LTS 


dmoons:
Ubuntu is most recently release  Ubuntu 18.04 LTS
Linux version 4.15.0-22-generic (buildd@lgw01-amd64-013) (gcc version 7.3.0 (Ubuntu 7.3.0-16ubuntu3)) #24-Ubuntu SMP Wed May 16 12:15:17 UTC 2018

dmoons@ubuntu:~/srs/trunk$ sudo ./configure
[sudo] password for dmoons:
regenerate config: --prefix=/usr/local/srs --with-hls --with-hds --with-dvr --without-nginx --with-ssl --without-ffmpeg --with-transcode --with-ingest --with-stat --with-http-callback --with-http-server --without-stream-caster --with-http-api --with-librtmp --without-research --with-utest --without-gperf --without-gmc --without-gmp --without-gcp --without-gprof --without-arm-ubuntu12 --without-mips-ubuntu12 --log-trace
check gcc/g++/gdb/make
depends tools are ok
Ubuntu detected, install tools if needed
Ubuntu install tools success
SED is sed_utility
st-1.9t is ok.
build http-parser-2.1
patching file Makefile
cc -I. -DHTTP_PARSER_STRICT=1  -Wall -Wextra -Werror -O0 -g  -c http_parser.c
http_parser.c: In function ‘http_parser_parse_url’:
http_parser.c:2093:18: error: this statement may fall through [-Werror=implicit-fallthrough=]
         found_at = 1;
         ~~~~~~~~~^~~
http_parser.c:2096:7: note: here
       case s_req_server:
       ^~~~
cc1: all warnings being treated as errors
Makefile:40: recipe for target 'http_parser.o' failed
make: *** [http_parser.o] Error 1
build http-parser-2.1 failed, ret=2

dmoons:
same as #1065

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/a554dd85e110ca4ce9ec35423c9886edaff90371 in 2.0.263

Issue #1129, opened 2018-04-26
Configure Error


Henrocker:
```
./configure
regenerate config: --prefix=/usr/local/srs --with-hls --with-hds --with-dvr --wi           thout-nginx --with-ssl --without-ffmpeg --with-transcode --with-ingest --with-st           at --with-http-callback --with-http-server --without-stream-caster --with-http-a           pi --with-librtmp --without-research --with-utest --without-gperf --without-gmc            --without-gmp --without-gcp --without-gprof --without-arm-ubuntu12 --without-mip           s-ubuntu12 --log-trace
check gcc/g++/gdb/make
depends tools are ok
Ubuntu detected, install tools if needed
Ubuntu install tools success
SED is sed_utility
st-1.9t is ok.
build http-parser-2.1
patching file Makefile
cc -I. -DHTTP_PARSER_STRICT=1  -Wall -Wextra -Werror -O0 -g  -c http_parser.c
http_parser.c: In function ‘http_parser_parse_url’:
http_parser.c:2093:18: error: this statement may fall through [-Werror=implicit-           fallthrough=]
         found_at = 1;
         ~~~~~~~~~^~~
http_parser.c:2096:7: note: here
       case s_req_server:
       ^~~~
cc1: all warnings being treated as errors
make: *** [Makefile:40: http_parser.o] Fehler 1
build http-parser-2.1 failed, ret=2
```

starnc:
my error is same to you ,i cannot build my src

RocFang:
https://github.com/ossrs/srs/issues/1065

Henrocker:
Gonna close, as the solution, provided by Daniate in #1065 works for me.

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/a554dd85e110ca4ce9ec35423c9886edaff90371 in 2.0.263

PR #1312, opened 2019-01-26
Fix compile http-parser-2.1 fallthrouth error above gcc7

lightningkay:
by the way, I format some spaces in the shell, sorry for my obession
https://gcc.gnu.org/onlinedocs/gcc-7.3.0/gcc/Warning-Options.html#index-Wimplicit-fallthrough

### DIFF
--- a/trunk/auto/depends.sh
+++ b/trunk/auto/depends.sh
@@ -17,7 +17,7 @@
 function require_sudoer()
 {
     sudo echo "" >/dev/null 2>&1
-    
+
     ret=$?; if [[ 0 -ne $ret ]]; then 
         echo "\"$1\" require sudoer failed. ret=$ret";
         exit $ret; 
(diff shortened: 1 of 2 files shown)


trunk/auto/depends.sh
```diff
@@ -45,7 +45,7 @@ function Ubuntu_prepare()
             fi
         fi
     fi
-    
+
```
winlinvip: Please revert these empty lines

winlinvip:
Won't fix for SRS2.

winlinvip:
It's a common error for #1065 #1159 #1220 #1154 #1129 #1159

trunk/auto/depends.sh
lightningkay: I think this spaces is unnessary.

PR #1321, opened 2019-02-18
correct spell error

hy05190134:


winlinvip:
For master, we could accept these spell erros.
Please file PR to 3.0release, which is not realeased now.

PR #1551, opened 2020-01-02
fix memory leak

alphonsetai:
fix audio_samples don't release in SrsRtpPacket::copy


winlinvip:
Merged manually, thanks!

Issue #1554, opened 2020-01-04
srs 2.0 logrotate copytruncate 产生文件空洞


fastfading:
**描述(Description)**

> 描述你遇到了什么问题(Please description your issue here)
srs 2.0 logrotate copytruncate 产生文件空洞
1. SRS版本(Version): `2.0`
1. SRS的日志如下(Log):
```
\0\0\0 ....
normal log
```

**重现(Replay)**

> 在logrotate 中配置copy truncate
当rotate 执行的时候，下一个文件就会产生空洞

参考 fix
```
d = ::open(filename.c_str(), O_RDWR | O_APPEND); 
if(fd == -1 && errno == ENOENT) { 
fd = open(filename.c_str(), 
O_RDWR | O_CREAT | O_TRUNC | O_APPEND/*logrotate copytruncate need this flag*/, 
S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH 
); 
}
```
refer
https://blog.csdn.net/caomiao2006/article/details/12752359
**期望行为(Expect)**
log文件没有空洞

winlinvip:
Wiki:

https://github.com/ossrs/srs/wiki/v3_CN_LogRotate#copytruncate
https://github.com/ossrs/srs/wiki/v3_EN_LogRotate#copytruncate

limjoe:
I deployed version 3.0 of k8s cluster, and on one of the host machines, I used the copytruncate method to rotate logs. However, I encountered the issue of sparse files. Is it possible that version 3.0 release 0 does not support the copytruncate method yet, and currently only version 2.0 supports it?

`TRANS_BY_GPT3`

PR #1560, opened 2020-01-06
fix logrotate empty file hole issue

wnpllrzodiac:
refer to https://github.com/ossrs/srs/issues/1554


wnpllrzodiac:
make a mistake, select wrong branch to merge

PR #1566, opened 2020-01-09
clean up source and add publisher status

NodeBoy2:
打开清理SrsSource的代码，参考：https://github.com/ossrs/srs/issues/713
同时增加SrsSource是否活跃的变量，处理fetch_or_create()之后没有使用（publish或cosume）的情况。解决https://github.com/ossrs/srs/issues/1509 https://github.com/ossrs/srs/issues/1553 的问题

Issue #1553, opened 2020-01-03
ST: Low performance when got large amount of coroutines.


NodeBoy2:
**Description**'

Please ensure that the markdown structure is maintained.

> Long running, severe memory growth, and increased close_wait on the server. Using perf to investigate, it was found that heap_delete and heap_insert are causing significant delays.

1. SRS version: `2.0.258`
1. SRS logs:
   - Normal

**Replay**

**How to replay bug?**

> Steps to reproduce the bug

**1. Long-time operation/live streaming (using random values for streaming URLs, all different)**

**Expected behavior (Expect)**

> Is it because `std::map<std::string, SrsSource*> SrsSource::pool` is not released that causes the issue? Is there any solution available?

`TRANS_BY_GPT3`

NodeBoy2:
Using gdb to debug a long-running program (while streaming), in the _st_thread_create() function of the st library, the _st_active_count variable has reached 1695. At the same time, in the fetch_or_create() function of srs_app_source.cpp, the value of the pool is outputted, and it is found that the previous streaming records still exist.
`
(gdb) p _st_active_count
$1 = 1695
`
`
(gdb) p pool
$1 = std::map with 845 elements = {
  ["/mttest/fengyifan_test"] = 0x1a1e1c0,
  ["/test/test_200"] = 0x29d5b20,
 ["origin.test.com/test/003d94d6.10c9a7116ac01061"] = 0x3621f80,
...
}
`

`TRANS_BY_GPT3`

NodeBoy2:
Refer to https://github.com/ossrs/srs/issues/713. In SrsSource::fetch(), add source->die_at = -1; At the same time, add the active_at variable to SrsSource to record the last active time (update it in on_metadata, on_video, on_audio). When cleaning up SrsSource objects, first check if die_at has timed out, then check if the last active time has timed out. Can this solve the problem?

`TRANS_BY_GPT3`

winlinvip:
The waiting performance of ST needs to be optimized.

`TRANS_BY_GPT3`

watchpoints:
me too, CPU 20%'.

`TRANS_BY_GPT3`

winlinvip:
There is a PR https://github.com/ossrs/state-threads/pull/5 related to this issue, improving the performance for iterating all coroutines.

Issue #713, opened 2016-12-13
srs偶发崩溃，由清理无用SrsSource机制造成的


haofz:
加入清理无用SrsSource机制后，
[https://github.com/ossrs/srs/commit/c7b97aa1c3de380451e2d229fc5bef702e2d264d](url)
[https://github.com/ossrs/srs/commit/590e9517398c4d442bcf125421353b4b338217d1](url)
测试中发现srs程序崩溃问题，日志如下：
2016-12-13 14:16:42.790|trace|6104|243876|0|RTMP client ip=192.168.10.20
2016-12-13 14:16:42.792|trace|6104|243876|0|complex handshake success
2016-12-13 14:16:42.792|trace|6104|243876|0|connect app, tcUrl=rtmp://172.20.1.2:1935/live, pageUrl=, swfUrl=, schema=rtmp, vhost=__defaultVhost__, port=1935, app=live, param=, args=null
2016-12-13 14:16:42.792|trace|6104|243876|0|out chunk size to 60000
2016-12-13 14:16:42.872|trace|6104|243876|0|input chunk size to 60000
2016-12-13 14:16:42.873|trace|6104|243876|0|client identified, type=fmle-publish, stream_name=livestream, duration=-1.00
2016-12-13 14:16:42.873|trace|6104|243876|0|update req of soruce for auth ok
2016-12-13 14:16:42.873|trace|6104|243876|0|source url=__defaultVhost__/live/livestream, ip=192.168.10.20, cache=1, is_edge=0, source_id=-1[-1]
2016-12-13 14:16:42.888|trace|6104|243218|0|cleanup die source, total=43
2016-12-13 14:16:42.953|trace|6104|243876|0|>>>>>>>>>>>>> FMLE start to publish stream livestream, url=__defaultVhost__/live/livestream.

**分析原因：**
在一次推流的connect和正式推流之间的时间内有阻塞，该source被clear掉了，程序崩溃。

**过程：**
1、一次推流成功，断开；
2、29秒后二次推流，SrsSource被fetch()出来，fetch()方法中die_at未被置成-1，（此处与create不用）；
3、connect与publishing之间有阻塞，如expect_message()，如果此时正好到30秒，该source会被1s一次的循环清理机制干掉；
4、再使用该SrsSource时就崩溃了；

**解决：**
 SrsSource::fetch()中，加入source->die_at = -1;    个人测试可解决。

winlinvip:
[...]

Issue #1509, opened 2019-12-10
Source清理：大量推流时，Source泄漏导致OOM


2k9laojia:
**描述**
描述你遇到了什么问题。
在给srs推流时，出现srs进程core掉
**环境**
1. 操作系统：CentOS
1. 编码器(工具和版本)：```...```
1. 播放器(工具和版本)：```...```
1. SRS版本: 2.0.248
1. SRS的日志如下:
client identified, type=fmle-publish, stream_name=62c1352ea3a64621999bfd946081f392, duration=-1.00
connected stream, tcUrl=rtmp://172.16.129.97/netposa, pageUrl=, swfUrl=, schema=rtmp, vhost=__defaultVhost__, port=1935, app=netposa, stream=62c1352ea3a64621999bfd946081f392, args=null
**st_thread_create failed. ret=1017(Cannot allocate memory)**

**重现**
重现Bug的步骤如下：
1. 启动SRS，运行 ```...```
1. 推流，运行 ```...```
1. 重现了Bug，关键信息如下：

```
...
```

**期望行为**
描述你期望发生的事情。

通过查看堆栈分析，原因如下：
推流时调用SrsSource::initialize——》hls->initialize，由于hls->initialize中出现st_thread_create failed，所以返回错误码，在函数SrsSource::fetch_or_create中开始调用srs_freep(source)，进入SrsSource的析构函数，开始释放成员变量play_edge，调用其析构函数，进而调用SrsPlayEdge成员变量ingester析构函数，调用stop函数，最终调用_source->on_unpublish()，由于调用hls->initialize失败直接返回了，没有调用play_edge->initialize(this, _req)，所以此时SrsEdgeIngester中的_source是空指针，所以造成core掉。所以在使用_source前加判空处理可以解决问题。

**期望大神可以帮忙回答下，为什么st_thread_create failed会调用失败？**

winlinvip:
The reason is clearly stated in the log: 'Cannot allocate memory'. 
Can the scenario of running out of memory be reproduced?

`TRANS_BY_GPT3`

2k9laojia:
My server had 60GB of remaining memory at that time, and the current SRS process was using 3.4GB of memory. Does running out of memory refer to the memory limit for the current process? This issue was occurring repeatedly while handling 500 streaming routes, closing and restarting them.

`TRANS_BY_GPT3`

2k9laojia:
[
[...]

PR #1568, opened 2020-01-09
fix: clean up source and add publisher status

NodeBoy2:
打开清理SrsSource的代码，参考：#713
同时增加SrsSource是否活跃的变量，处理fetch_or_create()之后没有使用（publish或cosume）的情况。解决#1509 #1553 的问题。测试结果参考：https://github.com/ossrs/srs/pull/1566

### DIFF
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -1681,6 +1681,8 @@ SrsSource* SrsSource::fetch(SrsRequest* r)
     // and we only need to update the token of request, it's simple.
     source->req->update_auth(r);
     
+    source->die_at = 0;
+    source->active_at = srs_get_system_time();
     return source;
 }
 
@@ -1719,7 +1721,7 @@ srs_error_t SrsSource::do_cycle_all()
         // TODO: FIXME: support source cleanup.
         // @see https://github.com/ossrs/srs/issues/713
         // @see https://github.com/ossrs/srs/issues/714
-#if 0
+#if 1
         // When source expired, remove it.
         if (source->expired()) {
             int cid = source->source_id();
@@ -1764,6 +1766,7 @@ SrsSource::SrsSource()
     _can_publish = true;
     _pre_source_id = _source_id = -1;
     die_at = 0;
+    active_at = srs_get_system_time();
     
     play_edge = new SrsPlayEdge();
     publish_edge = new SrsPublishEdge();
@@ -1816,11 +1819,6 @@ srs_error_t SrsSource::cycle()
 
 bool SrsSource::expired()
 {
-    // unknown state?
-    if (die_at == 0) {
-        return false;
-    }
-    
     // still publishing?
     if (!_can_publish || !publish_edge->can_publish()) {
         return false;
@@ -1828,6 +1826,16 @@ bool SrsSource::expired()
     
     // has any consumers?
     if (!consumers.empty()) {
+        return false;
+    }
+
+    // unknown state?
+    if (die_at == 0) {
+        // inactive source?
+        int64_t now = srs_get_system_time();
+        if (now > active_at + SRS_SOURCE_CLEANUP * 1.5) { // use SRS_CONSTS_RTMP_RECV_TIMEOUT_US ?
+            return true;
+        }
         return false;
     }
     
@@ -2005,6 +2013,8 @@ bool SrsSource::can_publish(bool is_edge)
 
 srs_error_t SrsSource::on_meta_data(SrsCommonMessage* msg, SrsOnMetaDataPacket* metadata)
 {
+    active_at = srs_get_system_time();
+
     srs_error_t err = srs_success;
     
     // if allow atc_auto and bravo-atc detected, open atc for vhost.
@@ -2051,6 +2061,8 @@ srs_error_t SrsSource::on_meta_data(SrsCommonMessage* msg, SrsOnMetaDataPacket* 
 
 srs_error_t SrsSource::on_audio(SrsCommonMessage* shared_audio)
 {
+    active_at = srs_get_system_time();
+
     srs_error_t err = srs_success;
     
     // monotically increase detect.
@@ -2159,6 +2171,8 @@ srs_error_t SrsSource::on_audio_imp(SrsSharedPtrMessage* msg)
 
 srs_error_t SrsSource::on_video(SrsCommonMessage* shared_video)
 {
+    active_at = srs_get_system_time();
+    
     srs_error_t err = srs_success;
     
     // monotically increase detect.
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -1681,8 +1681,11 @@ SrsSource* SrsSource::fetch(SrsRequest* r)
     // and we only need to update the token of request, it's simple.
     source->req->update_auth(r);
     
-    source->die_at = 0;
-    source->active_at = srs_get_system_time();
+    // all consumers quit and no publisher?
+    if(source->is_die()) {
+        source->die_at = srs_get_system_time();
+    }
+    
     return source;
 }
 
@@ -1765,8 +1768,7 @@ SrsSource::SrsSource()
     
     _can_publish = true;
     _pre_source_id = _source_id = -1;
-    die_at = 0;
-    active_at = srs_get_system_time();
+    die_at = srs_get_system_time(); // all consumers quit and no publisher, update last die time
     
     play_edge = new SrsPlayEdge();
     publish_edge = new SrsPublishEdge();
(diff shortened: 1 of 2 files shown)


NodeBoy2:
https://github.com/ossrs/srs/issues/209 The problem of being unable to delete HLS segment files when deleting SrsSource. Since deleting HLS depends on SrsSource in the pool, the hls_dispose parameter will be ineffective, and the HLS files cannot be deleted at regular intervals or when the service restarts (the expiration time specified by hls_window is valid). It is necessary to consider a solution for deleting HLS segments in the future.

`TRANS_BY_GPT3`

winlinvip:
Points that need to be confirmed:
* Whether the slicing protocols like HLS and DASH are affected, as HLS is under Source and the slicing number and SequenceNumber of HLS need to be continuous in the next streaming.
* There will be complex reference relationships, and there was a crash after enabling cleaning before, which needs to be confirmed. #713 #714

`TRANS_BY_GPT3`

winlinvip:
In https://github.com/ossrs/srs/issues/1509, two points were optimized. When there are 4k different streams (sources), the memory decreased from 125MB to 45MB, and there will no longer be two coroutines for each source constantly residing.

* Resolve coroutine issue: [link](https://github.com/ossrs/srs/commit/857c783efcf42bb11a9cb9e90f395f6b1dca5796)
* Reduce source memory usage: [link](https://github.com/ossrs/srs/commit/7240fe304092a91c97bc36fb3f0c1d34f41f506c)

These two modifications have no impact on stability.

`TRANS_BY_GPT3`

NodeBoy2:
After deleting SrsSource, there is a problem that causes a coredump when accessing http-flv. When establishing a push stream, if http-flv is enabled, the stream will be mounted on the http-server, and the handler will establish a reference to SrsSource. However, there is no mechanism to unmount the stream, so releasing it will cause a coredump. It is necessary to clarify the impact caused by SrsSource. There is an idea, whether it is possible to introduce shared_ptr and weak_ptr to manage SrsSource? --- Deleting SrsSource will not cause a coredump. SrsSource will set the stream as unavailable in on_unpublish. It is because SrsLiveStream detects the stream as unavailable, it will recreate SrsSource and mount it to the http-server. Since there is no push stream, closing the client will not trigger the on_unpublish operation, which leads to a coredump when accessing it again.

`TRANS_BY_GPT3`

winlinvip:
There is another idea to support smooth exit for SRS, and I have roughly thought about it:

1. Remove the exclusive lock on the PID file, close the API port, and allow a new SRS to start.
2. Use REUSEPORT to start a new SRS, where both the old and new SRS are providing services using the same PID file.
3. The old SRS will no longer accept new connections. It will exit after serving existing clients or after a certain period of time, such as 12 hours.

In this way, the old SRS can easily and safely release the created sources. The only problem is that when both the old and new SRS are providing services, the API is provided by the new SRS. Therefore, the number of clients in the system will not be accurate, as the old SRS will miss counting the clients it is serving.

> Remark: If it is a source cluster, the stream will be on the old SRS, which will result in the inability to detect the stream. In such cases, it is necessary to forcefully disconnect the stream. The client needs to support retries in order to smoothly support this, and an Edge can be placed before the source cluster to support retries in this way.

`TRANS_BY_GPT3`

winlinvip:
I have provided a detailed description of the plan in #1579. If you agree with this plan, I will close this PR.

`TRANS_BY_GPT3`

NodeBoy2:
> 方案我详细描述在了 #1579 ，如果同意这个方案，我会关闭这个PR。

NodeBoy2:
> I have described the plan in detail in #1579. If you agree with this plan, I will close this PR.

Okay, you can close this PR. Meanwhile, I will continue to refine the logic relationship related to SrsSource. 🙏🙏🙏.

`TRANS_BY_GPT3`

winlinvip:
Thank you, you did a great job and also inspired me to find new solutions. If there are better solutions, we can continue discussing them. I believe there is no best solution, only more stable and simpler ones.

`TRANS_BY_GPT3`

duiniuluantanqin:
> > I have described the plan in detail in #1579. If you agree with this plan, I will close this PR.
> 
> Okay, you can close this PR. Meanwhile, I will continue to refine the logic relationship of SrsSource. 🙏🙏🙏.

Hey, may I ask if you have continued to refine this issue recently?
Because I don't think hot upgrade is necessarily the best solution. It would be more in line with the requirements if we can clean up the source in a timely manner.
If you have already resolved this issue, I would also appreciate it if you could submit a PR.

`TRANS_BY_GPT3`

NodeBoy2:
> > > I have described the plan in detail in #1579. If you agree with this plan, I will close this PR.
> > 
> > 
> > Okay, you can close this PR. Meanwhile, I will continue to refine the logic relationship of SrsSource. 🙏🙏🙏.
> 
> Hey, may I ask if you have continued to refine this issue recently?
> Because I don't think hot upgrade is necessarily the best solution. It would be more in line with the requirements if we can clean up the source in a timely manner.
> If you have already resolved this issue, I would appreciate it if you could submit a PR.

The reference relationships involved in SrsSource are quite complex, and there are many issues with removing the imports. It poses a high risk to the production environment, so we have not continued to follow up on the plan to delete SrsSource.

`TRANS_BY_GPT3`

duiniuluantanqin:
> > > > I have described the plan in detail in #1579. If you agree with this plan, I will close this PR.
> > > 
> > > 
> > > Okay, you can close this PR. Meanwhile, I will continue to refine the logic relationship of SrsSource. 🙏🙏🙏.
> > 
> > 
> > Hey, may I ask if you have continued to refine this issue recently?
> > Because I think hot upgrade may not be the optimal solution, it would be more in line with the requirements to clean up the source in a timely manner.
> > If this issue has already been resolved, I would still appreciate it if you could submit a PR.
> 
> The reference relationships involved in SrsSource are quite complex, and there are many issues with deleting the imports. It poses a high risk to the production environment, so we have not continued to follow up on the plan to delete SrsSource.

> When I load tested with 1000 streams, each at 720P with a 1Mbps bitrate, the memory usage increased to 3GB. This would make it impractical for commercial use. The author mentioned that the memory usage for 4K streams was only 45MB, but I'm not sure how they tested it.
> 
> Lastly, I wanted to ask if you handled this issue using the smooth exit method mentioned by the author, or if there are other solutions available?
> 
> Thank you very much!

`TRANS_BY_GPT3`

NodeBoy2:
> > > > > I have described the plan in detail in #1579. If you agree with this plan, I will close this PR.
> > > > 
> > > > 
> > > > Okay, you can close this PR. Meanwhile, I will continue to refine the logic relationship of SrsSource. 🙏🙏🙏.
> > > 
> > > 
> > > Hey, may I ask if you have continued to refine this issue recently?
> > > Because I think hot upgrade may not be the optimal solution, it would be more in line with the requirements to clean up the source in a timely manner.
> > > If this issue has already been resolved, I would appreciate it if you could submit a PR.
> > 
> > 
> > The reference relationships involved in SrsSource are quite complex, and there are many issues with deleting the imports. It poses a high risk to the production environment, so we have not continued to follow up on the plan to delete SrsSource.
> 
> I stress-tested with 1000 streams, each streaming at 720P with a bitrate of 1Mbps, and the memory usage increased to 3GB. This would make it unsuitable for commercial use. The author mentioned that 4K streams only occupy 45MB of memory, but I'm not sure how they tested it.
> 
> Lastly, may I ask if you handled this issue using the smooth exit method mentioned by the author? Or do you have any other solutions?
> 
> Thank you very much!

> Currently, I am using the method of releasing SrsSource. However, the known issue is that it affects HLS cleanup and HTTP-FLV. The server currently does not have HTTP-FLV enabled, and the cleanup of HLS fragments is handled by another service. Mr. Yang has a fix for this issue in this link: https://github.com/ossrs/srs/issues/1509. It involves deleting the two resident coroutines of SrsSource when stopping the stream, and also changing the default size of the fast vector from 8k to 8 bytes. Have you tried merging this fix and testing it?

`TRANS_BY_GPT3`

duiniuluantanqin:
> The method I am currently using is to release SrsSource. However, the known issue at the moment is that it will affect hls cleaning and http-flv. The server temporarily does not have http-flv enabled, and cleaning hls slices is done by another service. Mr. Yang has fixed this issue in #1509 by deleting the two resident coroutines of SrsSource when stopping the stream, and also changing the default size of fast vector from 8k to 8 bytes. Have you tried merging and testing it?

> Changing the default size of fast vector did not solve the problem. May I ask if you have uploaded the code for cleaning SrsSource to GitHub? Is it this one: https://github.com/NodeBoy2/srs/tree/fix_cleansource3.0/trunk? If it is, I will refer to it and make the necessary modifications. Thank you.

`TRANS_BY_GPT3`

winlinvip:
Dup to #1509
Solution: https://github.com/ossrs/srs/issues/1579#issuecomment-736232139

PR #1599, opened 2020-02-12
add srt parameter configure

runner365:
add srt parameter configure: maxbw, latency, peerlatency, recvlatency, sendbuf, recvbuf;


trunk/conf/srt.conf
```diff
@@ -18,6 +19,10 @@ http_server {
 srt_server {
     enabled on;
     listen 10080;
+    maxbw 1000000000;
+    connect_timeout 4000;
+    peerlatency 300;
+    recvlatency 300;
```
winlinvip, line 25: You have modified srt.conf, so full.conf should also be modified accordingly, and make sure to add comments below.
We need to ensure that full.conf contains all the configurations.

`TRANS_BY_GPT3`

PR #1615, opened 2020-02-24
update streamid decode for get more encoder such as VMIX

runner365:
update streamid decode for get more encoder such as VMIX.

If streamid is "123456", the rtmp address become rtmp://127.0.0.1/live/123456 in default publish mode.
This is for encoder such as VMIX.

### DIFF
--- a/trunk/src/srt/srt_conn.cpp
+++ b/trunk/src/srt/srt_conn.cpp
@@ -4,28 +4,43 @@
 #include <vector>
 
 bool is_streamid_valid(const std::string& streamid) {
-    int mode = ERR_SRT_MODE;
-    std::string url_subpash;
-
-    bool ret = get_streamid_info(streamid, mode, url_subpash);
-    if (!ret) {
-        return ret;
-    }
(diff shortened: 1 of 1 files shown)


trunk/src/srt/srt_conn.cpp
```diff
@@ -46,13 +61,21 @@ bool get_key_value(const std::string& info, std::string& key, std::string& value
...
 
-    size_t pos = streamid.find("#!::h");
+    mode = PUSH_SRT_MODE;
+
+    size_t pos = streamid.find("#!::");
     if (pos != 0) {
-        return false;
+        pos = streamid.find("/");
+        if (pos == streamid.npos) {
+            url_subpath = "live/" + streamid;
```
winlinvip, line 74: https://github.com/ossrs/srs/commit/90afd06c85fb76e81f6a9d9931649fffde1713b0

Specify the default app name in the configuration file, which is set to "live" by default.

`TRANS_BY_GPT3`

PR #1631, opened 2020-03-07
add h264 sei filter configure

runner365:
add h264 sei filter configure in srt2rtmp


trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -459,6 +459,12 @@ srs_error_t rtmp_client::on_ts_video(std::shared_ptr<SrsBuffer> avs_ptr, uint64_
         if (nal_unit_type == SrsAvcNaluTypeAccessUnitDelimiter) {
             continue;
         }
+
+        if (_srs_config->get_srt_sei_filter()) {
```
winlinvip, line 463: Reading the config for every video frame may pose certain risks. It is recommended to cache this configuration at the rtmp_client level and update it if necessary using reload.

`TRANS_BY_GPT3`

Issue #1500, opened 2019-11-26
支持对接GB28181吗？监控，智能摄像头，国标推流，RTSP


vickllny:
## Usage

编译SRS，需要切换到Develop分支，并开启gb28181功能：

```
git checkout develop &&
./configure --with-gb28181 && 
make clean && make
```

然后使用配置文件`conf/push.gb28181.conf `启动：

```
./objs/srs -c conf/push.gb28181.conf 
```

> Remark: 一定要修改配置文件中的`host`配置，改成你的服务器的IP，摄像头能访问到的这个IP。后续会改进为自动获取，目前还需要修改配置。

然后按下面的操作配置摄像头，推流到SRS：

![image](https://user-images.githubusercontent.com/2777660/78043273-bb749800-73a5-11ea-97e8-01f01077692c.png)

最后，观看RTMP流：[rtmp://localhost:1935/live/34020000001320000001](http://ossrs.net/players/srs_player.html?app=live&stream=34020000001320000001&server=localhost&port=1935&autostart=true&vhost=localhost)

> Remark: 海康摄像头默认会连接萤石云，可以在APP上看到摄像头的内网IP，然后访问这个IP进入管理页面。其他摄像头也有对应办法，可以配置摄像头使用GB28181推流。

> Remark：如果看不了流，请确认你的服务器IP，还有流名称也就是`视频通道编码ID`是否正确。

> Remark: 默认不开启声音，如果需要开启声音，要打开配置`audio_enable on;`，并且在摄像头的音频配置中开启。

> Remark: 若需要开启音频，需要将编码设置为AAC，同时采样率设置为44100HZ，否则Flash可能无法观看。

## 资料

[SRS4.0 GB28181配置说明与API接口文档 说明](https://github.com/xialixin/srs_code_note/blob/master/doc/srs_gb28181.md)

[SRS GB28181 PS流压测工具使用说明](https://github.com/xialixin/srs_code_note/blob/master/doc/srs_gb28181_sb_load_publish.md)

SIP认证：https://github.com/ossrs/srs/issues/1500#issuecomment-657480580

## Q&A

1. SRS支持RTSP服务器吗？

> 不支持，没有应用场景，请参考[#2304](https://github.com/ossrs/srs/issues/2304#issuecomment-826009290)。

luowuf:
Generally, it is to integrate with GB28181 devices to obtain video streams, and then push the streams to SRS to obtain a playback address. Development needs to be done by oneself.

[...]

PR #1633, opened 2020-03-10
For #1500, support gb28181

winlinvip:


winlinvip:
#1656 implemented a simple SIP server and GB28181, I will merge this PR into #1656.
Thanks @beikesong, hope we can continue to maintain and update gb28181 together.

`TRANS_BY_GPT3`

PR #1654, opened 2020-03-20
更新CMakeLists.txt文件，使项目能在clion编译通过

cainiaoDJ:
原先的CMakeLists.txt 在clion无法编译通过。修改后测试推拉流正常。


oaup:
ubuntu18 clion srs cmakefie file error: clear read-only status

The problem is caused by insufficient file permissions. To modify the file, you need to change the file permissions. 
Enter the root account and execute the command: chmod -R 777 file path 
Then you can load the Clion project normally.

`TRANS_BY_GPT3`

Issue #1680, opened 2020-03-30
Cluster redirect to different vhost


vladimir131313:
**描述(Description)**
I am trying to use srs cluster mode to redirect srs-edge to different srs-origins with different vhosts.
For example:
Step One:
client play rtmp -> srs-edge.test.vhost  (remote mode) -> srs-balancer.test.vhost (local mode) -> local coworker (this is my application that responds to an API cluster request)
Step Two:
local coworker (my application)  Send api answer with Origin: {IP,PORT, vhost=pub1-vhost}-> srs-balancer  Send RTMP Redirect with RTMP://IP:PORT/stream?vhost=srs-edge.test.vhost (it ignores vhost=pub1-vhost) ->  srs-edge.test.vhost  (remote mode) go to rtmp redirect

> 描述你遇到了什么问题(Please description your issue here)

Unfortunately, srs only uses the **Host** and **Port** from the api cluster response, but **Vhost** is still required for my cluster scheme.

1. SRS版本(Version): `3.0.140`
1. SRS的日志如下(Log):
 srs-edge.test.vhost  Log
```
[2020-03-30 19:26:34.459][Trace][61390][354] connect app, tcUrl=rtmp://srs-edge.test.vhost:20135/test, pageUrl=, swfUrl=, schema=rtmp, vhost=srs-edge.test.vhost, port=20135, app=test, args=null
[2020-03-30 19:26:34.459][Trace][61390][354] protocol in.buffer=0, in.ack=0, out.ack=0, in.chunk=128, out.chunk=128
[2020-03-30 19:26:34.561][Trace][61390][354] ignore AMF0/AMF3 command message.
[2020-03-30 19:26:34.642][Trace][61390][354] ignore AMF0/AMF3 command message.
[2020-03-30 19:26:34.642][Trace][61390][354] client identified, type=Play, vhost=srs-edge.test.vhost, app=test, stream=test, param=, duration=-1ms
[2020-03-30 19:26:34.642][Trace][61390][354] connected stream, tcUrl=rtmp://srs-edge.test.vhost:20135/test, pageUrl=, swfUrl=, schema=rtmp, vhost=srs-edge.test.vhost, port=20135, app=test, stream=test, param=, args=null
[2020-03-30 19:26:34.642][Trace][61390][354] source url=srs-edge.test.vhost/test/test, ip=46.242.119.122, cache=1, is_edge=1, source_id=-1[-1]
[...]

PR #1682, opened 2020-03-31
Vhost: For #1680, added support vhost in rtmp redirect

vladimir131313:


vladimir131313:
Hello @winlinvip . What do you think about it?
Thanks.

winlinvip:
I think it's better to use "alias" for vhost in origin, rather than change the vhost in edge.

Thanks for your PR~

Issue #863, opened 2017-04-28
APPLICATION: 当使用http-flv播放时，调用srs http api查看流量统计不对，客户端连接数是对的


xialixin:
当使用http-flv播放时，查看流量统计不对，send_bytes, kbps.send_30s两个参数一定数后一直不会增加，客户端连接是对的
查看了srs代码，发现在http-flv相关代码中统计代码
```
int SrsLiveStream::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessage* r)
{   ...
    // update the statistic when source disconveried.
    SrsStatistic* stat = SrsStatistic::instance();
    if ((ret = stat->on_client(_srs_context->get_id(), req, NULL, SrsRtmpConnPlay)) != ERROR_SUCCESS) {
        srs_error("stat client failed. ret=%d", ret);
        return ret;
    }
   ...
}
```
继续跟踪代码在SrsHttpConn中相关流量统计代码如下
```
void SrsHttpConn::resample()
{
    // TODO: FIXME: implements it
}

int64_t SrsHttpConn::get_send_bytes_delta()
{
    // TODO: FIXME: implements it
    return 0;
}

int64_t SrsHttpConn::get_recv_bytes_delta()
{
    // TODO: FIXME: implements it
    return 0;
}

void SrsHttpConn::cleanup()
{
    // TODO: FIXME: implements it
}
```
SrsHttpConn,SrsRtmpConn的基类都是SrsConnection。 在SrsConnection类中调用了SrsKbps。为了验让是否跟上面代码有关, 直接将上面代码修改如下，调用基类SrsConnection中的相关方法
```
void SrsHttpConn::resample()
{
    SrsConnection::resample()
}

int64_t SrsHttpConn::get_send_bytes_delta()
{
    return SrsConnection::get_send_bytes_delta()
}

int64_t SrsHttpConn::get_recv_bytes_delta()
{
    return SrsConnection::get_recv_bytes_delta();
}
void SrsHttpConn::cleanup()
{
     SrsConnection::cleanup()
}
```
重新编译后，启动srs,然后能过api查看相关参数，send_bytes,kbps.send_30s两个参数都一直在增加。
以上是测试过程中发现问题，提出来供大家参考。

winlinvip:
I will find time to take a look at why it was necessary to rewrite the base class for bandwidth statistics at that time.

`TRANS_BY_GPT3`

superxrb:
[...]

PR #1788, opened 2020-05-30
Properly initialize memory

HarukaMa:
Those two pointers were passed in without being initialized. If they don't get initialized here, uninitialized values will be used in caller function and will result in unusable statistics data in this case.

### DIFF
--- a/trunk/src/app/srs_app_http_conn.cpp
+++ b/trunk/src/app/srs_app_http_conn.cpp
@@ -75,6 +75,7 @@ SrsHttpConn::~SrsHttpConn()
 void SrsHttpConn::remark(int64_t* in, int64_t* out)
 {
     // TODO: FIXME: implements it
+    *in = *out = 0;
 }
 
 srs_error_t SrsHttpConn::do_cycle()


HarukaMa:
Not sure why it was not discovered in #863. However before it's actually implemented, this is required to correctly track inbound traffic.

LeafHCode:
> Not sure why it was not discovered in #863. However before it's actually implemented, this is required to correctly track inbound traffic.
Do you have try your srs after change and it works normal? remark is a virtual function, if srs set (*in=*out=0 ), srs http-flv won't work.

HarukaMa:
@LeafHCode Have you tried it before saying it won't work? We've been using this in production for quite a while now.

How is it a virtual function? It was empty and was just doing nothing. It is being actually called for every http connection.

winlinvip:
Hi guys, thanks for your PR, SRS 4.0 had fixed it:

```
void SrsHttpConn::remark(int64_t* in, int64_t* out)
{
    kbps->remark(in, out);
}
```

Very appreciate for your work. 👍

PR #1831, opened 2020-07-02
RTC: Support planar audio like aac.

tany0007:
The sample format of decoded AAC is planar(AV_SAMPLE_FMT_FLTP) not interlaced.


winlinvip:
👍

PR #1859, opened 2020-07-20
Enhancement: Add param and stream to on_connect

matinzd:
Add stream and param to hook call on_connect for better use


matinzd:
Could you please review this pr and merge into master? @winlinvip

trunk/src/app/srs_app_http_hooks.cpp
```diff
@@ -70,6 +70,8 @@ srs_error_t SrsHttpHooks::on_connect(string url, SrsRequest* req)
     obj->set("ip", SrsJsonAny::str(req->ip.c_str()));
     obj->set("vhost", SrsJsonAny::str(req->vhost.c_str()));
     obj->set("app", SrsJsonAny::str(req->app.c_str()));
+    obj->set("stream", SrsJsonAny::str(req->stream.c_str()));
```
winlinvip, line 73: stream does not have a value in `on_connect`, right? It only has a value when play or publish occurs.

However, adding it is also possible, but it may be an empty value.

`TRANS_BY_GPT3`

matinzd: Yeah, you can note that in the documentation that this may be null.

Issue #1871, opened 2020-07-23
视频时间戳交错增加，导致hls录制失败，以及queue丢帧


NodeBoy2:
**描述(Description)**
使用腾讯云SDK进行推流，偶发hls频繁切片，并且无法回调on_hls。情况触发时，音视频时间戳差距较大。
参考：https://github.com/ossrs/srs/issues/526 和 https://github.com/ossrs/srs/issues/1163
1. SRS版本(Version): `2.0.258` 以及 `3.0.141`
1. SRS的日志如下(Log):
```
version: 2.0.258
2020-05-17 12:30:02.673 - com.sankuai.cloud.media.live [WARN] 45995 258717 #XMDJ#{errno:2}#XMDJ# hls timestamp jump 371580930=>367559910
2020-05-17 12:30:02.906 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - audio drop ts segment, sequence_no=57, uri=8328eb72827940dca4968669ced7c47f/1589689802673.ts, duration=44.73, start=367559910
2020-05-17 12:30:02.921 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - audio drop ts segment, sequence_no=57, uri=8328eb72827940dca4968669ced7c47f/1589689802906.ts, duration=44.72, start=367577190
2020-05-17 12:30:02.934 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - audio drop ts segment, sequence_no=57, uri=8328eb72827940dca4968669ced7c47f/1589689802922.ts, duration=44.57, start=367594470
2020-05-17 12:30:03.525 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - -> HLS time=163004, stream dts=371616120(4129068ms), sno=58, ts=8328eb72827940dca4968669ced7c47f/1589689802934.ts, dur=44.30, dva=0p
2020-05-17 12:30:03.600 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - audio drop ts segment, sequence_no=57, uri=8328eb72827940dca4968669ced7c47f/1589689802934.ts, duration=44.30, start=367629030
2020-05-17 12:30:03.682 - com.sankuai.cloud.media.live [DEBUG] 45995 258717 - audio drop ts segment, sequence_no=57, uri=8328eb72827940dca4968669ced7c47f/1589689803601.ts, duration=44.40, start=367632720
[...]

PR #1872, opened 2020-07-23
fix https://github.com/ossrs/srs/issues/1871: 修复视频时间戳交错增加，导致hls录制失败，以及queue丢帧的问题

NodeBoy2:
***时间戳交替增加的问题***
会导致以下问题：
1. hls录制文件频繁触发，并且直接删除
2. 设置time_jitter  off，播放丢失视频（或音频）


winlinvip:
First of all, thank you very much for submitting the PR 👍 

HLS has never enabled timestamp correction algorithm because TS does not have a clear requirement for timestamp monotonicity. If corrected, it may cause other issues, but not correcting it may also lead to some problems. 

So, this PR still needs some time to consider whether the current problem requires timestamp correction and whether there are other solutions available. 

`TRANS_BY_GPT3`

winlinvip:

PR #1882, opened 2020-07-30
rtc_queue: bugifx compile fail

pengranpku:
Bugfix compile fail when configure setting --log-info.

Signed-off-by: Ran Peng <ranrpeng@tencent.com>


winlinvip:
Merged. 👍

Issue #1607, opened 2020-02-15
OCM(OriginClusterMaster): Use external application for Origin Cluster to discovery services


winlinvip:
在 #1501 中，描述了docker部署源站集群时，SRS实际上不知道自己的可访问的IP，docker启动SRS后分配的是一个NAT地址，而对外的地址只有docker才能知道。

解决方案是源站集群内部请求时，加了一个参数`coworker`，参考[#1501](https://github.com/ossrs/srs/issues/1501#issuecomment-559946431)的描述。比如：

```

+----------------------------------+                       +-----------------------------------+
|   +------------------------+     |                       |   +------------------------+      |
|   |  Origin A(172.16.0.10) |     | /api/v1/clusters?     |   |  Origin B(172.16.0.11) |      |
|   +------------------------+     | coworker=192.168.0.21 |   +------------------------+      |
|                                  +-->--------------------+                                   +
|     docker A(192.168.0.20)       +----------<------------+     docker B(192.168.0.21)        |
|   coworkers [ B<192.168.0.21> ]  | origin=192.168.0.21   |   coworkers [ A<192.168.0.20> ]   |
+----------------------------------+                       +-----------------------------------+
```

这个方案有个前提，就是我们预先知道docker A和B的地址，另外，地址变更后也会比较麻烦，需要更新其他源站。这种就是：手动服务发现和更新。

在K8s这种自动服务发现的框架中，当然部署源站集群多了也需要这么做，手动发现当然就不合适了，应该用单独的一个源站集群服务发现的服务来解决问题，这个单独的服务叫OCM(OriginClusterMaster)。比如：

1. 启动OriginA，OriginB，OriginC等一系列源站，配置coworkers为OCM的地址或域名，比如：`coworkers: ocm-service`，当然可以配置多个。
1. 启动OCM(OriginClusterMaster)服务，暴露的k8s服务名为ocm-service，集群内可以用这个DNS名称来连接集群。
1. 还需要配置OriginA、OriginB、OriginC等源站，将流的信息更新为ocm-service，比如`on_publish`和`on_unpublish`等等，可能还需要其他的事件，后续会更新。

> Note: OCM这个方案，每个Origin是单独提供服务，需要有自己的K8s service和DNS名字（自动就是K8s service name），Origin需要配置自己的service后通知OCM。

winlinvip:
The design goal of the OCM scheme is to support 100,000 routes.

[...]

PR #1887, opened 2020-08-02
added: cluster coworkers update of raw api

aesirteam:
### K8S环境中origin cluster实现coworkers自动发现

**原理:**

1. 通过K8S API创建源站Ｈeadless Service　　
2. 通过K8S API创建源站StatefulSet
3. 通过查询K8S DNS查询headless得到源站pod ip
4. 通过RAW API完成coworkers的更新，实现源站HPA弹性　　
5. 利用livenessProbe调用shell脚本来完成coworkers更新，也可兼顾healthcheck维持正确的pod生命周期


aesirteam:
### K8S Deployment File
[k8s-srs-clusters.yaml](https://github.com/aesirteam/srs-docker/blob/v3/k8s-srs-clusters.yaml)

`TRANS_BY_GPT3`

winlinvip:
Thank you very much for submitting this PR and providing a solution to the problem. 👍 

By using the service to discover the origin server and updating the configuration of each origin server through the HTTP RAW API, new instances of the origin server can be discovered. This is a very direct approach and does not rely on other services, which is great. 

The essence of this problem is that when the origin server cluster is scaled, it cannot automatically discover new instances of the origin server because the configuration file of the origin server has fixed IP or domain names. The most direct solution is to dynamically modify the configuration file. 

The problem with this solution lies in the HTTP RAW API, which is very unstable because it requires modifying the configuration and then reloading it. Especially in K8S, configuration files are usually configured and mounted to pods through configMap, and modifying the configMap will automatically trigger a reload. 

The RAW API solution is actually conflicting with the reload configuration solution. RAW requires the configuration to be writable, while configMap is read-only (or it is best not to be written by the program to avoid synchronization issues with multiple writes). 

The best way to scale the origin server cluster is to configure coworkers not with their individual IP addresses, but with the address of a service that manages the origin servers. Please refer to OCM: https://github.com/ossrs/srs/issues/1607#issuecomment-587297568. 

`TRANS_BY_GPT3`

PR #1895, opened 2020-08-06
fix bug : Segmentation fault when cleanup die source, because the member 'writer' uninitialize, in srs_app_hls.cpp

duiniuluantanqin:
fix bug : Segmentation fault when cleanup die source, because the member 'writer' uninitialize, in srs_app_hls.cpp[line 191]


winlinvip:
Has fixed. Thanks also. 👍

PR #1971, opened 2020-09-27
RTC: fix memory leak bug

duiniuluantanqin:


trunk/src/app/srs_app_hls.cpp
```diff
@@ -191,6 +191,7 @@ string SrsDvrAsyncCallOnHlsNotify::to_string()
 SrsHlsMuxer::SrsHlsMuxer()
 {
     req = NULL;
+    writer = NULL;
```
winlinvip, line 194: Below is the code with `wrinter=NULL`.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_queue.cpp
```diff
@@ -47,6 +47,11 @@ SrsRtpRingBuffer::SrsRtpRingBuffer(int capacity)
 
 SrsRtpRingBuffer::~SrsRtpRingBuffer()
 {
+    for (int i = 0; i < capacity_; ++i)
```
winlinvip, line 50: The `srs_freepa` below will release these packets, but if it loops once, it will cause duplicate releases, right?

`TRANS_BY_GPT3`

duiniuluantanqin: srs_freepa does not release the memory pointed to by the first-level pointer, and VLD also indicates a memory leak here.
In my actual test, when making 1000 repeated requests to WebRTC, the memory usage increases to 700M and does not get released.

After adding the code I submitted, VLD no longer indicates a memory leak. When making 1000 repeated requests to WebRTC, the memory usage increases by more than 10M.

Test method: Push RTMP with ffmpeg and loop the stream pulling using rtc_player.html.

`TRANS_BY_GPT3`

winlinvip: Hmm, after careful examination, it is necessary to release each pointer in the array.

```
SrsRtpPacket2* arr = new SrsRtpPacket2[10];
delete[] arr;

SrsRtpPacket2** arr = new SrsRtpRingBuffer*[10];
// It is also necessary to loop through each element and release it.
delete[] arr;
```

`TRANS_BY_GPT3`

winlinvip:
Merged by ba29095d1

trunk/src/app/srs_app_rtc_queue.cpp
duiniuluantanqin: Well, finally able to contribute to the open-source community, haha

`TRANS_BY_GPT3`

winlinvip: 👍

PR #1988, opened 2020-10-13
rtc:fix bug:clean up inactived stream in statistic

duiniuluantanqin:
rtc:fix bug:clean up inactived stream in statistic

https://github.com/ossrs/srs/issues/1947


duiniuluantanqin:
This PR has a bug, please do not use it. I will resolve it and resubmit.

`TRANS_BY_GPT3`

chia7:
@duiniuluantanqin What bug are you referring to? It seems like there is no problem?

`TRANS_BY_GPT3`

duiniuluantanqin:
> @duiniuluantanqin What bug are you referring to? It seems like there is no problem?

It will sometimes occur that there is "streaming" but the client count is 0, which is extremely difficult to reproduce.

`TRANS_BY_GPT3`

duiniuluantanqin:
> @duiniuluantanqin What bug are you referring to? It seems like there is no problem?

It will sometimes occur that there is "streaming" but the number of clients is 0, which is extremely difficult to reproduce.

`TRANS_BY_GPT3`

duiniuluantanqin:
> @duiniuluantanqin What bug are you referring to? It seems like there is no problem?

https://github.com/ossrs/srs/pull/2169/commits/d286091f0c6495974f645d963201d021f94890e9
`bug` has been fixed, try using this.

`TRANS_BY_GPT3`

NodeBoy2:
> > @duiniuluantanqin What bug are you referring to? It seems like there is no problem?
> 
> There may be a situation where there is "streaming" but the client count is 0, which is extremely difficult to reproduce.

> Expert, I would like to ask under what circumstances this situation may occur. It seems that there is not much difference in the logic between the two commits.

`TRANS_BY_GPT3`

duiniuluantanqin:
If you directly play the stream without pushing it, and then stop playing, this issue will occur. It's not a big problem and can be ignored.

`TRANS_BY_GPT3`

PR #2011, opened 2020-11-02
WebRTC：解决RTMP转RTC的爆音问题

PieerePi:
WebRTC：解决RTMP转RTC的爆音问题

### DIFF
--- a/trunk/src/app/srs_app_rtc_codec.cpp
+++ b/trunk/src/app/srs_app_rtc_codec.cpp
@@ -130,10 +130,10 @@ srs_error_t SrsAudioDecoder::decode(SrsSample *pkt, char *buf, int &size)
             return srs_error_new(ERROR_RTC_RTP_MUXER, "Failed to calculate data size");
         }
 
-        for (int i = 0; i < frame_->nb_samples; i++) {
-            if (size + pcm_size * codec_ctx_->channels <= max) {
-                memcpy(buf + size,frame_->data[0] + pcm_size*codec_ctx_->channels * i, pcm_size * codec_ctx_->channels);
-                size += pcm_size * codec_ctx_->channels;
+        for (int i = 0; i < codec_ctx_->channels; i++) {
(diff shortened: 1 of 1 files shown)


wnpllrzodiac:
old code only support mono audio?

PieerePi:
@wnpllrzodiac 

No, the incoming audio is stereo.

It's a very common case, you can reproduce this issue easily. Use OBSStudio to publish stream to SRS, and use WebRTC to play.

(gdb) p *frame_
$7 = {data = {
    0x2199b00 ""..., 
    0x219bb10 ""..., 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, linesize = {8192, 0, 0, 0, 0, 0, 0, 0}, extended_data = 0x20e4ec0, width = 0, 
  height = 0, nb_samples = 1024, format = 8, key_frame = 1, pict_type = AV_PICTURE_TYPE_NONE, sample_aspect_ratio = {num = 0, den = 1}, 

The size of first plane is 8192, but only has 1024 samples, that is 1024*4(FLTP) = 4096 bytes in total.
The size of second plane is 8192, but only has 1024 samples, that is 1024*4(FLTP) = 4096 bytes in total.

Since decoded audio frame is planar format, the linesize is greater than the size of useful data. In old code, the second channel data is lost and filled by another half useless data in first plane (that's noise).

@winlinvip Please have a look and merge this PR, thanks!

winlinvip:
Thank you very much, we have already reviewed this PR and there is indeed this issue. I will look into merging it as soon as possible.

`TRANS_BY_GPT3`

PieerePi:
Okay, thank you!
There is also a PR #2012, can you help take a look when you have time?

Also, you mentioned there is a discussion group for GB28181. How can I join it?

`TRANS_BY_GPT3`

winlinvip:
See also https://github.com/ossrs/srs/issues/2172

winlinvip:
For #2106

PR #2012, opened 2020-11-02
srs：修复srs随机器启动但有时启动不成功的问题

PieerePi:
srs：修复srs随机器启动但有时启动不成功的问题

### DIFF
--- a/trunk/src/main/srs_main_server.cpp
+++ b/trunk/src/main/srs_main_server.cpp
@@ -450,6 +450,7 @@ srs_error_t run_directly_or_daemon()
     }
     
     if(pid > 0) {
+        sleep(3);
         srs_trace("father process exit");
         exit(0);
     }


trunk/src/main/srs_main_server.cpp
```diff
@@ -450,6 +450,7 @@ srs_error_t run_directly_or_daemon()
     }
     
     if(pid > 0) {
+        sleep(3);
```
winlinvip: Why do we have to wait for 3 seconds? Can you provide some information?

`TRANS_BY_GPT3`

PieerePi: OpenSSL has some issues with shared processes between parent and child. Some references are provided for your reference.
https://wiki.openssl.org/index.php/Random_fork-safety
https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_fork_child.html (These new methods are available after 1.1.1, but currently srs is using 1.1.0e.)
In this case, I am asking the parent process to wait for a while, allowing the child process to successfully complete _srs_rtc_dtls_certificate->initialize before the parent process exits. Otherwise, there is a high probability of _srs_rtc_dtls_certificate->initialize failing, and srs cannot start with the machine.

Waiting for 3 seconds in the parent process is just a workaround and there may be better methods available. For example, the approach used by Janus is to have the parent process wait until the child process has finished initializing before exiting.
janus.c:4124, /* Ok, we're the parent: let's wait for the child to tell us everything started fine */
Is this problem only present when there is WebRTC functionality?

`TRANS_BY_GPT3`

winlinvip: This solution also solves the problem of initializing openssl, using sleep is definitely a temporary solution. Let me see how to solve it completely.

`TRANS_BY_GPT3`

shitizenlism: Tested, not working!

`TRANS_BY_GPT3`

PieerePi: > Tested, not working!

> Oh, it's possible that we're not dealing with the same issue. Please try another method.

`TRANS_BY_GPT3`

winlinvip:
https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_fork_child.html

> Platforms without fork(2) will probably not need to use these functions. Platforms with fork(2) but without pthread_atfork(3) will probably need to call them manually, as described in the following paragraph. Platforms such as **Linux that have both functions will normally not need to call these functions** as the OpenSSL library will do so automatically.

It seems that Linux has no problem.

`TRANS_BY_GPT3`

winlinvip:
Although this PR is not being merged, thank you very much @PieerePi.

`TRANS_BY_GPT3`

PR #2033, opened 2020-11-15
(1)support gb28181 tcp for rtp;(2)support gb28181 h265

yinjiaoyuan:


yinjiaoyuan:
(1) gb28181 supports the transmission of audio and video data using the TCP protocol. It can be configured in the push.gb2818.conf file by setting tcp_enable. 
(2) gb28181 supports H.265 video.

`TRANS_BY_GPT3`

iceylm:
Hello, after enabling tcp_enable, do I need to configure the camera? I have a Dahua camera, but I don't see any settings for tcp or udp. Previously, when I tested with udp, the video playback was normal. But after using tcp, there is no video stream. Is there any other configuration that needs to be done? Thank you.

`TRANS_BY_GPT3`

PR #2037, opened 2020-11-16
gb28181 support h265

yinjiaoyuan:


q446328246:
..There's a small bug, where is the reference.

`TRANS_BY_GPT3`

q446328246:
image

PR #2083, opened 2020-12-09
feat: add api v1 snapshot for initiative stream snapshot

ghostsf:
add api v1 snapshot for initiative stream snapshot

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -3584,7 +3584,7 @@ srs_error_t SrsConfig::check_normal_config()
         for (int i = 0; conf && i < (int)conf->directives.size(); i++) {
             SrsConfDirective* obj = conf->at(i);
             string n = obj->name;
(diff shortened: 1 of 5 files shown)


winlinvip:
Provide API calls for FFmpeg screenshot, the problem-solving capability is limited, and soon there will be many parameters that need to be supported, as well as the need for transcoding.

So it is better to provide a separate service, which is to call FFmpeg, not only can it take screenshots but also can do everything that FFmpeg can do.

It is not very suitable to wrap the FFmpeg API in SRS.

`TRANS_BY_GPT3`

duiniuluantanqin:
在安防场景，可以在web端通过canvas截图，也可以在推流端控制ipc截图。
通过api控制srs截图的场景我不熟悉。

duiniuluantanqin:
My suggestion is also unnecessary.
In short, if you want to use ffmpeg to capture screenshots, why not write your own program to control it.
Let's temporarily close it, and if someone still feels the need, we can reopen it.

`TRANS_BY_GPT3`

PR #2104, opened 2020-12-23
fix multiple PPS parse error

mypopydev:
Signed-off-by: Jun Zhao <barryjzhao@tencent.com>


winlinvip:
If there are multiple SPS/PPS, how should they be parsed?
According to the current modification, only the first SPS/PPS will be parsed, right? Is that not a problem?

`TRANS_BY_GPT3`

mypopydev:
> If there are multiple SPS/PPS, how should they be parsed?
> According to the current modification, only the first SPS/PPS will be parsed, right? Is that not a problem?

Actually, there is a potential issue with that, but I think it's a separate problem.

`TRANS_BY_GPT3`

PR #2014, opened 2020-11-02
GB28181：支持作为GB/T 28181上级平台; 各种异常和问题修复

PieerePi:
gb28181模块可用性增强

主要改动，
1. 支持作为GB/T 28181上级平台
2. 新的目录接口sip_query_devicelist (/api/v1/gb28181?action=sip_query_devicelist)
3. 各种异常和问题修复
4. 其他一些小改动

以上改动基于feature/rtc分支（现为4.0release分支），因为需要网页用WebRTC来拉GB28181的监控流，而feature/gb28181分支代码有点老了。

更详细的改动说明，请见提交日志。

PR #2109, opened 2020-12-24
bugfix-修复srs  --with-gb28181无法对接平台问题及一些小bug

xbpeng121:
bugfix-修复srs  --with-gb28181无法对接平台问题及一些小bug


trunk/conf/push.gb28181.conf
```diff
@@ -2,6 +2,7 @@
 
 listen                  1935;
 max_connections         1000;
+srs_log_level 	        info;
```
winlinvip: Don't modify the code that doesn't need to be changed, please undo it.

`TRANS_BY_GPT3`

trunk/conf/push.gb28181.conf
```diff
@@ -54,7 +55,7 @@ stream_caster {
 
     # 是否开启rtp缓冲
     # 开启之后能有效解决rtp乱序等问题
-    jitterbuffer_enable  on;
+    jitterbuffer_enable  off;
```
winlinvip: Here too, if it's not the code you need to change, please undo it.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -301,35 +301,41 @@ srs_error_t SrsGb28181PsRtpProcessor::on_rtp_packet(const sockaddr* from, const
...
-                //auto create channel generated id
-                std::stringstream ss, ss1;
-                ss << "chid" << pkt.ssrc;
-                std::string tmp_id = ss.str();
-
-                SrsGb28181StreamChannel channel;
-                channel.set_channel_id(tmp_id);
-                channel.set_port_mode(RTP_PORT_MODE_FIXED);
-                channel.set_ssrc(pkt.ssrc);
+            // //First, search according to the channel_id. Otherwise, search according to the SSRC.
```
winlinvip: Why do you want to comment it out? Debugging code? Or is there any problem?

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
```diff
@@ -461,28 +462,62 @@ srs_error_t SrsSipStack::parse_xml(std::string xml_msg, std::map<std::string, st
...
+}
 srs_error_t SrsSipStack::do_parse_request(SrsSipRequest* req, const char* recv_msg)
 {
     srs_error_t err = srs_success;
 
-    std::vector<std::string> header_body = srs_string_split(recv_msg, SRS_RTSP_CRLFCRLF);
-    if (header_body.empty()){
-        return srs_error_new(ERROR_GB28181_SIP_PRASE_FAILED, "parse reques message"); 
-    }
+    // std::vector<std::string> header_body = srs_string_split(recv_msg, SRS_RTSP_CRLFCRLF);
```
winlinvip: Why is it commented out?

`TRANS_BY_GPT3`

PieerePi:
bugfix - Fix the issue of unable to connect with the platform using --with-gb28181 and some minor bugs.

The develop branch has already resolved the issue of unable to connect with the platform. Please refer to PR #2014 for more details.

`TRANS_BY_GPT3`

PieerePi:
@winlinvip, please kindly take a look at PR #2011.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: The parsing here is problematic. It assumes that the received data is divided into two parts using CRLFCRLF. The first part is the first line + header, and the second part (body = header_body.at(1);) is the body. However, since the body may contain CRLFCRLF, header_body.at(1) is not the entire body. Therefore, a different method of division is needed, which is to use the first CRLFCRLF as the dividing point. The part before it is the first line + header, and the remaining part is the body.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: The two functions here:
1-on_rtp_packet_jitter()
2-on_rtp_packet()
Both functions here will use this piece of logic. In the `on_rtp_packet_jitter()` function, this logic is encapsulated in the `fetch_rtmpmuxer()` function. Here, the code is commented out and replaced with a call to `fetch_rtmpmuxer()`. This way, any future modifications to this logic can be made in just one place. For example, checking whether the `sip_invite_port_fixed` configuration is enabled can be written once in the `fetch_rtmpmuxer()` function.

`TRANS_BY_GPT3`

trunk/conf/push.gb28181.conf
xbpeng121: Okay, the commits of the conf file will be removed. It should be a previous negligence when submitting the commit, which also included the configuration file.

`TRANS_BY_GPT3`

trunk/conf/push.gb28181.conf
xbpeng121: Okay, the commits of the conf file will be removed. It should be a previous negligence when submitting the commit, which also included the configuration file.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: sip协议格式：
generic-message  =  start-line
                             *message-header
                             CRLF
                             [ message-body ]

xbpeng121 (commented):
> bugfix-修复srs --with-gb28181无法对接平台问题及一些小bug
> 
> develop分支已经无法对接平台问题了。详见PR #2014。

Thanks, 之前没看到你提交的PR，我修改的问题基本上和你修改的内容差不多。

PieerePi:
> > bugfix-Fixed the issue of unable to connect to the platform with --with-gb28181 and some minor bugs.
> > The issue of unable to connect to the platform has been resolved in the develop branch. Please refer to PR #2014 for details.
> 
> Thanks, I didn't see the PR you submitted before. The issues I modified are basically similar to what you modified.

Please take a look at the develop branch again, to see if there are any usable code snippets in your part.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -283,7 +287,7 @@ srs_error_t SrsGb28181SipSession::do_cycle()
                         invite_duration = 0;
                 }
 
-                srs_trace("gb28181: sip session=%s device=%s status(%s, %s), duration(%u)",
+                srs_info("gb28181: sip session=%s device=%s status(%s, %s), duration(%u)",
```
winlinvip, line 289: Why modify this?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -329,23 +333,33 @@ srs_error_t SrsGb28181SipSession::cycle()
...
-{
+{//map key:"device_id" value:"status,parental,parentid,name"
     std::map<std::string, std::string>::iterator it;
     for (it = lst.begin(); it != lst.end(); ++it) {
         std::string id = it->first;
-        std::string status = it->second;
-
+        std::vector<std::string> value = srs_string_split2(it->second, ",");
+        std::string status = value[0];
+        std::string parental = value[1];
```
winlinvip: Directly taking 1, 2, 3, won't there be any out-of-bounds issue? For example, with an unregulated client?

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.hpp
```diff
@@ -104,6 +104,7 @@ extern int srs_string_count(std::string str, std::string flag);
 extern std::string srs_string_min_match(std::string str, std::vector<std::string> flags);
 // Split the string by flag to array.
 extern std::vector<std::string> srs_string_split(std::string str, std::string flag);
+extern std::vector<std::string> srs_string_split2(std::string s, std::string seperator);
```
winlinvip: What is the difference between srs_string_split and srs_string_split2? Can you provide an example to illustrate, for instance:

```
srs_string_split("abc", "b") = ["a", "c"]
```

`TRANS_BY_GPT3`

xbpeng121: `srs_string_split("1,A,,address", ",") = ["1", "A", "address", ""]`
`srs_string_split2("1,A,,address", ",") = ["1", "A", "", "address"]`

Uncertain if srs_string_split() has any special requirements, so a new function srs_string_split2() has been added.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: The reason for changing to the 'info' level is that when connecting to the platform, there are usually several thousand or tens of thousands of devices registered. It is too much to print the status information of each device every time in the loop (approximately once every minute, and each time it is all devices), which would affect the visibility of other information. The status information of the devices can be viewed through an HTTP interface, which can be called when needed.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: No, the key-value content of the map here is encapsulated in the `do_parse_request()` function in the `srs_sip_stack.cpp` file. The value content is already guaranteed when writing, so if the client does not have that field or content, an additional delimiter "," will be added. Therefore, the number of splits will definitely be 4 and only 4.

Of course, this implementation is not the best. In theory, a class should be defined and then assigned values to the class.

The code for assigning values to the map is located in lines 678 to 697 of the `srs_sip_stack.cpp` file.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
PieerePi: Actually, this comment should be removed. Otherwise, there will be too many log files, and it will be difficult to find useful information.

`TRANS_BY_GPT3`

xbpeng121: `srs_info` will not have any output by default during compilation.

`TRANS_BY_GPT3`

PieerePi: Haha, I see. ;-)
I studied the code and found that the default log level is trace. Changing it to srs_info is indeed better, and when needed, changing the log level to info will provide this information.

I added an interface in the develop branch, /api/v1/gb28181?action=sip_query_devicelist. You can try it out, it can retrieve the entire directory, which is quite useful.

`TRANS_BY_GPT3`

xbpeng121: > Haha, I see. ;-)
> I studied the code and found that the default log level is trace. Changing it to srs_info is indeed better. When needed, changing the log level to info will provide this information.
> 
> I added an interface in the develop branch, /api/v1/gb28181?action=sip_query_devicelist. You can try it out, it can retrieve the entire directory, which is quite useful.

This content has also been implemented in this PR, but the implementation method is slightly different from yours. It reused the previous interface, and in the query session's HTTP API interface, it can output the device's "id, name, status, whether it is a parent node, parent node id" (including virtual directories). ;-)

`TRANS_BY_GPT3`

PieerePi: The original interface was too simple and not suitable for directory parsing.

The new interface `/api/v1/gb28181?action=sip_query_devicelist` in the `develop` branch has improved in the following aspects:

1. It supports connecting to multiple superiors, with each superior distinguished by `TopDeviceID`.
2. To determine if the directory is complete, it includes `SumNum` from the protocol and the actual `RealSumNum`.
3. Encoding conversion from GB2312 to UTF-8 is performed to ensure that the names are not garbled.
4. All items with content are preserved, as some directories are hierarchical based on `CivilCode`.

On the `develop` branch, I only added the `Name` to the existing interface and filtered out items that only have `Status`, excluding directories and similar elements.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.hpp
winlinvip: View Python: 

```
>>> "1,A,,address".split(",")
['1', 'A', '', 'address']
```

So the definition of srs_string_split2 is inappropriate, it would be better to check if it is string.empty() when calling it.

`TRANS_BY_GPT3`

xbpeng121: I don't know if I misunderstood your meaning. The difference I see between srs_string_split() and other split() functions is in the sorting of its results. 

For a normal split() function: `“1,A,,address” results in ["1", "A", "", "address"]`, but `srs_string_split("1,A,,address", ",") = ["1", "A", "address", ""]`, the position of the empty string has changed. It should have been in the third position, but the output result is in the fourth position. Python's split() function also places it in the third position, consistent with the newly defined srs_string_split2().

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
```diff
@@ -619,49 +654,74 @@ srs_error_t SrsSipStack::do_parse_request(SrsSipRequest* req, const char* recv_m
...
                 //Response@DeviceList@Item@DeviceID:3000001,3000002
-                std::vector<std::string> vec_device_id = srs_string_split(body_map["Response@DeviceList@Item@DeviceID"], ",");
+                std::vector<std::string> vec_device_id = srs_string_split(req->xml_body_map["Response@DeviceList@Item@DeviceID"], ",");
                 //Response@DeviceList@Item@Status:ON,OFF
-                std::vector<std::string> vec_device_status = srs_string_split(body_map["Response@DeviceList@Item@Status"], ",");
-                 
-                //map key:devicd_id value:status 
+                std::vector<std::string> vec_device_status = srs_string_split(req->xml_body_map["Response@DeviceList@Item@Status"], ",");
+                /*    edit by xbpeng 20201217  start */
+                //Response@DeviceList@Item@Name:平湖出入口,地铁A口
```
winlinvip: Should these pieces of information be removed? It feels like they involve privacy a bit.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
winlinvip: If the code is indeed useless, please remove it directly, without using long comments.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
winlinvip: If the code is indeed useless, please just remove it.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.hpp
winlinvip: Then your modification is correct:
```
srs_string_split("1,A,,address", ",") = ["1", "A", "address", ""] // Incorrect
srs_string_split("1,A,,address", ",") = ["1", "A", "", "address"] // Correct
```

Please directly modify srs_string_split for me.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: If generally connecting to a subordinate platform, it is highly likely that we will need to reconsider the division of virtual structures and the naming of cameras at this level.

Currently, our plan for usage is to use an interface to retrieve all camera information and save it in the business system. Then, we will call SRS to play the videos. Since SRS enables SIP service, this information can only be output by SRS and cannot be obtained elsewhere.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: Okay

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
PieerePi: @winlinvip @xbpeng121
Both, currently this PR and the develop branch have code duplication and conflicts, mainly related to commit https://github.com/ossrs/srs/commit/ffae1720eccb31e5d38c0cfc1ba1d26f1819ebad, which is PR #2014. It cannot be merged.
I suggest closing this PR first. xbepeng121, please take some time to think and reorganize this code on the develop branch, and then submit a new PR.

@xbpeng121, you modified the original interface, remember to also modify trunk/research/players/srs_gb28181.html to maintain consistency in functionality.

`TRANS_BY_GPT3`

xbpeng121: Ok, I will take some time to review your previous commit, make some modifications, and then submit the PR again.
Additionally, trunk/research/players/srs_gb28181.html does not need any modifications as it has already been tested. The added attribute information is only for use by other business systems and does not affect the functionality of srs_gb28181.html.

`TRANS_BY_GPT3`

PieerePi: Okay, then switch to the develop branch to work. PR #2014 is already on it, take a look at how to add your code.

`TRANS_BY_GPT3`

winlinvip: I have switched the PR to Develop. There are some conflicting areas, you can click on the `Resolve Conflicts` below to resolve the conflicts.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -306,35 +306,7 @@ srs_error_t SrsGb28181PsRtpProcessor::on_rtp_packet(const sockaddr* from, const
...
-              
-                srs_error_t err2 = srs_success;
-                if ((err2 = _srs_gb28181->create_stream_channel(&channel)) != srs_success){
-                    srs_warn("gb28181: RtpProcessor create stream channel error %s", srs_error_desc(err2).c_str());
-                    srs_error_reset(err2);
-                };
-
-                muxer = _srs_gb28181->fetch_rtmpmuxer(tmp_id);
-            }
+            muxer = fetch_rtmpmuxer(channel_id,pkt.ssrc);
```
PieerePi, line 309: Is this your adjusted code? @xbpeng121
There are significant changes here, right? Did you remove auto_create_channel?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2409,7 +2387,28 @@ void SrsGb28181Manger::rtmpmuxer_unmap_by_ssrc(uint32_t ssrc)
...
+        srs_warn("gb28181: invaild input at update_ssrc_by_invite_respond(), id is empty or new_ssrc=0");
+        return;
+    }
 
+    SrsGb28181RtmpMuxer* muxer = fetch_rtmpmuxer(id);
+    if (muxer) {//find
+        uint32_t old_ssrc = muxer->get_channel().get_ssrc();
+        if (old_ssrc!=new_ssrc) {
+            rtmpmuxer_map_by_ssrc(muxer,new_ssrc);
+            muxer->get_channel().set_ssrc(new_ssrc);
```
PieerePi: Why are you still submitting this code? Isn't there already a function called update_rtmpmuxer_to_newssrc_by_id? You should delete this part, shouldn't you?
muxer->get_channel().set_ssrc(new_ssrc);
There seems to be a problem with this line. The get_channel function returns a temporary class instance. Take a look at the update_rtmpmuxer_to_newssrc_by_id function.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -618,20 +619,20 @@ srs_error_t SrsGb28181SipService::on_udp_sip(string peer_ip, int peer_port,
 
             if (req->status == "200") {
                 srs_trace("gb28181: device unique id is %s@%s", sip_session->session_id().c_str(), req->sip_auth_id.c_str());
-                // if srs is external realm, ssrc is generated by source realm rather than srs
```
PieerePi: Did you delete the previous code?...

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
```diff
@@ -548,31 +549,46 @@ srs_error_t SrsSipStack::parse_xml(std::string xml_msg, std::map<std::string, st
...
+    if (header_body_split_pos == string::npos || header_body_split_pos==0){
         return srs_error_new(ERROR_GB28181_SIP_PRASE_FAILED, "parse reques message"); 
     }
-
-    std::string header = header_body.at(0);
-    //Must be added SRS_RTSP_CRLFCRLF in order to handle the last line header
-    header += SRS_RTSP_CRLFCRLF; 
-    std::string body = "";
-
-    if (header_body.size() > 1){
```
PieerePi: Did you delete the previous code again?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -332,37 +335,32 @@ srs_error_t SrsGb28181SipSession::cycle()
 }
 
 void SrsGb28181SipSession::update_device_list(std::map<std::string, std::string> lst)
-{
+{//map key:"device_id" value:"status,parental,parentid,name"
```
PieerePi: Does 'srs_gb28181.html' need to be changed?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: Has your code been used well in the production environment for a while?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: This part of the code. 
             `//First, search according to the channel_id. Otherwise, search according to the SSRC.
             //Some channel_id are created by RTP pool, which are different ports.
             //No channel_id are created by multiplexing ports, which are the same port
             if (!channel_id.empty()){
                 muxer = _srs_gb28181->fetch_rtmpmuxer(channel_id);
             }else {
                 muxer = _srs_gb28181->fetch_rtmpmuxer_by_ssrc(pkt.ssrc);
             }

             //auto crate channel
             if (!muxer && config->auto_create_channel){
                 //auto create channel generated id
                 std::stringstream ss, ss1;
                 ss << "chid" << pkt.ssrc;
                 std::string tmp_id = ss.str();

                 SrsGb28181StreamChannel channel;
                 channel.set_channel_id(tmp_id);
                 channel.set_port_mode(RTP_PORT_MODE_FIXED);
                 channel.set_ssrc(pkt.ssrc);

                 srs_error_t err2 = srs_success;
                 if ((err2 = _srs_gb28181->create_stream_channel(&channel)) != srs_success){
                     srs_warn("gb28181: RtpProcessor create stream channel error %s", srs_error_desc(err2).c_str());
                     srs_error_reset(err2);
                 };

                 muxer = _srs_gb28181->fetch_rtmpmuxer(tmp_id);
             }`
The code inside the `fetch_rtmpmuxer(channel_id,pkt.ssrc)` function is basically identical. The only difference is that in the `on_rtp_packet()` function, it is written directly, while in the `on_rtp_packet_jitter()` function, they are encapsulated inside the `fetch_rtmpmuxer()` function. Here, the `fetch_rtmpmuxer()` function is called, and the logic of `auto_create_channel` is not modified. @PieerePi

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: Here, there is no need to modify srs_gb28181.html. It calls an interface that returns a JSON, and the displayed content is also a JSON. The specific page logic will only use device_id and status, which is unchanged from the previous code.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: The problem being solved here is the same. I didn't pay attention when resolving conflicts and suddenly got used to keeping my own code. The previous code should be kept here. I will change it back later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: I can't see why `muxer->get_channel()` returns a temporary class instance. The reason for using the `update_ssrc_by_invite_respond()` function here is because it feels redundant to perform the copy actions in `update_rtmpmuxer_to_newssrc_by_id()`. Could you please explain why we can't use `muxer->get_channel().set_ssrc(new_ssrc)`?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: The functionality logic implemented here is the same, but only one can be used because it is considered that the content of the writing in update_rtmpmuxer_to_newssrc_by_id is somewhat redundant, so new code is adopted. If the writing in update_rtmpmuxer_to_newssrc_by_id is necessary, I will modify it back to use the code in update_rtmpmuxer_to_newssrc_by_id().

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: muxer->get_channel().set_ssrc(new_ssrc);
You only updated the ssrc of the temporary class instance, but did not update the ssrc of the muxer. Remember that get_channel returns a temporary class instance.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
PieerePi: srs_gb28181.html only has top-level directories and cameras. You have added all these hierarchical directories, how do you operate on the interface?
The top-level directory is the device, which can notify and update directories. The cameras (with status) can make calls and hang up. What about those intermediate directories? Should they be displayed or should we improve the hierarchy on this web interface?

We already have /api/v1/gb28181?action=sip_query_devicelist, so I don't think we need to update this rudimentary interface anymore.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: Remember that `get_channel` returns a temporary class instance." 

```python
show me the code
```

`void SrsGb28181StreamChannel::copy(const SrsGb28181StreamChannel *s){
    channel_id = s->get_channel_id();
    port_mode = s->get_port_mode();
    app = s->get_app();
    stream = s->get_stream();
    ip = s->get_ip();
    rtp_port = s->get_rtp_port();
    rtmp_port = s->get_rtmp_port();
    ssrc = s->get_ssrc();
    rtp_peer_ip = s->get_rtp_peer_ip();
    rtp_peer_port = s->get_rtp_peer_port();
    rtmp_url = s->get_rtmp_url();
    recv_time_str = s->get_recv_time_str();
    recv_time = s->get_recv_time();
}
void SrsGb28181RtmpMuxer::copy_channel(SrsGb28181StreamChannel *s)
{
    channel->copy(s);
}

SrsGb28181StreamChannel SrsGb28181RtmpMuxer::get_channel()
{
   return *channel;
}`

I didn't see that `muxer->get_channel().set_ssrc(new_ssrc);` modifies the SSRC and...
SrsGb28181StreamChannel mc = muxer->get_channel();
mc.set_ssrc(ssrc);
muxer->copy_channel(&mc);
What is the difference between them? Can you explain it more clearly?

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
```diff
@@ -718,32 +743,38 @@ srs_error_t SrsSipStack::do_parse_request(SrsSipRequest* req, const char* recv_m
...
+                        value = value + vec_device_status.at(i);
                     }
-              
-                    req->device_list_map[vec_device_id.at(i)] = status;
-                }
-                #endif
-                for(int i=0 ; i< (int)req->item_list.size(); i++){
-                    std::map<std::string, std::string> one_item = req->item_list.at(i);
-                    std::string status;
-                    if (one_item.find("Status") != one_item.end() && one_item.find("Name") != one_item.end()) {
```
PieerePi: As mentioned above, I have already improved this part on the develop branch.
The original interface only had id, status, and invite_status. I added name to facilitate displaying the name on the interface and performing operations. I also filtered out items without status.

On /api/v1/gb28181?action=sip_query_devicelist, I have created a comprehensive directory interface.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
PieerePi: Even if there are problems with the original code, modifications should be made on top of the existing code, whether it is for optimization or bug fixing. Moreover, before that, carefully consider whether it is the original code that has issues or if it is my code that has problems. Think more and understand it first.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: After your operation, follow or print with gdb, and you will understand. The return of get_channel is a temporary class instance. What's the use of updating its value? It is only useful if you copy it back.

`TRANS_BY_GPT3`

xbpeng121: In actual integration testing with the platform, it did indeed work, otherwise the video would not have been released. So here, it is confirmed that this update is effective.

Looking at the actual code analysis, "mc" is actually a member variable of the "muxer" object, and "muxer->copy_channel()" performs a copy of the "muxer" object's member variable "channel". So it seems that the above operation is just the channel copying itself, and the copy does not involve creating a new object, but rather just assigning the values of many member variables, remaining the same as the original object.

`TRANS_BY_GPT3`

xbpeng121: In addition, assuming that `channel` is indeed a temporary class instance, it is not clear why this operation can successfully copy and save it. `SrsGb28181StreamChannel mc = muxer->get_channel();`
mc.set_ssrc(ssrc);
muxer->copy_channel(&mc);
Can you please explain it? (The question may be a bit basic, but I really didn't understand it myself)

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
```diff
@@ -548,6 +549,30 @@ srs_error_t SrsSipStack::parse_xml(std::string xml_msg, std::map<std::string, st
     return srs_success;
 }
 
+srs_error_t SrsSipStack::parse_sdp(std::string sdp_msg, std::map<std::string, std::string> &json_map){
```
PieerePi: Last night, I thought it was pretty good about this, but I always felt something was off.

After thinking about it today, let me talk about this issue:

1. SDP should actually be "on-the-spot unified parsing", there is no need to use a map to store and parse different values such as y and a separately. The parsed values should be stored in variables for later use, instead of parsing them again when needed. Currently, for GB28181's SDP, only y needs to be parsed. When parsing a, do we need to put several a values into a map and then parse them? Therefore, the map is unnecessary, and it won't be needed for parsing other specific content in the future.

2. Here, the map connects the same types with commas, but commas are also part of the SDP content, for example...
a=fmtp:97 profile-level-id=42c01f;max-fs=3600;max-mbps=108000
a=fmtp:98 profile-level-id=42c00b;max-fs=240;max-mbps=3600
a=imageattr:97 send [x=1280,y=720] recv [x=1280,y=720]
a=imageattr:98 send [x=320,y=180] recv [x=320,y=180]
3. Actually, the name of this function is not parse_sdp, at most it should be store_sdp_to_map.
4. Additionally, when using a map, remember to always find the value before retrieving it, to maintain good programming habits.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -618,20 +619,20 @@ srs_error_t SrsGb28181SipService::on_udp_sip(string peer_ip, int peer_port,
...
-                }
                 send_ack(req, from, fromlen);
                 SrsGb28181Device *device = sip_session->get_device_info(req->sip_auth_id);
                 if (device){
                     device->invite_status = SrsGb28181SipSessionInviteOk;
                     device->req_inivate.copy(req);
                     device->invite_time = srs_get_system_time();
                 }
+                uint32_t ssrc_in_sdp = strtoul(srs_string_split(req->sdp_body_map["y"],",").at(0).c_str(), NULL, 10);//gb28181 protocol,sdp-type-y is the ssrc
+                std::string sess_in_sdp = srs_string_split(srs_string_split(req->sdp_body_map["o"],",").at(0)," ").at(0);
```
PieerePi: sess_in_sdp and chid_in_sdp, retrieve them from the SIP signaling, they already exist.
sip_session->session_id()+"@"+req->sip_auth_id

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: `SrsGb28181StreamChannel SrsGb28181RtmpMuxer::get_channel()
{
return *channel;
}`
What is returned is *channel, a temporary class instance. It has the same value as muxer's channel, but it is no longer related to muxer's channel. If you update this temporary class instance, muxer's channel will not change. You need to copy it back with muxer->copy_channel(&mc).

This is a relatively basic C++ usage question.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2459,10 +2458,12 @@ srs_error_t SrsGb28181Manger::start_ps_rtp_listen(std::string id, int port)
         return srs_error_wrap(err, "start rtp listen port is mux port"); 
     }
 
-    map<std::string, SrsGb28181RtmpMuxer*>::iterator key = rtmpmuxers.find(id);
-    if (key == rtmpmuxers.end()){
-       return srs_error_wrap(err, "start rtp listen port rtmp muxer is null"); 
-    }
+    /* delete by xbpeng 20201222 should not check rtmpmuxers, becasue it always not find*/
```
PieerePi, line 2442: I don't know why you commented out this.
Why can't it be found? There is no such line in my log at all?

I feel like you made a mistake somewhere.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.hpp
winlinvip: Great, I see it has already been fixed.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: Oh, I checked it out. This was resolved in PR #2014 and it's already in the develop branch.

`TRANS_BY_GPT3`

xbpeng121: The muxer here cannot be found for sure. You can take a look at the first few lines (lines 2519-2525) of the implementation code of the function `srs_error_t SrsGb28181Manger::create_stream_channel(SrsGb28181StreamChannel *channel)` in the file `srs_app_gb28181.cpp`.
`muxer = fetch_rtmpmuxer(id);
    if (muxer){
       SrsGb28181StreamChannel s = muxer->get_channel();
       channel->copy(&s);
       //return ERROR_GB28181_SESSION_IS_EXIST;
       return err;
    }`
Here, a check will be performed. If the muxer is obtained, it will directly return and the function `start_ps_rtp_listen()` will not be executed.

The creation of the muxer is done after the execution of `start_ps_rtp_listen()`. The specific line numbers are 2629 and 2643 (the exact code line positions may have slight differences, as I am referring to my merged code).
`if ((err = fetch_or_create_rtmpmuxer(id, &request, &muxer)) != srs_success)`
`rtmpmuxer_map_by_ssrc(muxer, ssrc);`


Take a look at the logic here and analyze whether I have misunderstood it.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: Hmm,
1- It is indeed inappropriate to separate with ",".
2- I do not agree with "should be resolved on the spot". The original design of the SrsSipRequest class already includes `SrsRtspSdp* sdp;`, but it was not implemented due to time constraints and the requirements at that time. A better approach here would be to assign a value to `SrsRtspSdp* sdp;` based on the protocol specifications of sdp. Storing it in a map is currently considered a simpler and more extensible method.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: Yes, it will be easier to retrieve from the session here, and I will update it later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
PieerePi: I see a difference, is this part of the code for non-fixed ports?
I have not used the features of non-fixed ports and automatic channel creation before.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
PieerePi: I know there is an sdp field inside, but I didn't use it because I think it's too heavy at the moment.
SrsRtspSdp::parse, SrsSdp::parse, can you check if they are both "in-place parsing"?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: Okay, okay, the information obtained from sip_query_devicelist is more complete. I will modify it back to the previous content later.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: Okay, okay, the information obtained from sip_query_devicelist is more complete. I will modify it back to the previous content later.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: Maybe I have a slightly different understanding of "on-site unified resolution" compared to yours. Shouldn't our thinking be consistent? It means that we need to resolve the SDP in one place, save it, and then directly retrieve the value in the subsequent places where it is needed.

`TRANS_BY_GPT3`

xbpeng121: The complete analysis of sdp here seems a bit complicated to me. Would it be more efficient to first use a map for storage and then make changes to this part when the map is no longer sufficient?
For the issue with field 'a', we choose a relatively uncommon character as a delimiter to avoid problems caused by connecting identical fields.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: Here, `SrsGb28181RtmpMuxer::get_channel()` returns the channel object of the muxer. Why is it not possible to directly assign a value to it using `setssrc`, but it works when using `SrsGb28181StreamChannel mc = muxer->get_channel();`?
mc.set_ssrc(ssrc);
Does assigning `muxer->copy_channel(&mc)` work in this case? It seems to me that they are both operating on the same object.
image
image
I believe their effects are the same.


Here, it is also possible to accept modifications to your code. I believe there won't be much difference in actual execution.

`TRANS_BY_GPT3`

xbpeng121:
@PieerePi I have made changes to the relevant content, please take some time to review and see if there are any areas that still need improvement. Make sure to maintain the markdown structure.
ThankYou :-)

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2367,10 +2345,12 @@ void SrsGb28181Manger::update_rtmpmuxer_to_newssrc_by_id(std::string id, uint32_
     SrsGb28181RtmpMuxer* muxer = NULL;
 
     if (rtmpmuxers.find(id) == rtmpmuxers.end()) {
+        srs_warn("gb28181: at update_rtmpmuxer_to_newssrc_by_id() client_id not found. client_id=%s",id.c_str());
         return;
     }
 
     muxer = rtmpmuxers[id];
+    //TODO: find out whether muxer->get_channel().set_ssrc(new_ssrc) is different with the under codes
```
PieerePi: Delete this comment.
You will understand if you debug it with gdb. Need to consolidate C++ knowledge.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2382,6 +2362,7 @@ void SrsGb28181Manger::update_rtmpmuxer_to_newssrc_by_id(std::string id, uint32_
     mc.set_ssrc(ssrc);
     muxer->copy_channel(&mc);
     rtmpmuxer_map_by_ssrc(muxer, ssrc);
+    srs_trace("gb28181: update_rtmpmuxer_to_newssrc_by_id success. client_id=%s, new_ssrc=%#x, old_ssrc=%#x",id.c_str(),ssrc,old_ssrc);
```
PieerePi: Delete this print
There is already a print above.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.hpp
```diff
@@ -108,6 +108,7 @@ class SrsGb28181Conn;
 class SrsGb28181Caster;
 
 //ps rtp header packet parse
+//ps doc:
```
PieerePi: Delete this comment.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
PieerePi: To be honest, nothing can convince me.
This is only useful for retrieving y, but the analysis of y is already available.
Analyzing a\m\c, on the other hand, would be cumbersome.

You have also seen code elsewhere that does not use a map to store sdp, and sdp is uniformly parsed in place (as soon as it is received) and then used.
Instead of storing it in a map and parsing it when needed, resulting in multiple parsing instances if used multiple times.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -68,10 +68,11 @@ std::string srs_get_sip_session_status_str(SrsGb28181SipSessionStatusType status
 SrsGb28181Device::SrsGb28181Device()
 {
     device_id = ""; 
-    device_name = "";
     invite_status = SrsGb28181SipSessionUnkonw;
     invite_time = 0;
     device_status = "";
+    device_name = "";
```
PieerePi: Go back to the original.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
```diff
@@ -559,6 +560,7 @@ srs_error_t SrsGb28181SipService::on_udp_sip(string peer_ip, int peer_port,
        
         //reponse status 
         if (req->cmdtype == SrsSipCmdRequest){
+            send_status(req, from, fromlen);
```
PieerePi: Repeated? Go back to the original.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.hpp
```diff
@@ -59,8 +59,8 @@ class SrsGb28181Device
     virtual ~SrsGb28181Device();
 public:
     std::string device_id;
-    std::string device_name;
     std::string device_status;
+    std::string device_name;
```
PieerePi: Go back to the original.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.hpp
```diff
@@ -103,7 +103,7 @@ extern int srs_string_count(std::string str, std::string flag);
 // Find the min match in str for flags.
 extern std::string srs_string_min_match(std::string str, std::vector<std::string> flags);
 // Split the string by flag to array.
-extern std::vector<std::string> srs_string_split(std::string str, std::string flag);
+extern std::vector<std::string> srs_string_split(std::string s, std::string seperator);
```
PieerePi, line 106: 你这儿有两个声明？
而且声明好像有问题？

trunk/src/protocol/srs_sip_stack.cpp
xbpeng121: Can you take some time to write the "in-place unified parsing" of SDP? I am not familiar with the relevant protocols and currently don't have time to study the details, but I think it's not ideal to parse Y when using Y. It would be better to avoid doing too much work when using the map for storage.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.hpp
xbpeng121: Modify and submit later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: Modify and submit later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181_sip.cpp
xbpeng121: Modify and submit later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.hpp
xbpeng121: Modify and submit later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: Modify and submit later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_gb28181.cpp
xbpeng121: Already deleted, will follow up with relevant supporting materials.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_sip_stack.cpp
PieerePi: In the GB28181 on-demand function section, the information of c/m/a in SDP is not very useful. It is sufficient to parse the y and PS streams.
The PS stream contains audio and video information, so the m/a information is not very useful. The y information contains the SSRC of the PS stream, so the c/m information is not very useful. SRS distinguishes streams based on SSRC.
There is also an f for audio and video information, but many manufacturers do not provide it.
Whether to use UDP/TCP for receiving streams in m is determined through configuration. SRS is always the caller, so it directly decides. The camera cannot negotiate.

So, in this part of GB28181, there is not much use in completing a comprehensive SDP parsing.

`TRANS_BY_GPT3`

xbpeng121: Oh, okay. In that case, it seems like the other information in the sdp doesn't have much effect. I will modify the code.

`TRANS_BY_GPT3`

xbpeng121:
all done, you can merge now

`TRANS_BY_GPT3`

winlinvip:
> all done, ready to merge

@PieerePi Please confirm if there are no issues.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
```diff
@@ -480,31 +480,22 @@ int srs_string_count(string str, string flag)
...
-    if (!s.empty()) {
-        arr.push_back(s);
-    }
-    
-    return arr;
-}
+
+vector<string> srs_string_split(string s, string seperator) {
+     vector<string> result;
+     unsigned int posBegin = 0;
```
PieerePi: Change posBegin to size_t type.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
```diff
@@ -480,31 +480,22 @@ int srs_string_count(string str, string flag)
...
+     size_t posSeperator = s.find(seperator);
+ 
+     while (posSeperator != string::npos) {
+         result.push_back(s.substr(posBegin, posSeperator - posBegin));
+         posBegin = posSeperator + seperator.size(); // next byte of seperator
```
PieerePi: seperator.size()' should be changed to 'seperator.length()', just like below.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
```diff
@@ -480,31 +480,22 @@ int srs_string_count(string str, string flag)
...
-    }
-    
-    if (!s.empty()) {
-        arr.push_back(s);
-    }
-    
-    return arr;
-}
+
+vector<string> srs_string_split(string s, string seperator) {
```
PieerePi, line 484: The new srs_string_split seems to not handle the case where the separator is empty (the original code does handle it).
Note that string::find("") returns 0, not string::npos.

`TRANS_BY_GPT3`

PieerePi:
> > all done, ready to merge
> 
> @PieerePi Please confirm if there are no issues.

After reviewing the code again, I found that there are still some issues with srs_string_split.

I haven't used or tested the three scenarios in GB28181: auto_create_channel set to on, auto_play set to on, and invite_port_fixed set to off. Xbppeng121 has done some code merging and setting completion, and upon reviewing, it seems fine to me. However, we should test it after merging.

@xbpeng121, please modify the commit message.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
winlinvip: I will supplement UTest for this core function to verify if it meets the expectations, which I missed before.
Let's put this issue aside for now. After merging, I will add UTest and make any necessary changes if there are any issues.

`TRANS_BY_GPT3`

winlinvip:
> ，

srs_string_split这个先放放，我补充下UTest后再一起看下。
如果其他没有问题我就先合并了。

另外，麻烦加下我微信号吧，官网有二维码：https://ossrs.net

winlinvip:
> > > all done, ready to merge
> > 
> > 
> > @PieerePi Please confirm if there are no issues.
> 
> I reviewed the code again, and there are still some issues with `srs_string_split`.
> 
> I haven't used or tested the three scenarios in GB28181 where `auto_create_channel` is set to "on", `auto_play` is set to "on", and `invite_port_fixed` is set to "off". xbppeng121 has done some code merging and setting completion, and I have only reviewed it and it looks fine. We need to test it after merging.
> 
> @xbpeng121, please modify the commit message.

@PieerePi 

srs_string_split, let's put it aside for now. I will supplement it with UTest and then we can review it together.
If there are no other issues, I will merge it first.

Additionally, please add my WeChat ID. You can find the QR code on the official website: [https://ossrs.net](https://ossrs.net)

`TRANS_BY_GPT3`

xbpeng121:
change log
1- Fix the bug in srs_string_split():
   Before fix: srs_string_split("1,A,,address", ",") = ["1", "A", "address", ""]
   After fix: srs_string_split2("1,A,,address", ",") = ["1", "A", "", "address"]
2- Remove duplicate code in on_rtp_packet() in srs_app_gb28181.cpp and replace it with the fetch_rtmpmuxer() function.
3- Rename the create_rtmpmuxer() function to fetch_rtmpmuxer() because if the muxer already exists in the function, it will be fetched instead of creating a new one.
4- Change a specific srs_trace output to srs_info because in the actual integration platform, this output is very frequent and affects the information viewing.
5- Fix the bug where the sip_invite_port_fixed parameter does not take effect.

@PieerePi @winlinvip 

`TRANS_BY_GPT3`

PieerePi:
Okay, I don't see any problem.

`TRANS_BY_GPT3`

winlinvip:
Merged as https://github.com/ossrs/srs/commit/47422b781937f3f16e5734a7d9977856ec603577

PR #2138, opened 2021-01-10
Fix memory leak in 3.0 which was fixed in 4.0 already

digitalm4:
It had already been fixed earlier by someone else in develop branch.
Thank you.


digitalm8:
К сожалению остались еще другие утечки. Этот патч исправляет проблему не полностью. В моем случае скорость расходования памяти сократилась в 2-3 раза.

digitalm8:
Unfortunately, there are still other leaks. This patch does not completely fix the problem. In my case, the speed of memory consumption was reduced by 2-3 times.

PR #2156, opened 2021-01-19
h265: fix some important bugs

PieerePi:
1. optimize when to send hevc vps sps pps header
2. fix hevc key frame dropped by rtc before sending to consumer
3. fix overread in nalu parsing
4. ignore unknown payload data, otherwise it will be processed in other place


runner365 (approved):

trunk/src/app/srs_app_source.cpp
```diff
@@ -2362,7 +2362,9 @@ srs_error_t SrsSource::on_video_imp(SrsSharedPtrMessage* msg)
 
     // For bridger to consume the message.
     if (bridger && (err = bridger->on_video(msg)) != srs_success) {
-        return srs_error_wrap(err, "bridger consume video");
+        // rtc doesn't support hevc, here fails if it's a hevc key frame
+        // so consumers can't get 1c01 VideoTag
+        // return srs_error_wrap(err, "bridger consume video");
```
winlinvip, line 2367: This 'err' is an object, if not handled, it will leak. The handling method is as follows:

```
srs_warn("rtc on video err %s", srs_error_desc(err).c_str());
srs_freep(err);
```

After printing the error, release the error.

* Printing is optional. If you print, the benefit is that no information is missed, but it can result in more warning logs.
* If you don't print, then this error will be lost, meaning there will be no log indicating the problem occurred here. If you are sure there is no problem, you can choose not to print.

`TRANS_BY_GPT3`

PR #2168, opened 2021-01-25
GB28181: G711 support and new RTC transcoder

PieerePi:
G711 support and new RTC transcoder
1. GB28181: Transcode G711 to AAC before sending to SRS by RTMP. You can disable it by undefining GB28181_TRANSCODE_G711_TO_AAC in srs_app_gb28181.hpp.
2. WebRTC: Transcode G711 to OPUS.
ffmpeg -re -i test.flv -acodec pcm_mulaw -ac 1 -ar 8000 -vcodec copy -f flv -y rtmp://xx.xx.xx.xx/live/test_stream
3. WebRTC: New RTC transcoder; Support changing audio codec of the same URL.
4. New added PCM encoder and decoder files are from ffmpeg 4.2.2, because I don't know the exact version of ffmpeg-4.2-fit. Please delete objs/Platform-Linux-XXX directory and reconfigure your project to rebuild ffmpeg.

### DIFF
--- a/trunk/3rdparty/ffmpeg-4.2-fit/libavcodec/codec_list.c
+++ b/trunk/3rdparty/ffmpeg-4.2-fit/libavcodec/codec_list.c
@@ -1,9 +1,13 @@
 static const AVCodec * const codec_list[] = {
     &ff_aac_encoder,
     &ff_opus_encoder,
+    &ff_pcm_alaw_encoder,
+    &ff_pcm_mulaw_encoder,
     &ff_libopus_encoder,
     &ff_aac_decoder,
     &ff_aac_fixed_decoder,
     &ff_aac_latm_decoder,
+    &ff_pcm_alaw_decoder,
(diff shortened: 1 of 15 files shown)


duiniuluantanqin:
The G711 format is mainly used in security cameras.

1. Use RTMP playback
   In most cases, there is no strict requirement for G711 encoding. It is better to directly change the audio output format of the IPC to AAC, which eliminates the performance consumption of transcoding.

2. Use WebRTC playback
   WebRTC does not support AAC, but it supports G711. So,
    
2.1 For AAC format, it will be transcoded into Opus.
2.2 For G711, it will be directly passed through, providing better performance. Currently, there is a G711 branch that has implemented the functionality of passing through G711.

So I suggest that we don't need to merge this PR for now.

`TRANS_BY_GPT3`

winlinvip:
Not merging for now.

`TRANS_BY_GPT3`

PR #2186, opened 2021-02-04
fixed frame loss when time_jitter is off

NodeBoy2:
**描述(Description)**

1. SRS版本(Version): `3.0release分支`
1. SRS的日志如下(Log):
```
shrinking, size=2, removed=2, max=30000ms
dispatch cached gop success. count=49, duration=-1
create consumer, active=1, queue_size=0.00, jitter=30000000
```
1. SRS的配置如下(Config):
```
vhost __defaultVhost__ {
    time_jitter     off;
}
```

**重现(Replay)**

> 重现Bug的步骤(How to replay bug?)

1. 配置time_jitter  off;
2. 推流
3. 30s后，使用ffplay播放，前2s有明显的音画不同步，画面播放慢的现象。

**问题**
1. 有消费者时，srs会发送：metadata->audio sequence header->video sequence header->gop cache->...
2. sequence header时间戳通常为0。在关闭时间校准后，在gop cache放入SrsMessageQueue时，判断队列中的duration已经超过30s了，就会丢弃第一个帧（通常为视频关键帧）。

**修复**
在SrsMessageQueue计算队列中数据累计时长时，排查sequence header

### DIFF
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -269,7 +269,9 @@ srs_error_t SrsMessageQueue::enqueue(SrsSharedPtrMessage* msg, bool* is_overflow
 {
     srs_error_t err = srs_success;
     
-    if (msg->is_av()) {
+    // maybe the SH timestamp much short than A/V packet,
+    // shrink does not contain SH
+    if ((msg->is_audio() && !SrsFlvAudio::sh(msg->payload, msg->size)) || (msg->is_video() && !SrsFlvVideo::sh(msg->payload, msg->size))) {
         if (av_start_time == -1) {
             av_start_time = srs_utime_t(msg->timestamp * SRS_UTIME_MILLISECONDS);
         }


winlinvip:
After streaming for a certain period of time, there will be this log when playing:

```bash
[2021-10-28 07:22:09.511][Trace][91550][0eq5u805] shrinking, size=2, removed=2, max=30000ms
[2021-10-28 07:22:09.511][Trace][91550][0eq5u805] dispatch cached gop success. count=458, duration=-1
```

Capturing packets, it was found that there is a SequenceHeader.

image

After reviewing the code, if the data that is shrunk contains a SequenceHeader, it will be retrieved again.

```cpp
void SrsMessageQueue::shrink()
{
    .............

    // update av_start_time
    av_start_time = av_end_time;
    //push_back secquence header and update timestamp
    if (video_sh) {
        video_sh->timestamp = srsu2ms(av_end_time);
        msgs.push_back(video_sh);
    }
    if (audio_sh) {
        audio_sh->timestamp = srsu2ms(av_end_time);
        msgs.push_back(audio_sh);
    }
    
    if (!_ignore_shrink) {
        srs_trace("shrinking, size=%d, removed=%d, max=%dms", (int)msgs.size(), msgs_size - (int)msgs.size(), srsu2msi(max_queue_size));
    }
```

The SequenceHeader will be preserved, but the keyframes will indeed be discarded. So in the image above, the first one is the SequenceHeader, and the second one is a P-frame.

The simplest solution is to just check if it is 0, because if other packets are also 0 but not a sequence header, it will still cause issues.

Of course, the best approach is to disable shrink during playback.

`TRANS_BY_GPT3`

PR #2261, opened 2021-03-27
Update srs_app_edge.cpp

bluestn:
fix: close the dangling push connection while upstream is shutdown.

### DIFF
--- a/trunk/src/app/srs_app_edge.cpp
+++ b/trunk/src/app/srs_app_edge.cpp
@@ -622,7 +622,13 @@ srs_error_t SrsEdgeForwarder::do_cycle()
         
         // sendout messages, all messages are freed by send_and_free_messages().
         if ((err = sdk->send_and_free_messages(msgs.msgs, count)) != srs_success) {
-            return srs_error_wrap(err, "send messages");
+            // return srs_error_wrap(err, "send messages");
+            // problem: return error will let the push connection still be alive although upstream connection already was shutdown.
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_edge.cpp
```diff
@@ -622,7 +622,13 @@ srs_error_t SrsEdgeForwarder::do_cycle()
...
         // sendout messages, all messages are freed by send_and_free_messages().
         if ((err = sdk->send_and_free_messages(msgs.msgs, count)) != srs_success) {
-            return srs_error_wrap(err, "send messages");
+            // return srs_error_wrap(err, "send messages");
+            // problem: return error will let the push connection still be alive although upstream connection already was shutdown.
+            // fix: set the error for this thread, and while SrsEdgeForwarder::proxy is called later, it will check the error, 
+            // if the error is found, push connection will be closed 
+            srs_error("edge push send message to upstream failed. err=%s", srs_error_desc(err).c_str());
+            send_error_code = srs_error_code(err);
+            srs_error_reset(err);
```
winlinvip, line 631: The current plan is that if the upstream connection is disconnected, we will not disconnect the streaming connection. This way, when the upstream is upgraded or restarted, the streaming or playback clients will not be disconnected.

Of course, if the upstream says that the stream already exists, then indeed an error should be returned and the publish streaming should be disconnected. Because in this case, retrying the streaming will continue to result in errors, as the stream already exists.

So here, we cannot handle it directly like this, we need to determine what the error is.

`TRANS_BY_GPT3`

winlinvip:

Issue #2181, opened 2021-02-01
DVR：从API获取某个流的复杂的录制规则，dvr_apply作为固定的规则


houxuexianhan:
**描述(Description)**

>  dvr_apply不支持通配符。当配置了dvr_apply为某个流之后，自定义输出的文件保存路径不生效。 

1. SRS版本(Version): `v3.0.155`
1. SRS的日志如下(Log):
```
xxxxxxxxxxxx
```
1. SRS的配置如下(Config):
```
 dvr{
  	enabled         on;
	dvr_apply	dvr/a
	dvr_path	/var/www/html/[app]/[stream].[2006][01][02][15][04][05][999].mp4;
	dvr_plan	session;
	dvr_duration	30;
	dvr_wait_keyframe       on;
	time_jitter             full; 
    }

```

**期望行为(Expect)**

> 希望dvr_apply支持类似于  live/* 这种通配符的配置
>当配置了dvr_apply为某个流之后，录制的文件保存路径需要按dvr_path生效

winlinvip:
DVR能力会增强，欢迎PR。

robertluoxu:
I hope the DVR path can support online OSS storage.

`TRANS_BY_GPT3`

winlinvip:
Regarding supporting cloud storage, refer to #1193.

> Hope the DVR path can support online OSS storage.

`TRANS_BY_GPT3`

winlinvip:
Previously, it was accomplished through configuring dvr_apply, but this method cannot handle complex matching and is only suitable for fixed rule matching, which is prone to problems [#2282](https://github.com/ossrs/srs/pull/2282/files#r698879205).

For complex matching, such as wildcard or dynamic adjustment needs, it is better to call the HTTP API to determine whether to record a certain stream, for example:

```
dvr_filter http://xxx
```

> Note: Please note that this `dvr_filter` needs to be implemented by your own service, not by SRS.

By requesting this API, it returns whether to record this stream and some necessary configurations.

This way, you don't have to rely on SRS to perform complex recording logic.

`TRANS_BY_GPT3`

lxge:
> dvr_filter

Hello, may I ask how to dynamically configure 4.0, adjust the recording time period? After calling the raw API and getting a return code of 0, there is no actual effect.

`TRANS_BY_GPT3`

guoronghui:
> >

4.0 no longer supports RAW API recording, right?

`TRANS_BY_GPT3`

[...]

PR #2282, opened 2021-04-13
bugfix: SrsAsyncCallWorker::cycle to crash

bluestn:
```
    dvr {
        # the filter for dvr to apply to.
        #       all, dvr all streams of all apps.
        #       <app>/<stream>, apply to specified stream of app.
        # for example, to dvr the following two streams:
        #       live/stream1 live/stream2
        # default: all
        dvr_apply       all;
    }
```

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -279,7 +279,13 @@ bool srs_config_apply_filter(SrsConfDirective* dvr_apply, SrsRequest* req)
 {
     static bool DEFAULT = true;
     
-    if (!dvr_apply || dvr_apply->args.empty()) {
+    // if dvr_apply is null, return false to keep consistency with get_dvr_enabled() function
+    // if not, will occasionally cause SrsAsyncCallWorker::cycle to crash down with trd == 0x00
+    if ( !dvr_apply ){
+        return false;
+    }
+    
+    if ( dvr_apply->args.empty()) {
         return DEFAULT;
     }
     


srgt305:
leave a

trunk/src/app/srs_app_config.cpp
```diff
@@ -279,7 +279,13 @@ bool srs_config_apply_filter(SrsConfDirective* dvr_apply, SrsRequest* req)
 {
     static bool DEFAULT = true;
     
-    if (!dvr_apply || dvr_apply->args.empty()) {
+    // if dvr_apply is null, return false to keep consistency with get_dvr_enabled() function
+    // if not, will occasionally cause SrsAsyncCallWorker::cycle to crash down with trd == 0x00
+    if ( !dvr_apply ){
```
winlinvip, line 284: dvr_apply为NULL，就是没有配置，例如：

```
    dvr {
        enabled         on;
```

这时候如果返回false，会导致DVR被禁用。所以不能返回false。

当然得看下为何会导致crash。

trunk/src/app/srs_app_config.cpp
```diff
@@ -279,7 +279,13 @@ bool srs_config_apply_filter(SrsConfDirective* dvr_apply, SrsRequest* req)
...
+        return false;
+    }
+    
+    if ( dvr_apply->args.empty()) {
```
winlinvip, line 288: I checked the situation where `trd` is NULL in `SrsAsyncCallWorker::cycle`, and it only occurs when DvrPlan is being deleted.

```
SrsDvrPlan::~SrsDvrPlan()
{
    srs_freep(async);
}

SrsAsyncCallWorker::~SrsAsyncCallWorker()
{
    srs_freep(trd);
}
```

When DvrPlan stops, it will have a stop and clean task, which will cause a switch.

```
void SrsDvrPlan::on_unpublish()
{
    async->stop();
}

void SrsAsyncCallWorker::stop()
{
    flush_tasks();
    srs_cond_signal(wait);
    trd->stop();
}
```

In combination with the reload of `dvr_apply`, the above logic will be triggered frequently.

```
srs_error_t SrsDvr::on_reload_vhost_dvr_apply(string vhost)
{
    SrsConfDirective* conf = _srs_config->get_dvr_apply(req->vhost);
    bool v = srs_config_apply_filter(conf, req);
    
    on_unpublish();
    
    if ((err = on_publish()) != srs_success) {
        return srs_error_wrap(err, "on publish");
    }
}
```

So the problem should be in this place. To be determined specifically.

`TRANS_BY_GPT3`

winlinvip:
A more suitable solution:

1. `dvr_apply` does not support reload, so we can support it through the HTTP API by recording different streams.
2. Move `SrsDvrPlan.async` to the global scope to avoid each Plan having its own asynchronous worker, thus preventing frequent opening and closing of workers.

I mentioned a new issue: #2181.

`TRANS_BY_GPT3`

winlinvip:
Still need to thank @bluestn for the Pull Request.

`TRANS_BY_GPT3`

PR #2284, opened 2021-04-14
srs rejects request while downstream server disconnects and then reconnects in a short time.

bluestn:
bugfix: while downstream server disconnects and reconnects in a short time, srs rejects the push request because SrsForwarder::cycle is sleeping for SRS_FORWARDER_CIMS, and the stream is still busy.

### DIFF
--- a/trunk/src/app/srs_app_forward.cpp
+++ b/trunk/src/app/srs_app_forward.cpp
@@ -179,24 +179,31 @@ srs_error_t SrsForwarder::on_video(SrsSharedPtrMessage* shared_video)
 srs_error_t SrsForwarder::cycle()
 {
     srs_error_t err = srs_success;
-    
+
     while (true) {
-        // We always check status first.
-        // @see https://github.com/ossrs/srs/issues/1634#issuecomment-597571561
-        if ((err = trd->pull()) != srs_success) {
-            return srs_error_wrap(err, "forwarder");
-        }
 
         if ((err = do_cycle()) != srs_success) {
             srs_warn("Forwarder: Ignore error, %s", srs_error_desc(err).c_str());
             srs_freep(err);
         }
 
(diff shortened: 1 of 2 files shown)


winlinvip:
Hmm, this critical situation indeed deserves careful handling.
In addition to the Forwarder, the Edge might also encounter this situation, so we need to consider it together.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_forward.cpp
```diff
@@ -179,25 +179,32 @@ srs_error_t SrsForwarder::on_video(SrsSharedPtrMessage* shared_video)
...
             srs_freep(err);
         }
 
+        //
+        // 将trd->pull()检测移到do_cycle之后，以便当正常的下游服务器断开连接的时候，不用再
+        // srs_usleep(SRS_FORWARDER_CIMS)，避免下游服务器马上重连的时候失败的情况发生。
+        //
+        // We always check status first.
+        // @see https://github.com/ossrs/srs/issues/1634#issuecomment-597571561
+        if ((err = trd->pull()) != srs_success) {
```
winlinvip, line 196: Thank you very much 👍

It would be more appropriate to add error detection before and after the Forward cycle.

I have already merged this place: [link](https://github.com/ossrs/srs/commit/8125dd4dd1f639d76a282c566556efb271a994e8)

`TRANS_BY_GPT3`

duiniuluantanqin:
> Well, this critical situation indeed deserves careful handling. Besides the Forwarder, the Edge might also have this situation, so we need to consider it together.


Edge should also be needed. Otherwise, it will keep retrying. @winlinvip
https://github.com/ossrs/srs/pull/2707#issuecomment-959079140

`TRANS_BY_GPT3`

PR #2315, opened 2021-04-26
Support multiple threads for performance

winlinvip:
For this PR:

1. ST: Simplify it, only Linux/Darwin, epoll/kqueue, single process
2. ST: Support thread-local state-threads
3. Threads: Support Hybrid+API+LOG threads

Why it's so important?

1. RTC server is not only IO server, it's IO+CPU server, so we must refine the arch for RTC.
2. Multiple Isolated Threads, is better than multiple processes arch, especially the API thread need to communicate with hybrid thread.
3. We must use a dedicated thread for disk writing, such as LOG or DVR threads.

What's now and next?

1. This PR implements a basic Multiple Isolated Threads arch for RTC.
2. However, default to one thread for stability and reliability, which similar to single thread arch.
3. Need to fix bugs for RTMP multiple threads.
4. Need to support API to collect data from all hybrids.
5. It's possible to start more threads for FFmpeg or MCU in the future.


winlinvip:
Closed for https://github.com/ossrs/srs/issues/2188#issue-801759588

PR #2344, opened 2021-05-10
Fix(log): fix compile error when the options --log-info, --log-verbose are enabled

stitch8467:
I got errors when I tried to do following commands:

```
cd trunk
./configure --log-info --log-verbose && make
```

After checking, there are few parts need to be corrected. Please help to review, thanks.


winlinvip:
4.0 The latest code is now problem-free.

However, I still want to express my gratitude to @stitch8467 for submitting the PR. Thank you 😀

`TRANS_BY_GPT3`

Issue #2371, opened 2021-05-24
rtmp推流webrtc播放，Chrome正常，Firefox丢包卡顿


duiniuluantanqin:
**描述(Description)**

> rtmp推流，webrtc播放，Chrome正常，Firefox丢包卡顿

1. SRS版本(Version): `XCORE-SRS/5.0.2(Leo)`

1. SRS的配置如下(Config):
```
rtc.conf
```

**重现(Replay)**

> 重现Bug的步骤(How to replay bug?)

1. `用ffmpeg推送rtmp`
2. `分别用Chrome和Firefox通过webrtc协议拉流`
3. `即可看到Firefox视频卡顿`
4. `分析抓包数据，Firefox有丢包`

**附件为Wireshark抓包数据**
[【Wireshark抓包】Firefox_pt_126_卡顿和Chrome_pt_125_正常.zip](https://github.com/ossrs/srs/files/6529465/Firefox_pt_126_.Chrome_pt_125_.zip)

![RTP包分析](https://user-images.githubusercontent.com/4254164/119288658-9151f980-bc7b-11eb-9cb0-3b5779266482.png)

duiniuluantanqin:
#### Reasons
1. Chrome **enables** nack by default, while Firefox **disables** nack by default.
2. The answer sdp from SRS does not contain the `rtcp-fb nack` field.

#### Conclusion
During network jitter, Firefox's lack of nack transmission leads to video stuttering.

#### Solution
To resolve this issue, generate an answer sdp based on the offer sdp and add the `rtcp-fb nack` field.


`TRANS_BY_GPT3`

duiniuluantanqin:
https://github.com/ossrs/srs/pull/2373#issue-651149318

winlinvip:
Chrome 

offer
```
v=0
o=- 3580466860054219675 2 IN IP4 127.0.0.1
s=-
t=0 0
a=group:BUNDLE 0 1
a=extmap-allow-mixed
a=msid-semantic: WMS
m=audio 9 UDP/TLS/RTP/SAVPF 111 103 104 9 0 8 106 105 13 110 112 113 126
c=IN IP4 0.0.0.0
a=rtcp:9 IN IP4 0.0.0.0
a=ice-ufrag:Xu5z
a=ice-pwd:Gu3bx4vsq/pm0C9ieK1p3y34
a=ice-options:trickle
[...]

PR #2373, opened 2021-05-24
RTC: Fix NACK negotiation bug for Firefox.

duiniuluantanqin:
#### 原因
1. SRS回复的answer sdp里面，没有rtcp-fb nack字段。
1. Chrome默认**开启**nack，所以尽管SRS的answer没有开启NACK，也没问题。
1. Firefox默认**关闭**nack，所以SRS的answer没有开启NACK，就会有丢包导致的卡顿。

#### 结论
   网络抖动时，由于Firefox不会发送nack，造成视频卡顿。

#### 解决
   增加rtcp-fb nack字段，相当于手动开启firefox的nack功能，解决丢包造成的卡顿。

@see https://github.com/ossrs/srs/issues/2371

### DIFF
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -3099,6 +3099,10 @@ srs_error_t SrsRtcConnection::negotiate_play_capability(SrsRtcUserConfig* ruc, s
             // We should clear the extmaps of source(publisher).
             // @see https://github.com/ossrs/srs/issues/2370
             track->extmaps_.clear();
+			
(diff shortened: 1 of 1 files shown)


winlinvip:
I confirmed that this is indeed a bug and can be reproduced.

I need to supplement regression testing to prevent similar issues from occurring again in the future.

`TRANS_BY_GPT3`

winlinvip:
Scenarios to be added for regression testing:

1. If there is a nack in the client's offer and the SRS configuration enables nack, then the answer must have a nack.
2. If there is no nack in the client's offer or the SRS configuration disables nack, then the answer must not have a nack.

`TRANS_BY_GPT3`

PR #2374, opened 2021-05-24
fix bug:appears CLOSE_WAIT when counts of client connections exceeds …

lbli:
http-flv拉流，当拉流的客户端数量大于SRS服务器配置文件里的max_connections时，客户端挂断后，会出现CLOSE_WAIT，出现的CLOESE_WAIT数量是客户端连接数减去max_connections，以上修改解决的是上述情况服务器出现CLOSE_WAIT的情况

### DIFF
--- a/trunk/src/app/srs_app_server.cpp
+++ b/trunk/src/app/srs_app_server.cpp
@@ -1570,7 +1570,7 @@ srs_error_t SrsServer::accept_client(SrsListenerType type, srs_netfd_t stfd)
     ISrsStartableConneciton* conn = NULL;
     
     if ((err = fd_to_resource(type, stfd, &conn)) != srs_success) {
-        if (srs_error_code(err) == ERROR_SOCKET_GET_PEER_IP && _srs_config->empty_ip_ok()) {
+        if ((srs_error_code(err) == ERROR_SOCKET_GET_PEER_IP || srs_error_code(err) == ERROR_EXCEED_CONNECTIONS) && _srs_config->empty_ip_ok()) {
             srs_close_stfd(stfd); srs_error_reset(err);
             return srs_success;
         }


winlinvip:
It seems that when exceeding the connection limit, the error is hidden and the client is disconnected normally.
1. If the maximum connection limit set by the system is exceeded, an error should be returned instead of success.
2. What are the effects of returning an error? Will it print an error log or cause other problems?
3. It is essential to print an error log when the client cannot connect due to system overload. Why hide the error?

I didn't understand what problems there might be and what problems this PR solves.

`TRANS_BY_GPT3`

lbli:
> It seems that when exceeding the connection limit, the error is hidden and the client is disconnected normally.
> 
> 1. If the maximum connection limit set by the system is exceeded, an error should be returned, why is it returning success?
> 
> 2. What are the impacts when returning an error? Does it print error logs or cause any other issues?
> 
> 3. It is essential to print error logs when the client cannot connect due to system overload. Why would we want to hide the error?
> 
> 
> I didn't understand what issues this PR would cause and what problem it solves.



Problem Description:
1: In SRS version 4.0.122, I set max_connections to 100.
   Simulating 200 client http-flv pull streams: ./sb_http_load -c 200 -r http://172.24.0.74:8080/live/livestream.flv
   There will be 200 ESTABLISHED connections on the SRS server.
   SRS log: accept client failed, err is code=1074 : fd to resource : drop fd=301, ip=172.24.0.75:60020, max=100, cur=100 for exceed connection limits
   After killing the above sb_http_load command, there will be 100 CLOSE_WAIT on the SRS server. If I continue to pull streams with 200 clients, CLOSE_WAIT will continue to accumulate by 100 with each execution.

2:
   I verified on SRS 2.0 release version, and there were no CLOSE_WAIT occurrences. The number of ESTABLISHED connections simulated by the terminal is consistent with max_connections. After killing sb_http_load, no CLOSE_WAIT is generated.

PR Description:
The PR aims to solve the CLOSE_WAIT issue in version 4.0.
When the maximum connection limit of the system is exceeded, an error should be returned, and the connection should be closed (by executing srs_close_stfd) to ensure proper handling.
     
In order to return an error and disconnect the link, I have placed ERROR_EXCEED_CONNECTIONS (1074) within the if statement of the PR.

In conclusion, based on the test: version 4.0.122 cannot guarantee the maximum connection limit, and when the number of client connections exceeds the maximum limit, accumulation of CLOSE_WAIT occurs after the client disconnects.




`TRANS_BY_GPT3`

winlinvip:
Hiding errors is not an appropriate practice.

`TRANS_BY_GPT3`

winlinvip:
max_connections' does not precisely control the number of connections. If precise control is required, HTTP callback can be used.

Still, thank you very much for submitting the PR. 👍 @lbli

`TRANS_BY_GPT3`

PR #2384, opened 2021-05-28
add the g711a and g711u flv demuxer for supporting g711a and g711u in webrtc later

qichaoshen82:
In Video surveillance, many device's audio codec is g711a and g711u, these two audio codecs are supported in Chrome webrtc audio codec list, so we and support them without audio transcoding.
First, add the g711a and g711u flv demuxer, I will and other implement later.


qichaoshen82:
use wrong branch

PR #2404, opened 2021-06-08
当在线人数超过60s为0时自动清理国标视频流

Cherisher:
国标直播流无在线观看超过一定时间后自动通知GB28181发送BYE指令，停止推流，达到降低网络拥塞的目的

### DIFF
--- a/trunk/src/app/srs_app_statistic.cpp
+++ b/trunk/src/app/srs_app_statistic.cpp
@@ -1,8 +1,25 @@
-//
-// Copyright (c) 2013-2021 Winlin
-//
-// SPDX-License-Identifier: MIT
-//
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2013-2021 Winlin
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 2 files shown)


duiniuluantanqin (changes requested):
在SrsStatistic放个timer，破坏了纯粹性了，逻辑有点混乱，不太可取。

Cherisher:
The purpose of idle detection is to achieve on-demand push of video streams. When there is no business system, it is convenient for SRS to actively disconnect the push stream to save traffic (money). After all, the bit rate of national standards is usually very high, generally defaulting to 10M. For example, the bit rate of some panoramic cameras can easily reach 2, 30M. Of course, this can be ignored when not considering operating costs.

`TRANS_BY_GPT3`

duiniuluantanqin:
> The purpose of idle detection is to achieve on-demand push of video streams. When there is no business system, it is convenient for SRS to actively disconnect the push stream to save traffic (and money). After all, the bit rate of national standards is usually very high, generally defaulting to 10M. For example, the bit rate of some panoramic cameras can easily reach 2, 30M. Of course, this can be ignored when not considering operating costs.

The requirement is certain. However, it is not necessary to solve it within SRS. A better solution is to implement it in an external system.
On the other hand, if you want to customize SRS and want to incorporate this functionality into SRS, it should not be done in the SrsStatistic module, but rather in the gb module, such as within SrsGb28181SipSession.

`TRANS_BY_GPT3`

PR #2429, opened 2021-06-23
srt publish core dump bug

runner365:
solve the core dump bug when srt publish.
Thanks Eric a lot for help to provide the machine.


trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -284,12 +284,14 @@ srs_error_t rtmp_client::connect() {
     _rtmp_conn_ptr = std::make_shared<SrsSimpleRtmpClient>(_url, cto, sto);
 
     if ((err = _rtmp_conn_ptr->connect()) != srs_success) {
+        srs_error("rtmp client in srt2rtmp connect fail url:%s", _url.c_str());
```
winlinvip: Is this print statement for debugging purposes? It might be better to print it in the outer layer, because here err is returned, and there might be duplicate logging outside as well.

If you need the information of _url, it is already included when wrapping err.

`TRANS_BY_GPT3`

trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -406,13 +408,18 @@ srs_error_t rtmp_client::rtmp_write_packet(char type, uint32_t timestamp, char*
     srs_error_t err = srs_success;
     SrsSharedPtrMessage* msg = NULL;
 
+    if (!_rtmp_conn_ptr) {
+        return srs_error_wrap(err, "rtmp connection is close");
```
winlinvip: At this time, when err is success, wrap(success) will cause the code outside to be success, which is incorrect.

Here, the error should be created using srs_error_new.

`TRANS_BY_GPT3`

Issue #2390, opened 2021-05-31
SRT推流 花屏必现的一种情况


zshaobo:
**描述(Description)**

> SRT推流 花屏必现的一种情况

1. SRS版本(Version): `4.0release`
1. SRS的日志如下(Log):
```
日志正常
```
1. SRS的配置如下(Config):
```
使用conf自带配置
```

**重现(Replay)**

> 重现Bug的步骤(How to replay bug?)

1. `obs推srt流，设置输出参数，微调：zerolatency`， 此时通过rtmp观看视频花屏，但使用srt-live-server是正常的
2.当使用ffmpeg加微调 zerolatency推流的，一切正常无此现象

**期望行为(Expect)**
解决花屏问题

zhouxiaojun2008:
This design corresponds to a situation where one frame corresponds to multiple slices.

`TRANS_BY_GPT3`

xiaozhihong:
I seem to be unable to appear. Can you please take a screenshot of your complete OBS configuration?

`TRANS_BY_GPT3`

zhouxiaojun2008:
> I seem to be unable to appear, can you please take a complete screenshot of your OBS configuration and show it to me?

> 'Zerolatency divides frames into multiple slices, usually one frame corresponds to one slice. The reason for the problem is in the `on_ts_h264` function inside `srt_to_rtmp`. When calling `write_h264_ipb_frame`, originally it was called once per frame. After setting zerolatency, one frame corresponds to n slices, so it should still be called once. However, in reality, it is called n times, which causes the screen to flicker. The overall architecture of the SRT integration module in SRS is generally fine, but there are still some issues in certain details.
![image](https://user-images.githubusercontent.com/31402829/122875390-c5960400-d366-11eb-91b1-7221173ca4ed.png)


`TRANS_BY_GPT3`

runner365:
> > I seem to be unable to appear, can you please take a screenshot of your complete OBS configuration and show it to me?
> 
> The zerolatency feature divides frames into multiple slices, usually one frame corresponds to one slice. The problem arises in the "on_ts_h264" function within the "srt_to_rtmp" module. When calling "write_h264_ipb_frame", originally it should be called once per frame. However, when zerolatency is enabled and one frame corresponds to n slices, it should still be called once. But in reality, it is called n times, which causes the screen to flicker. Overall, the architecture of the SRT integration module in SRS is mostly fine, but there are some issues with certain details.
[...]

PR #2440, opened 2021-06-24
SRT: Fix bug for multiple NALUs, when configure OBS in zerolatency.

runner365:
issue: https://github.com/ossrs/srs/issues/2390
the bug happen when obs push srt stream and configue h264 in zerolatency.


trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -524,11 +524,12 @@ srs_error_t rtmp_client::on_ts_video(std::shared_ptr<SrsBuffer> avs_ptr, uint64_
         
         // ibp frame.
         // TODO: FIXME: we should group all frames to a rtmp/flv message from one ts message.
-        srs_info("mpegts: demux avc ibp frame size=%d, dts=%d", frame_size, dts);
-        if ((err = write_h264_ipb_frame(frame, frame_size, dts, pts)) != srs_success) {
+        srs_info("mpegts: demux avc ibp frame size=%d, dts=%d", avs_ptr->left() + frame_size, dts);
+        if ((err = write_h264_ipb_frame(avs_ptr->head() - frame_size, avs_ptr->left() + frame_size, dts, pts)) != srs_success) {
```
xiaozhihong, line 529: Is there a mistake here, head() - frame_size is a wild pointer?

`TRANS_BY_GPT3`

xiaozhihong: I probably understand.
Slice multiple frames and NALUs, previously sent one by one.
Now, after parsing the SPS/PPS, the rest is ignored and sent directly merged.

`TRANS_BY_GPT3`

xiaozhihong: 我大概看明白了
多slice一个帧多个nalu, 以前是一个一个发
现在是把sps/pps解析完, 剩下的不管了, 直接合并发送

xiaozhihong: 我大概看明白了
多slice一个帧多个nalu, 以前是一个一个发
现在是把sps/pps解析完, 剩下的不管了, 直接合并发送

xiaozhihong (approved):

zhouxiaojun2008:
Shi Dashen, I have some doubts as follows:

This modification is based on the fact that the callback from ts outputs a complete frame of data. Therefore, here we directly send the entire frame of data at once. However, there may be a problem, for example, in the case of zerolatecy, where a frame may have multiple slices. The first slice is 00 00 00 01 xx, and the following slices are also 00 00 01 xx. If we send them all at once, the start code of the first slice is removed, but the start code of the following slices is not removed. This may pose a potential risk. Will there be any issues when this data is sent to SRS? Additionally, I will attach two images.



My modification is as follows:
`std::vector<std::string> vec_nalus;

    // send each frame.
    while (!avs_ptr->empty()) {
        char* frame = NULL;
        int frame_size = 0;
        if ((err = _avc_ptr->annexb_demux(avs_ptr, &frame, &frame_size)) != srs_success) {
            return srs_error_wrap(err, "demux annexb");
        }

        // 5bits, 7.3.1 NAL unit syntax,
        // ISO_IEC_14496-10-AVC-2003.pdf, page 44.
        //  7: SPS, 8: PPS, 5: I Frame, 1: P Frame
        SrsAvcNaluType nal_unit_type = (SrsAvcNaluType)(frame[0] & 0x1f);
        //srs_trace("frame size:%d,type:%d", frame_size, nal_unit_type);
        // ignore the nalu type sps(7), pps(8), aud(9)
        switch (nal_unit_type) {
            case SrsAvcNaluTypeSPS: {
                std::string sps;
                if ((err = _avc_ptr->sps_demux(frame, frame_size, sps)) != srs_success) {
                    return srs_error_wrap(err, "demux sps");
                }

                if (_h264_sps == sps) {
                    continue;
                }
                _h264_sps_changed = true;
                _h264_sps = sps;

                if ((err = write_h264_sps_pps(dts, pts)) != srs_success) {
                    return srs_error_wrap(err, "write sps/pps");
                }
                continue;
            }
            case SrsAvcNaluTypePPS: {
                std::string pps;
                if ((err = _avc_ptr->pps_demux(frame, frame_size, pps)) != srs_success) {
                    return srs_error_wrap(err, "demux pps");
                }

                if (_h264_pps == pps) {
                    continue;
                }
                _h264_pps_changed = true;
                _h264_pps = pps;

                if ((err = write_h264_sps_pps(dts, pts)) != srs_success) {
                    return srs_error_wrap(err, "write sps/pps");
                }
                continue;
            }
            case SrsAvcNaluTypeAccessUnitDelimiter:
            case SrsAvcNaluTypeSEI: {
                continue;
            }
            default: {
                // ibp frame
if (!avs_ptr->empty()) { // If it is not empty, it means it is a multi-slice
                    vec_nalus.push_back(std::string(frame, frame_size));
                }
                else {
                    if (vec_nalus.empty()) {
                        if ((err = write_h264_ipb_frame(frame, frame_size, dts, pts)) != srs_success)
                            return srs_error_wrap(err, "write h264 ipb frame");
                    }
                    else {
                        vec_nalus.push_back(std::string(frame, frame_size));
                        if ((err = write_h264_ipb_frame(vec_nalus, dts, pts)) != srs_success)
                            return srs_error_wrap(err, "write h264 ipb slice frame");
                    }
                    _last_live_ts = now_ms();
                }
                continue;
            }
        }
    }`

`TRANS_BY_GPT3`

PR #2459, opened 2021-07-03
Srt new log

runner365:
For srt has dependent thread which is different with st, the srt module should not use srs log directly.
We send the log happened in srt module to srs st which can print the log message.


trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -466,14 +512,14 @@ srs_error_t rtmp_client::on_ts_video(std::shared_ptr<SrsBuffer> avs_ptr, uint64_
...
         
-        //srs_trace_data(frame, frame_size, "video annexb demux:");
         // 5bits, 7.3.1 NAL unit syntax,
         // ISO_IEC_14496-10-AVC-2003.pdf, page 44.
         //  7: SPS, 8: PPS, 5: I Frame, 1: P Frame
         SrsAvcNaluType nal_unit_type = (SrsAvcNaluType)(frame[0] & 0x1f);
         
-        // ignore the nalu type sps(7), pps(8), aud(9)
-        if (nal_unit_type == SrsAvcNaluTypeAccessUnitDelimiter) {
+        // ignore the nalu type aud(9), pad(12)
```
xiaozhihong, line 520: These 3 diffs below have already been merged, it's best not to merge them again.
[PR 2440](https://github.com/ossrs/srs/pull/2440/files)

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_rtc_rtcp.cpp
```diff
@@ -474,7 +474,7 @@ SrsRtcpRR::SrsRtcpRR(uint32_t sender_ssrc)
     header_.version = kRtcpVersion;
     header_.length = 7;
     ssrc_ = sender_ssrc;
-    memset(&rb_, 0, sizeof(SrsRtcpRB));
+    //memset(&rb_, 0, sizeof(SrsRtcpRB));
```
xiaozhihong, line 477: This is unrelated to the content of the fix, is it a mistaken submission?

`TRANS_BY_GPT3`

runner365: Hmm, this code has a warning during compilation.
rb_ itself is a class object, and the constructor has a cleaning action for the parameters.
There is no need to use memset.

`TRANS_BY_GPT3`

trunk/src/srt/srt_to_rtmp.cpp
runner365: How should we handle this? Why is there still a diff for the newly merged changes?

`TRANS_BY_GPT3`

xiaozhihong: Please recommit, if you commit first and then merge other branches, this problem will occur, you have submitted it repeatedly.

`TRANS_BY_GPT3`

runner365:
submit a new one, so I have to delete this.

PR #2483, opened 2021-07-21
RTC: Support statistic for HTTP-API, HTTP-Callback and Security

duiniuluantanqin:


olivercloseoff6969:
Kk

chen-guanghua (approved):
add TODO for rtc client statistic use rtmp client type

duiniuluantanqin:
> add TODO for rtc client statistic use rtmp client type

add SrsRtcConnPlay and SrsRtcConnPublish

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -529,6 +533,12 @@ srs_error_t SrsRtcPlayStream::start()
             return srs_error_wrap(err, "on start play");
         }
     }
+	
+    // update the statistic when client discoveried.
+    SrsStatistic* stat = SrsStatistic::instance();
+    if ((err = stat->on_client(cid_.c_str(), req_, session_, SrsRtcConnPlay)) != srs_success) {
+        srs_trace("webrtc: add client failed!");
```
winlinvip: 1. Can this err be ignored?
2. Even if it is ignored, srs_freep(err) should still be done to avoid memory leaks.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_rtmp_stack.cpp
```diff
@@ -1628,7 +1628,9 @@ SrsResponse::~SrsResponse()
 string srs_client_type_string(SrsRtmpConnType type)
 {
     switch (type) {
-        case SrsRtmpConnPlay: return "Play";
+        case SrsRtmpConnPlay:
```
winlinvip: Is it better to distinguish between these two and return rtmp-play and rtc-play?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.hpp
```diff
@@ -486,6 +486,9 @@ class SrsRtcConnection : public ISrsResource, public ISrsDisposingHandler
 public:
     virtual const SrsContextId& get_id();
     virtual std::string desc();
+// Interface ISrsExpire.
+public:
+    virtual void expire();
```
winlinvip, line 491: This interface supports kickoff. We need to test if there will be any issues.

```
        client->conn->expire();
        srs_warn("kickoff client id=%s ok", client_id.c_str());
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_api.cpp
```diff
@@ -109,13 +112,18 @@ srs_error_t SrsGoApiRtcPlay::do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMe
...
-        srs_parse_rtmp_url(streamurl, tcUrl, stream_name);
 
-        int port;
-        string schema, host, vhost, param;
-        srs_discovery_tc_url(tcUrl, schema, host, vhost, app, stream_name, port, param);
+    string tcUrl;
+    srs_parse_rtmp_url(streamurl, tcUrl, stream_name);
+
+    int port;
+    string schema, host, vhost, param;
```
winlinvip: There are too many scattered local variables, which can easily lead to misuse.
Can we move the declaration of SrsRtcUserConfig ruc; forward and directly use its variable, for example:

```
SrsRtcUserConfig ruc;
if (true) {
    srs_parse_rtmp_url(streamurl, tcUrl, ruc.stream);
```

`TRANS_BY_GPT3`

trunk/src/protocol/srs_rtmp_stack.cpp
duiniuluantanqin: At present, this field is not being used, but it can still be distinguished.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.hpp
duiniuluantanqin: Yes, it is used for the kickoff feature. I have tested it locally and it passed.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
duiniuluantanqin: on_client always returns srs_success, which can be ignored.
However, for the sake of code consistency, I will still change it to wrap.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_api.cpp
duiniuluantanqin: Okay, I'll make some changes.

`TRANS_BY_GPT3`

PR #2493, opened 2021-07-28
解决./configure编译参数与--help不符

zhh0000zhh:
./configure: error: invalid option "--iconv"


myifoo:
Using `./configure --iconv=on` will result in an error, but using `./configure --with-iconv` compiles successfully.

`TRANS_BY_GPT3`

winlinvip:
Merged as https://github.com/ossrs/srs/commit/f6c445b58c16fb1a4fa789726751e20bac6ee1cf

winlinvip:
@zhh0000zhh Thanks very much 👍

PR #2502, opened 2021-08-02
RTMP Publish, RTC Play, Add GOP Cache

xiaofeilong:
当使用RTMP推流，RTC播放时，增加RTC的gop cache，减少RTC播放的起播延迟。

### DIFF
--- a/trunk/3rdparty/st-srs/md_darwin.S
+++ b/trunk/3rdparty/st-srs/md_darwin.S
@@ -11,17 +11,17 @@
      */
     #define JB_RBX  0
     #define JB_RBP  1
-    #define JB_R12  2                   /* Backup IP, https://www.cnblogs.com/Five100Miles/p/8458561.html */
-    #define JB_R13  3                   /* Backup SP, https://www.cnblogs.com/Five100Miles/p/8458561.html */
-    #define JB_R14  4                   /* Backup LR, https://www.cnblogs.com/Five100Miles/p/8458561.html */
-    #define JB_R15  5                   /* Backup PC, https://www.cnblogs.com/Five100Miles/p/8458561.html */
+    #define JB_R12  2                   /* The first six integer or pointer arguments are passed in registers RDI, RSI, RDX, RCX, R8, R9. */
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_config.cpp
```diff
@@ -4838,6 +4838,42 @@ bool SrsConfig::get_rtc_enabled(string vhost)
...
+    if (!conf || conf->arg0().empty()) {
+        return DEFAULT;
+    }
+
+    return SRS_CONF_PERFER_FALSE(conf->arg0());
+}
+
+int SrsConfig::get_rtc_gop_cache_max_packets(std::string vhost)
+{
+    static bool DEFAULT = false;
```
duiniuluantanqin: This DEFAULT still needs to be of type int.

`TRANS_BY_GPT3`

xiaofeilong: It should be used int. Just finished modifying.

`TRANS_BY_GPT3`

duiniuluantanqin:
This mechanism seems to conflict with the low latency of RTC.

`TRANS_BY_GPT3`

xiaofeilong:
> This mechanism seems to conflict with the low latency of RTC.

This depends on the specific application scenario. For large-scale low-latency live streaming with one-to-many connections, increasing the GOP (Group of Pictures) can reduce the startup delay, and the delay can be controlled by the client's playback rate. Additionally, through self-testing with the play demo, it was found that the delay had little impact.

`TRANS_BY_GPT3`

xiaofeilong:
> 这个机制跟rtc的低延迟好像有冲突

duiniuluantanqin:
I tested the functionality and there are no issues.

@xiaozhihong, please take a look and see if it is necessary to merge?

`TRANS_BY_GPT3`

xiaofeilong:
Applicable scenarios include live streaming in a fast-paced environment.

`TRANS_BY_GPT3`

quanqigu:
Experts, is this still appropriate? We are also facing the same issue. We need this fix.

`TRANS_BY_GPT3`

PR #2509, opened 2021-08-05
support http hooks for rtc: on_play/stop/publish/unpublish

duiniuluantanqin:


duiniuluantanqin:
Due to the special nature of WebRTC, signaling and data are two separate channels.
Currently, the **request to play/stream** is done through the signaling channel, while the handling of **stopping** is done through the data channel.
Therefore, `on_play/publish` is placed in the signaling interaction process, while `on_stop/unpublish` is placed in the data processing flow.

`TRANS_BY_GPT3`

chen-guanghua (approved):

xiaozhihong:
Record an idea, usually there is a "start streaming" and a "stop streaming", but it seems that WebRTC doesn't pay much attention to this point. The player provided by SRS also does not provide an entry for "stop streaming". In the future, it may be considered to add a status button on the SRS playback page. In the WebRTC scenario, after "start streaming", it will check the streaming status. If the streaming is normal, the button will change to "stop streaming". Users can actively stop the WebRTC streaming through the http API (signaling).

`TRANS_BY_GPT3`

xiaozhihong (approved):

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -379,6 +379,10 @@ SrsRtcPlayStream::SrsRtcPlayStream(SrsRtcConnection* s, const SrsContextId& cid)
 
 SrsRtcPlayStream::~SrsRtcPlayStream()
 {
+    if (req_) {
+        http_hooks_on_stop();
```
winlinvip, line 383: This is the issue that Guanghua mentioned before, regarding the problem of calling HTTP hooks in the destructor function.

publish' and 'play' must be called synchronously because they have authentication functionality.

unpublish' and 'stop' definitely do not require authentication, so it is possible to call them asynchronously.

Refer to `ISrsAsyncCallTask`.

Calling HTTP in the destructor can lead to coroutine switching, which carries a high risk.

`TRANS_BY_GPT3`

Issue #2515, opened 2021-08-07
G711分支在摄像头停止推流后，再次推流时没声音


qqtian01:
**描述(Description)**

> G711分支：在海康或大华的G711摄像头使用GB28181推流时，首次推流正常，当摄像头停止推流后，再次推流时webrtc播放没声音

1. SRS版本(Version): `G711分支`
1. SRS的日志如下(Log):
```
[2021-08-07 09:43:48.500][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found
[2021-08-07 09:43:48.592][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found
[2021-08-07 09:43:48.592][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found


```
1. SRS的配置如下(Config):
```
标准配置
```

**重现(Replay)**

> 重现Bug的步骤(How to replay bug?)

1. `4.0版本启用G711后都会有这个问题`

qqtian01:
> **Description**
'
Make sure to maintain the markdown structure.
> 
> > G711 branch: When using GB28181 to stream with G711 cameras from Hikvision or Dahua, the initial streaming works fine. However, when the camera stops streaming and tries to stream again, there is no sound in the webrtc playback.
> 
> 1. SRS version: `G711 branch`
> 2. The SRS log is as follows:
> 
> ```
> [2021-08-07 09:43:48.500][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found
> [2021-08-07 09:43:48.592][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found
> [2021-08-07 09:43:48.592][Warn][43716][851v4qc8][11] RTC: Drop for ssrc 0 not found
> ```
> 
> 1. The SRS configuration is as follows:
> 
> ```
> Standard configuration
> ```
> 
> **Replay**
> 
> > **How to replay bug?**
> 
> 1. `This problem occurs after enabling G711 in version 4.0`


`TRANS_BY_GPT3`

PR #2517, opened 2021-08-09
Feature/g711

qichaoshen82:
fix repushlish audio which is not opus drop.
https://github.com/ossrs/srs/issues/2515


duiniuluantanqin (approved):
This PR can fix the bug of disconnecting and then pushing again. However, there is still a remaining bug: the DVR recording will crash. 
This is the issue of Knowledge Planet: https://t.zsxq.com/6q3Z7ei

`TRANS_BY_GPT3`

PR #2533, opened 2021-08-14
fix SrsSessionInfo about copy assignment operator  and comparison operator.

johzzy:
fix SrsSessionInfo about copy assignment operator  and comparison operator.


winlinvip:
Merged 7c1d3edd3

PR #2542, opened 2021-08-18
Use SrsAsyncCallWorker in http hooks instead, to covert to async call.

duiniuluantanqin:


chen-guanghua (approved):

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -379,9 +426,8 @@ SrsRtcPlayStream::SrsRtcPlayStream(SrsRtcConnection* s, const SrsContextId& cid)
 
 SrsRtcPlayStream::~SrsRtcPlayStream()
 {
-    // TODO: FIXME: Use SrsAsyncCallWorker in http hooks instead, to covert to async call.
     if (req_) {
-        http_hooks_on_stop();		
+        session_->server_->exec_async_work(new SrsRtcAsyncCallOnStop(cid_, req_));
```
winlinvip, line 430: This implementation of callbacks requires inheriting a subclass from the interface in C++98. However, it is much more convenient to use C++11, as it provides solutions to this problem similar to anonymous functions in Go and JS.

This is the only place I have seen so far where using C++11 makes sense and clearly simplifies the logic.

In the future, we can consider using C++11 to implement this portion of the code.

`TRANS_BY_GPT3`

PR #2544, opened 2021-08-19
Support for multiple SPS/PPS, then pick the first one.

duiniuluantanqin:


mypopydev:
Any special reason just pick the first one?

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -806,41 +806,55 @@ srs_error_t SrsFormat::avc_demux_sps_pps(SrsBuffer* stream)
...
+            return srs_error_new(ERROR_HLS_DECODE_ERROR, "decode SPS size");
+        }
+        uint16_t sequenceParameterSetLength = stream->read_2bytes();
+        if (!stream->require(sequenceParameterSetLength)) {
+            return srs_error_new(ERROR_HLS_DECODE_ERROR, "decode SPS data");
+        }
+        if (vcodec->sequenceParameterSetNALUnit.size() > 0) {
+            stream->skip(sequenceParameterSetLength);
+            continue;
+        }
```
xiaozhihong, line 824: It seems that pick the last one~ hhhhhhhhhhhhhhhh

xiaozhihong (approved):

duiniuluantanqin:
> Any special reason just pick the first one?

Indeed, no official information was found, only found this: [https://stackoverflow.com/questions/20682275/h264-with-multiple-pps-and-sps].
> Most decoders/muxers parse the first SPS/PPS as part of a setup process and ignore subsequent SPSs.

For streaming media, multiple SPS/PPS is meaningless. For example, if you want to change the resolution midway, you can simply place the latest SPS/PPS before the next IDR frame. Therefore, choosing the first or the last one does not make any difference.

`TRANS_BY_GPT3`

PR #2545, opened 2021-08-19
fix some crash in rtc.

johzzy:


trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1715,6 +1715,10 @@ bool SrsRtmpFromRtcBridger::check_frame_complete(const uint16_t start, const uin
     for (uint16_t i = 0; i < cnt; ++i) {
         int index = cache_index((start + i));
         SrsRtpPacket* pkt = cache_video_pkts_[index].pkt;
+        if (!pkt) {
```
duiniuluantanqin, line 1718: How to reproduce this situation?

`TRANS_BY_GPT3`

duiniuluantanqin (approved):

PR #2552, opened 2021-08-24
Refine service to restart SRS for systemd

mawenwu1983:


winlinvip:
This is restarting the SRS service after SRS exits? I looked at this article, do we need to set `RestartSec`? I see it is set to the default 100ms.

https://unix.stackexchange.com/a/324297

I would like to extend Rahul's answer a bit.

systemd tries to restart multiple times (StartLimitBurst) and stops trying if the attempt count is reached within StartLimitIntervalSec. Both options belong to the [unit] section.

The default delay between executions is 100ms (RestartSec) which causes the rate limit to be reached very fast.

systemd won't attempt any more automatic restarts ever for units with Restart policy defined:

> Note that units which are configured for Restart= and which reach the start limit are not attempted to be restarted anymore; however, they may still be restarted manually at a later point, from which point on, the restart logic is again activated.

Rahul's answer helps, because the longer delay prevents reaching the error counter within the StartLimitIntervalSec time. The correct answer is to set both RestartSec and StartLimitBurst to reasonable values though.


`TRANS_BY_GPT3`

mawenwu1983:
Recently, the project used SRS. My colleague encountered a problem where SRS would suddenly stop when testing the push-pull stream (no error logs, the process just disappeared, and a coredump file was generated). My colleague couldn't access SRS, so they contacted me and I found out that SRS had already stopped.

The solution I used was to add the "Restart=always" field in the srs.service file. If SRS crashes, the daemon process can immediately restart.

I didn't use RestartSec because I wanted SRS to start immediately and provide services.

After adding Restart=always in the test, when manually killing the SRS process, systemd will automatically restart SRS. Currently, this can solve the problem with the daemon process.

The downside is that for earlier distributions without systemd, such as CentOS 6 and Ubuntu 14, this file has no effect and the srs script under init.d needs to be rewritten.

`TRANS_BY_GPT3`

winlinvip:
Merged as https://github.com/ossrs/srs/commit/54b706d36efeb7b3a487dd4d17b5778a045d7fe7

Thanks a lot 👍 @mawenwu1983

PR #2578, opened 2021-09-03
Enhance HLS: support http callback on_play/stop, support statistic

duiniuluantanqin:


trunk/src/app/srs_app_http_static.cpp
```diff
@@ -171,6 +184,162 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
...
+        alive(secret);
+        return SrsHttpFileServer::serve_m3u8_secret(w, r, fullpath);
+    }
+
+    if ((err = http_hooks_on_play(req)) != srs_success) {
+        return srs_error_wrap(err, "HLS: http_hooks_on_play");
+    }
+
+    if (secret.empty()) {
+        // make sure unique
```
xiaozhihong, line 209: Here, you can consider using the UUID method to generate a unique ID for each client.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -171,6 +184,162 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
...
+        do {
+            secret = srs_random_str(8);
+        } while (secret_is_exist(secret));
+    }
+
+    std::string res = "#EXTM3U\r";
+    res += "#EXT-X-STREAM-INF:BANDWIDTH=1,AVERAGE-BANDWIDTH=1\r";
+    res += hr->path() + "?" + SRS_SECRET_IN_HLS + "=" + secret;
+
```
xiaozhihong, line 223: This should be used with \n? It would be more appropriate to use SRS_CONSTS_LF.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -171,6 +184,162 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
...
+        SrsHttpHooks::on_stop(url, req);
+    }
+
+    return;
+}
+
+srs_error_t SrsVodStream::on_timer(srs_utime_t interval)
+{
+    srs_error_t err = srs_success;
+
```
xiaozhihong, line 334: map_secret_validity_ Manage whether it is expired,
map_secret_req_ Match secret and req
Here's a suggestion, you can try using only one map to manage, and put req and timeout in one struct.
std::map<std::string, struct SrsM3u8SecretInfo> map_secret_info
The advantage of this is that you don't have to manage two maps and make multiple correspondences. If it times out, it will definitely be cleared.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -33,15 +33,28 @@ using namespace std;
 #include <srs_app_pithy_print.hpp>
 #include <srs_app_source.hpp>
 #include <srs_app_server.hpp>
+#include <srs_service_utility.hpp>
+#include <srs_app_http_hooks.hpp>
+#include <srs_app_statistic.hpp>
+
+#define SRS_SECRET_IN_HLS "srs_secret"
```
xiaozhihong: secret' feels like it would be better to change the name. My first reaction is always "key"... If you don't look closely, you might think this is some kind of authentication code. Maybe changing it to 'userid', 'uniqueid', or 'magicid' would be better?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
duiniuluantanqin: \r or \n seems to work, but changing it to a macro is indeed better.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
duiniuluantanqin: Using the same method as cid should be no problem.

`TRANS_BY_GPT3`

xiaozhihong (approved):

PR #2579, opened 2021-09-04
Solve the problem that ubuntu(16.04) use Clion IDE compile SRS has error. that is mainly due to the dependency order of ffmpeg static library and pthread dependency.

qingfuliao:
Solve the problem that ubuntu(16.04) use Clion IDE compile SRS has error. that is mainly due to the dependency order of ffmpeg static library and pthread dependency.


NickNYF:
thank you!

winlinvip:
Thanks a lot 👍 @qingfuliao

PR #2618, opened 2021-09-17
[rtc] *Fix Fua package bug(payload size minus one).

zozobreakzou:
* This can cause webrtc video PacketBuffer assemble corrupt when (nal size - 1) % 1300 == 0
* issues about webrtc caton all caused by this bug

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -1192,7 +1192,7 @@ srs_error_t SrsRtcFromRtmpBridger::package_fu_a(SrsSharedPtrMessage* msg, SrsSam
     uint8_t header = sample->bytes[0];
     uint8_t nal_type = header & kNalTypeMask;
 
-    int num_of_packet = 1 + (sample->size - 1) / fu_payload_size;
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1192,7 +1192,7 @@ srs_error_t SrsRtcFromRtmpBridger::package_fu_a(SrsSharedPtrMessage* msg, SrsSam
     uint8_t header = sample->bytes[0];
     uint8_t nal_type = header & kNalTypeMask;
 
-    int num_of_packet = 1 + (sample->size - 1) / fu_payload_size;
+    int num_of_packet = 1 + (nb_left - 1) / fu_payload_size;
     for (int i = 0; i < num_of_packet; ++i) {
```
xiaozhihong, line 1196: FU-A, divide a frame into multiple RTP segments, referring to RFC6184.
image
Because FU-A already has a header describing the NAL header, the first byte of the H.264 frame, which is the NAL header, does not need to be included in the packaging.

`TRANS_BY_GPT3`

xiaozhihong (approved):

PR #2643, opened 2021-09-30
fix the same 'client_id' error when asynchronous call

duiniuluantanqin:


xiaozhihong (approved):

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -202,17 +203,22 @@ srs_error_t SrsVodStream::serve_m3u8_ctx(ISrsHttpResponseWriter * w, ISrsHttpMes
...
         // make sure unique
         do {
             ctx = srs_random_str(8);  // the same as cid
         } while (ctx_is_exist(ctx));
     }
 
+    SrsContextId cid = SrsContextId();
+    cid.set_value(ctx);
+    SrsContextRestore(_srs_context->get_id());
+    _srs_context->set_id(cid);
```
winlinvip: Can you simplify it a bit?

```
SrsContextRestore(_srs_context->get_id());
_srs_context->set_id(SrsContextId().set_value(ctx));
```

This `cid` local variable doesn't have much meaning, it's better to use an anonymous object.

Because adding a local variable means adding something that needs to be understood, and it also increases the possibility of being used.

`TRANS_BY_GPT3`

duiniuluantanqin: Hmm, yes

`TRANS_BY_GPT3`

duiniuluantanqin:
Move to https://github.com/ossrs/srs/pull/2665

`TRANS_BY_GPT3`

Issue #2624, opened 2021-09-22
GB28181: SRS 60路摄像头推流转发为hls后频繁奔溃


KOM6177:
> 注意：不提供以下信息的Issue会被直接删除(Please follow issue template, or we will delete it)

> 注意：咨询和讨论请提交到SRS星球(Please ask question at) http://bbs.ossrs.net

**描述(Description)**

> 描述你遇到了什么问题(Please description your issue here)
SRS 60路海康摄像头rtsp使用的gb28181协议rtsp推流转发为hls后频繁奔溃
1. SRS版本(Version): `4.0.123`
1. SRS的日志如下(Log):
![image](https://user-images.githubusercontent.com/59348943/134294140-cae6ac0d-0666-4b70-a49a-ffad217fc1bc.png)
110736
[2021-09-22 11:12:25.262][Trace][19309][5p973922] TCP: before dispose resource(HttpStream)(0x29d6f10), conns=3, zombies=0, ign=0, inz=0, ind=0
^[[33m[2021-09-22 11:12:25.262][Warn][19309][5p973922][104] client disconnect peer. ret=1007
^[[0m[2021-09-22 11:12:25.262][Trace][19309][57qp6832] TCP: clear zombies=1 resources, conns=3, removing=0, unsubs=0
[2021-09-22 11:12:25.262][Trace][19309][5p973922] TCP: disposing #0 resource(HttpStream)(0x29d6f10), conns=3, disposing=1, zombies=0
[2021-09-22 11:12:25.262][Trace][19309][q6jh5687] TCP: before dispose resource(HttpStream)(0x199f6b0), conns=2, zombies=0, ign=0, inz=0, ind=0
^[[33m[2021-09-22 11:12:25.262][Warn][19309][q6jh5687][104] client disconnect peer. ret=1007
^[[0m[2021-09-22 11:12:25.262][Trace][19309][57qp6832] TCP: clear zombies=1 resources, conns=2, removing=0, unsubs=0
[2021-09-22 11:12:25.262][Trace][19309][q6jh5687] TCP: disposing #0 resource(HttpStream)(0x199f6b0), conns=2, disposing=1, zombies=0
[2021-09-22 11:12:25.276][Trace][19309][l2166cq0] TCP: before dispose resource(HttpStream)(0x1ff1d60), conns=1, zombies=0, ign=0, inz=0, ind=0
[...]

PR #2650, opened 2021-10-04
GB28181: Check is es_info_length overflow es_map_length

xialixin:
解析ps流中 program stream map 时 es_info_length overflow es_map_length, 导致core的问题


xialixin:
https://github.com/ossrs/srs/issues/2624 It may be caused by this problem.
[2021-09-22 11:59:04.461][Trace][25220][27p0czq3] gb28181: ps map audio es_type=svac(9b), es_id=da, es_info_length=15359

`TRANS_BY_GPT3`

PR #2420, opened 2021-06-18
Support Simulcast Stream (munging style sdp)

johzzy:
## Publisher: 

Set the simulcast layers using `numberOfSimulcastLayers`, `numberOfSimulcastLayers` options 1, 2, 3, where 1 is the default value, other values illegally, redirect to 1 when used.
When `numberofsimulcastLayers = 1`, `Simulcast Stream` degraded to `Single Stream`

    webrtc://127.0.0.1/live/livestream?numberOfSimulcastLayers=1  (equivalent to single stream)
    webrtc://127.0.0.1/live/livestream?numberOfSimulcastLayers=2  (2 layer simulcast stream)
    webrtc://127.0.0.1/live/livestream?numberOfSimulcastLayers=3  (3 layer simulcast stream)

## Player: 

The player uses the `layer` to select the stream layer, the optional value of the Layer is 0, 1, 2, 3, ...
If the `layer` value is illegal, the redirection is the default state. 

    low: webrtc://127.0.0.1/live/livestream?layer=0
    mid: webrtc://127.0.0.1/live/livestream?layer=1
    high: webrtc://127.0.0.1/live/livestream?layer=2
    other1: webrtc://127.0.0.1/live/livestream?layer=1,2
    other2: webrtc://127.0.0.1/live/livestream?layer=1,2&foo=bar

PR #2659, opened 2021-10-11
Support Simulcast Stream (RID Based Simulcast)

johzzy:
## Publisher: 

open Simulcast推流 (`/players/rtc_publisher_simulcast.html` page)

## Player: 

open RTC播放器 (`/players/rtc_player.html` page)

    low: webrtc://127.0.0.1/live/livestream?layer=low
    mid: webrtc://127.0.0.1/live/livestream?layer=mid
    high: webrtc://127.0.0.1/live/livestream?layer=high

Related Link: https://github.com/ossrs/srs/pull/2420

### DIFF
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -3338,6 +3338,7 @@ srs_error_t SrsRtcConnection::negotiate_play_capability(SrsRtcUserConfig* ruc, s
             }
 
             track->set_direction("sendonly");
+            // fixme: order of track
             sub_relations.insert(make_pair(publish_ssrc, track));
             if (track->rid_.ssrc > 0) {
                 srs_info("publish_ssrc=%u, play ssrc=%u, rid ssrc=%u, rid=%s", publish_ssrc, track->ssrc_, track->rid_.ssrc, track->rid_.rid.c_str());
@@ -3399,7 +3400,7 @@ void video_track_generate_play_offer(SrsRtcTrackDescription* track, string mid, 
 }
 
 // note: get [0,1,2] from "webrtc://127.0.0.1/live/livestream?layer=0,1,2&foo=bar"
-bool parse_layers_in_url(std::vector<int>& layers, std::string name) {
+bool parse_layers_in_url(std::vector<std::string>& layers, std::string name) {
     // livestream?layer=1
     // livestream?layer=1,2,3
     // livestream?layer=1,2,3&foo=bar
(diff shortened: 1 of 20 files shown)


trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -1748,6 +1748,14 @@ void SrsRtcPublishStream::update_send_report_time(uint32_t ssrc, const SrsNtp& n
     }
 }
 
+void SrsRtcPublishStream::bind_rid(const SrsRidInfo &rid_info) {
+    // NOTE: get unused video_track, and active as rid info.
+    SrsRtcVideoRecvTrack* video_track = get_video_track(0);
```
xiaozhihong, line 1760: What does '0' mean?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -3156,6 +3194,9 @@ srs_error_t SrsRtcConnection::generate_publish_local_sdp(SrsRequest* req, SrsSdp
         }
 
         SrsVideoPayload* payload = (SrsVideoPayload*)video_track->media_;
+        if (srs_string_contains(req->param,"?numberOfSimulcastLayers=spec3")) {
+            payload->h264_param_.x_google_bitrate = "x-google-min-bitrate=5000;x-google-max-bitrate=8000;x-google-start-bitrate=6000";
```
xiaozhihong: Would it be better for the user to control this? For example, the simulcast.html page can modify the SDP answer itself.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
johzzy: get unused video_track,

Before executing here, multiple tracks will be allocated based on simulcast SDP, with the default SSRC as 0, indicating unused.

When reaching this point, take an unused track and bind it with RID and SSRC.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_sdp.cpp
```diff
@@ -1177,3 +1209,215 @@ srs_error_t SrsSdp::update_msid(string id)
...
+    // a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
+    // a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
+
+    for (size_t i=0; i<kExtMapFieldSize; ++i) {
+        if (ext == kExtMapFieldArray[i]) {
+            ext_id_list_[i] = id;
+        }
+    }
+}
+
```
xiaozhihong: Please process the following code according to the SRS method. SRS has a structure called Rtp, which does not use the C memory structure but follows the decode method.

`TRANS_BY_GPT3`

xiaozhihong (approved):

PR #2665, opened 2021-10-12
fix the same 'client_id' error when asynchronous call

duiniuluantanqin:


trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -351,22 +351,22 @@ srs_error_t SrsRtcPLIWorker::cycle()
 
     return err;
 }
-
-SrsRtcAsyncCallOnStop::SrsRtcAsyncCallOnStop(SrsContextId c, SrsRequest * r)
-{
+
```
duiniuluantanqin, line 354: The modification here is to remove the ^M symbol, the same applies below.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -194,6 +195,12 @@ srs_error_t SrsVodStream::serve_m3u8_ctx(ISrsHttpResponseWriter * w, ISrsHttpMes
     srs_assert(hr);
 
     SrsRequest* req = hr->to_request(hr->host())->as_http();
+    // discovery vhost, resolve the vhost from config
+    SrsConfDirective* parsed_vhost = _srs_config->get_vhost(req->vhost);
```
duiniuluantanqin, line 199: Lines 199-202 search for vhost to resolve the issue of inconsistent stream IDs in the statistics information.

`TRANS_BY_GPT3`

PR #2675, opened 2021-10-13
Remove assert

runner365:
For there are lots of stream mpegts packet which has wrong pes header or pat/pmt header, we replace the assert by log and discard the wrong packet.


runner365:
erase the pr by using pr: https://github.com/ossrs/srs/pull/2670

PR #2686, opened 2021-10-18
RTC: support refer and security

duiniuluantanqin:


trunk/src/app/srs_app_rtc_api.hpp
```diff
@@ -27,11 +47,9 @@ class SrsGoApiRtcPlay : public ISrsHttpHandler
 private:
     virtual srs_error_t do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, SrsJsonObject* res);
     srs_error_t check_remote_sdp(const SrsSdp& remote_sdp);
-private:
-    virtual srs_error_t http_hooks_on_play(SrsRequest* req);
 };
 
-class SrsGoApiRtcPublish : public ISrsHttpHandler
+class SrsGoApiRtcPublish : public ISrsHttpHandler, public SrsRtcAccessControl
```
winlinvip, line 52: This inheritance relationship does not hold.

When using inheritance, one must be very careful and consider whether they truly have an inheritance relationship, rather than just looking at the code implementation.

`TRANS_BY_GPT3`

PR #2701, opened 2021-11-01
Package rpm support for CentOS/RHEL

purplegrape:


winlinvip:
@purplegrape Nice 👍

winlinvip:
If we could submit the RPM to the yum repository, that would be even better, as users would be able to install SRS directly using yum.

I wonder if it would be convenient to continue improving this? 😀

`TRANS_BY_GPT3`

purplegrape:
GitHub has a free static blog feature, which can be used as an official yum source. 
Refer to my blog.
https://purplegrape.github.io/centos/7/x86_64/repoview/srs.html


`TRANS_BY_GPT3`

PR #2710, opened 2021-11-04
减少无意义的输出

purplegrape:
减少无意义的输出，不必专门提示，懂的人会自然会看日志，尤其是tail -f ，反而显得不专业，不是吗？

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -2026,12 +2026,6 @@ srs_error_t SrsConfig::parse_options(int argc, char** argv)
         if (get_log_tank_file() && log_filename.empty()) {
             return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "no log file");
         }
-        if (get_log_tank_file()) {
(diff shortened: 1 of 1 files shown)


winlinvip:
SRS does not show off, we do not consider ourselves professional or not, we are just a group of losers, there is nothing professional about us.

`TRANS_BY_GPT3`

winlinvip:
Thank you very much for submitting the PR, 😄

`TRANS_BY_GPT3`

purplegrape:
The reason for considering it meaningless is because:
1. `tail -f` is indeed not user-friendly enough. Can't we use `cat` or `vim` instead? I'm not trying to argue, but changing the expression to something like `see log /var/log/xxx.log` would be better.
2. The output of `./init.d/srs status` forces users to only use one startup method and implies that the author also uses this method, which confuses users and is not conducive to using other startup methods. For example, `systemd start srs`, `service srs start`, or directly using the command line `srs -c xx.conf &`.
3. Other excellent software seems to have no precedent for outputting similar information.

`TRANS_BY_GPT3`

PR #2711, opened 2021-11-04
设置默认配置文件

purplegrape:
建议参考FHS文件目录标准，设置默认配置文件指定为/etc/srs/srs.conf , 如-c 指定配置文件，则按指定的为准
（本人非码农，见谅）

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -50,7 +50,7 @@ const char* _srs_version = "XCORE-" RTMP_SIG_SRS_SERVER;
 #define SRS_CONF_PERFER_TRUE(conf_arg) conf_arg != "off"
 
 // default config file.
-#define SRS_CONF_DEFAULT_COFNIG_FILE "conf/srs.conf"
+#define SRS_CONF_DEFAULT_COFNIG_FILE "/etc/srs/srs.conf"
 
 // '\n'
 #define SRS_LF (char)SRS_CONSTS_LF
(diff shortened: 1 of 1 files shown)


winlinvip:
@purplegrape Thanks a lot

I make a little change and merge as https://github.com/ossrs/srs/commit/1f4dad024e8e9e02b459c68e8225fc194af120bf

However, I still prioritize local configuration. First, I look for the configuration specified by the user, then I look for the default `conf/srs.conf`, and finally, I look for `/etc/srs/srs.conf`.

`TRANS_BY_GPT3`

purplegrape:
Suggest putting the default configuration in the compilation parameters, as there are many parameters involved, which is a sign of software maturity in accordance with FHS.

Refer to the compilation parameters of nginx.
--prefix=/usr/share/nginx
--conf-path=/etc/nginx/nginx.conf
--error-log-path=/var/log/nginx/error.log
--pid-path=/run/nginx.pid

In addition, it is more refreshing to start without any parameters by default, and it is also recommended to refer to nginx, which is equivalent to -c /etc/xx.conf.

Finally, it is not recommended for software to be designed for multiple local instances from the beginning. Simply copying the parent directory and changing the configuration file port can generate a new instance. This is a lazy approach. Multiple instances should be specified by different configuration files, and the paths should be written within the configuration files.

`TRANS_BY_GPT3`

winlinvip:
Currently, when the SRS is configured in a way that does not match the user-specified configuration file, it will print a warning and provide the used configuration file:
```
$ ./objs/srs
[2021-12-22 13:08:58.706][Trace][95002][d81x10st] XCORE-SRS/4.0.207(Leo)
[2021-12-22 13:08:58.707][Warn][95002][d81x10st][2] user config  does not exists, use conf/srs.conf instead
```

As well as the attempted configuration files that were not found:
```
$ ~/srs/objs/srs -c user/docker.conf
[2021-12-22 13:06:42.387][Trace][94789][20v73877] XCORE-SRS/4.0.207(Leo)
[2021-12-22 13:06:42.388][Error][94789][20v73877][2] Failed, code=1023 : config parse options : no config file at 
user/docker.conf, user/srs.conf, conf/srs.conf, /etc/srs/srs.conf
thread [94789][20v73877]: do_main() [src/main/srs_main_server.cpp:108][errno=2]
thread [94789][20v73877]: parse_options() [src/app/srs_app_config.cpp:2011][errno=2](No such file or directory)
```

Other suggestions, if you find them important, please submit a pull request (PR).

`TRANS_BY_GPT3`

Issue #1613, opened 2020-02-20
同时开启截图和录制，on_dvr获取不到param，IP地址也变成127.0.0.1


tadliu:
**描述(Description)**
同时开启截图和录制，on_dvr获取不到param，IP地址也变成127.0.0.1
不开启截屏，一切正常，感觉是复用错了了tcp连接
> 描述你遇到了什么问题(Please description your issue here)

1. SRS版本(Version): `SRS/3.0.121(OuXuli)`
https://github.com/ossrs/srs, licensed under The MIT License (MIT), built at 2020-02-19 13:34:57, configured by --x86-x64 , which means --prefix=/usr/local/srs --with-hls --with-hds --with-dvr --with-ssl --with-transcode --with-ingest --with-stat --with-http-callback --with-http-server --with-stream-caster --with-http-api --with-librtmp --without-research --with-utest --without-gperf --without-gmc --without-gmd --without-gmp --without-gcp --without-gprof --log-trace --cc=gcc --cxx=g++ --ar=ar --ld=ld --randlib=randlib
1. SRS的日志如下(Log):
```
[2020-02-20 11:28:40.747][Warn][90723][532][11] ignore task failed code=3008 : callback on_dvr http://xxx/auth.php : 
http post on_dvr uri failed, client_id=532, url=http://xxxx/auth.php, request=
{"action":"on_dvr","client_id":532,"ip":"127.0.0.1","vhost":"__defaultVhost__",
"app":"live","stream":"457556","param":"","cwd":"/root/srs/trunk",
"file":"./objs/nginx/html/live/2020-02/457556-1582169235680.flv"}, 
response=1, code=200 : http: response number code 1
```
1. SRS的配置如下(Config):
```
listen 1935;
max_connections 1000;
srs_log_tank file;
srs_log_file ./objs/srs.log;
http_api {
    enabled on;
    listen 1985;
    raw_api {
        enabled on;
        allow_update on;
    }
}
http_server {
    enabled on;
    listen 8080;
    dir ./objs/nginx/html;
}
stats {
    network 0;
    disk sda sdb xvda xvdb;
}
vhost __defaultVhost__ {
    dvr {
        enabled on;
        dvr_path ./objs/nginx/html/live/[2006]-[01]/[stream]-[timestamp].flv;
        dvr_plan session;
        dvr_apply live/xxx;
        dvr_wait_keyframe on;
    }
    http_remux {
        enabled on;
        mount [vhost]/[app]/[stream].flv;
[...]

PR #2714, opened 2021-11-04
DVR: copy req from publish

duiniuluantanqin:
see https://github.com/ossrs/srs/issues/1613#issuecomment-960623359

### DIFF
--- a/trunk/src/app/srs_app_dvr.cpp
+++ b/trunk/src/app/srs_app_dvr.cpp
@@ -598,8 +598,12 @@ srs_error_t SrsDvrPlan::initialize(SrsOriginHub* h, SrsDvrSegmenter* s, SrsReque
     return err;
 }
 
-srs_error_t SrsDvrPlan::on_publish()
+srs_error_t SrsDvrPlan::on_publish(SrsRequest* r)
 {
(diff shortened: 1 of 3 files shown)


trunk/src/app/srs_app_dvr.cpp
```diff
@@ -762,11 +766,11 @@ srs_error_t SrsDvrSegmentPlan::initialize(SrsOriginHub* h, SrsDvrSegmenter* s, S
     return srs_success;
 }
 
-srs_error_t SrsDvrSegmentPlan::on_publish()
+srs_error_t SrsDvrSegmentPlan::on_publish(SrsRequest* r)
```
xiaozhihong, line 769: Does 'SrsDvr::on_publish()' also need to pass in 'SrsRequest'?
Both 'SrsLiveSource::on_publish()' and 'SrsDvr::on_publish()' should be involved in this process.

`TRANS_BY_GPT3`

duiniuluantanqin: Yes, a copy of the req in SrsDvr is also made, ensuring that the entire DVR process uses the publisher's req.

`TRANS_BY_GPT3`

xiaozhihong (approved):

PR #2752, opened 2021-11-24
SrsHlsMuxer::segment_open if ts_url is not under m3u8_dir,but m3u8_dir is happenly the prefix of ts_url it will cause ts cannot be found while playing.

bluestn:
SrsHlsMuxer::segment_open if ts_url is not under m3u8_dir,but m3u8_dir is happenly the prefix of ts_url it will cause ts cannot be found while playing.


winlinvip:
I wonder why this PR changes 1359 lines of code? This could mean changes TAB to space, adding space for each line, updating the newline, or other small changes for each line.

Why did you change the entire file at line 1359? Can you only modify the necessary code without adding extra information?

`TRANS_BY_GPT3`

PR #2766, opened 2021-12-02
RTC: Fix nack encode seqnum

xiaofeilong:


xiaozhihong:
Thank you for submitting. Can you please provide a uTest? This way, we will never encounter any issues with future changes and we can also verify that the modified code is working correctly.

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_rtc.cpp
```diff
@@ -744,6 +744,32 @@ VOID TEST(KernelRTCTest, NACKFetchRTPPacket)
...
 
+VOID TEST(KernelRTCTest, NACKEncode)
+{
+    uint32_t ssrc = 123;
+    char buf_before[kRtcpPacketSize];
+    SrsBuffer stream_before(buf_before, sizeof(buf_before));
+    
+    SrsRtcpNack rtcpNack_encode(ssrc);
+    for(uint16_t i = 16; i < 50; ++i) {
+        rtcpNack_encode.add_lost_sn(i);
```
xiaozhihong: rtcpNack_encode, rtcpNack_decode, It is better to name them consistently with the style of srs, so change them to rtcp_nack_encode, rtcp_nack_decode would be better.

`TRANS_BY_GPT3`

xiaozhihong (approved):

winlinvip:
@xiaofeilong @xiaozhihong 👍

PR #2768, opened 2021-12-02
Fix bugs for RTC2RTMP.

xiaozhihong:
1. Cache IDR frame's rtp timestamp instead of avsync timestamp.
2. Cache clock rate calculate by sender report.
3. Using srs_rtp_seq_distance instead of direct minus.
4. Add utest of av timestamp sync when duplicated sender report.

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -1278,7 +1278,7 @@ SrsRtmpFromRtcBridger::SrsRtmpFromRtcBridger(SrsLiveSource *src)
     is_first_audio = true;
     is_first_video = true;
     format = NULL;
-    key_frame_ts_ = -1;
+    rtp_key_frame_ts_ = -1;
     header_sn_ = 0;
     memset(cache_video_pkts_, 0, sizeof(cache_video_pkts_));
 }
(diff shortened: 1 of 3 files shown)


trunk/src/app/srs_app_rtc_source.hpp
```diff
@@ -307,7 +307,7 @@ class SrsRtmpFromRtcBridger : public ISrsRtcSourceBridger
     RtcPacketCache cache_video_pkts_[s_cache_size];
     uint16_t header_sn_;
     uint16_t lost_sn_;
-    int64_t key_frame_ts_;
+    int64_t key_frame_ts_; // RTP timestamp, no rtmp.
```
chen-guanghua: 1. key_frame_ts_: The meaning expressed is the timestamp of the RTP key frame. The variable is defined as rtp_key_frame_ts_. No need to add comments, as the meaning is clear from the variable name.
2. If additional comments are required, they can be added as a new line above the variable. Follow the same formatting as other comments. 😄

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1505,23 +1505,23 @@ srs_error_t SrsRtmpFromRtcBridger::packet_video_key_frame(SrsRtpPacket* pkt)
     }
 
     if (-1 == key_frame_ts_) {
-        key_frame_ts_ = pkt->get_avsync_time();
+        key_frame_ts_ = pkt->header.get_timestamp();
         header_sn_ = pkt->header.get_sequence();
         lost_sn_ = header_sn_ + 1;
         // Received key frame and clean cache of old p frame pkts
         clear_cached_video();
-        srs_trace("set ts=%lld, header=%hu, lost=%hu", key_frame_ts_, header_sn_, lost_sn_);
```
chen-guanghua, line 1513: %lld can be used with %"PRId64 ", PRId64 has better compatibility with related output formats. For example: 

srs_trace("DASH: Config fragment=%" PRId64 ", period=%" PRId64, fragment, update_period);

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2299,40 +2304,41 @@ void SrsRtcRecvTrack::update_send_report_time(const SrsNtp& ntp, uint32_t rtp_ti
...
+        //   sender_report1: rtp_time1 = 10000, ntp_time1 = 40000
+        //   sender_report : rtp_time  = 10960, ntp_time  = 40020
+        //   (rtp_time - rtp_time1) / (ntp_time - ntp_time1) = 960 / 20 = 48,
+        // Now we can calcualte ntp time(ntp_x) of any given rtp timestamp(rtp_x),
+        //   (rtp_x - rtp_time) / (ntp_x - ntp_time) = 48   =>   ntp_x = (rtp_x - rtp_time) / 48 + ntp_time;
+        double sys_time_elapsed = static_cast<double>(last_sender_report_ntp_.system_ms_) - static_cast<double>(last_sender_report_ntp1_.system_ms_);
+
+        // Check sys_time_elapsed is equal to zero.
+        if (fpclassify(sys_time_elapsed) == FP_ZERO) {
+            return;
```
chen-guanghua, line 2322: fpclassify was introduced in C++11. Should we consider older versions of gcc?

https://en.cppreference.com/w/cpp/numeric/math/fpclassify

`TRANS_BY_GPT3`

xiaozhihong: image
This is a function in the standard C library, and there is also a corresponding version in std:: library.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
xiaozhihong: This has been changed to %u, it is easier to read when printed as a 32-bit unsigned integer.

`TRANS_BY_GPT3`

chen-guanghua (approved):

PR #2775, opened 2021-12-07
Add st stat summaries

attenuation:
add st stat info in sumarries
```
    "st": {
      "runq": 0,
      "ioq": 7,
      "zombieq": 0,
      "sleepq": 11
    }
```

### DIFF
--- a/trunk/3rdparty/st-srs/common.h
+++ b/trunk/3rdparty/st-srs/common.h
@@ -238,6 +238,10 @@ typedef struct _st_vp {
     _st_clist_t run_q;          /* run queue for this vp */
     _st_clist_t io_q;           /* io queue for this vp */
     _st_clist_t zombie_q;       /* zombie queue for this vp */
+
+    int runq_size;          /* number of threads on run queue */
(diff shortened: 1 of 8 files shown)


winlinvip:
First of all, thank you very much for submitting the Patch. 👍 

Secondly, ST is a very low-level library, so if you want to submit a Patch, you must add comprehensive UTest coverage to avoid introducing issues. Even if the code seems unlikely to cause problems, it still needs to be covered, as the consequences of any issues could be unimaginable. 

Lastly, collecting information about coroutines is a debugging feature that requires following certain conventions. For example, using macro definitions to disable statistics by default and only enabling them in SRS, so as not to affect other users. Additionally, querying through logs instead of API is recommended, as API is a general capability that may not be needed by all users. Furthermore, SRS provides gdb scripts to view the number of coroutines, and similar information can be obtained using gdb script methods as well. 

Based on the considerations mentioned above, I don't think this PR is suitable for merging. 

Once again, thank you. 

`TRANS_BY_GPT3`

Issue #2780, opened 2021-12-09
HTTP: 2GB录制的大文件下载失败


CodeflowSun:
**描述(Description)**

1. SRS版本(Version): `4.0`

2. SRS的日志如下(Log):
```
HTTP #0 39.152.7.125:25758 GET http://123.6.55.240:18080/live/202112070435020001.2021.12.09.11.30.36.mp4, content-length=-1
[2021-12-09 16:49:38.132][Trace][31841][31w7h85h] http match file=./objs/nginx/html/live/202112070435020001.2021.12.09.11.30.36.mp4, pattern=/, upath=/live/202112070435020001.2021.12.09.11.30.36.mp4
[2021-12-09 16:49:38.132][Trace][31841][31w7h85h] TCP: before dispose resource(HttpStream)(0x20346f0), conns=5, zombies=0, ign=0, inz=0, ind=0
[2021-12-09 16:49:38.132][Error][31841][31w7h85h][11] serve error code=3014 : process request=0 : mux serve : serve http : http mp4 streaming ./objs/nginx/html/live/202112070435020001.2021.12.09.11.30.36.mp4 overflow. size=2846362636, offset=0
```
3.具体情况
我们录制了一个时长为156分钟的视频，录制方式为MP4，录制之后，视频无法下载，ftp下载到本地后，视频可以通过软件正常播放，视频大小为2.65GB

**重现(Replay)**

1. `可以上传一个超过2GB的文件到srs服务器`
2. `通过http进行访问`

**期望行为(Expect)**
正常下载

问题已经找到，关键在于srs_app_http_static.cpp:157

```
    if (end == -1) {
        end = (int)(fs->filesize() - 1);
    }
    
    if (end > fs->filesize() || start > end || end < 0) {
        return srs_error_new(ERROR_HTTP_REMUX_OFFSET_OVERFLOW, "http mp4 streaming %s overflow. size=%" PRId64 ", offset=%d",
            fullpath.c_str(), fs->filesize(), start);
    }
```
以上代码为4.0源码，在强制类型转换时，大文件的end会为负数，导致后续进入异常处理环节

winlinvip:
Not bad, welcome to submit a PR, bro 👍

`TRANS_BY_GPT3`

CodeflowSun:
The PR has been submitted, and the concerns have been mentioned in the PR.
I don't understand C++ and have only made fixes based on my own understanding. I also don't have a testing environment locally, so this bare submission carries risks. Please understand.

[...]

PR #2781, opened 2021-12-10
Fix bug for HTTP, 2GB录制的大文件下载失败 #2780

CodeflowSun:
我是做Java的，对C++一窍不通，根据我的个人理解对#2780做了修复，如果有修改不当的地方敬请谅解，关闭这个pr就好，目前根据我的判断大部分的修改应该问题不大，但是有一行代码的修改感觉有一定的风险
```
if ((err = copy(w, fs, r, (int)left)) != srs_success)   >> if ((err = copy(w, fs, r, left)) != srs_success)
```
一行这个修改有两个疑虑
1.不确定stl的copy是否支持long的传参
2.在文件过大时是否对内存产生不良影响

### DIFF
--- a/trunk/src/app/srs_app_http_static.cpp
+++ b/trunk/src/app/srs_app_http_static.cpp
@@ -137,7 +137,7 @@ srs_error_t SrsVodStream::serve_flv_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
     return err;
 }
 
-srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, string fullpath, int start, int end)
+srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, string fullpath, int start, long end)
 {
     srs_error_t err = srs_success;
     
@@ -154,7 +154,7 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
     
     // parse -1 to whole file.
     if (end == -1) {
-        end = (int)(fs->filesize() - 1);
+        end = fs->filesize() - 1;
     }
     
     if (end > fs->filesize() || start > end || end < 0) {
@@ -180,8 +180,8 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
     fs->seek2(start);
     
     // send data
-    if ((err = copy(w, fs, r, (int)left)) != srs_success) {
-        return srs_error_wrap(err, "read mp4=%s size=%d", fullpath.c_str(), (int)left);
+    if ((err = copy(w, fs, r, left)) != srs_success) {
+        return srs_error_wrap(err, "read mp4=%s size=%" PRId64 , fullpath.c_str(), left);
     }
     
     return err;
--- a/trunk/src/app/srs_app_http_static.hpp
+++ b/trunk/src/app/srs_app_http_static.hpp
@@ -31,7 +31,7 @@ public:
     virtual ~SrsVodStream();
 protected:
     virtual srs_error_t serve_flv_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, std::string fullpath, int offset);
-    virtual srs_error_t serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, std::string fullpath, int start, int end);
+    virtual srs_error_t serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, std::string fullpath, int start, long end);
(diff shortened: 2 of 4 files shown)


winlinvip:
Thank you for the PR, but it's indeed not quite suitable. No worries, we will find time to take a look.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -137,7 +137,7 @@ srs_error_t SrsVodStream::serve_flv_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
     return err;
 }
 
-srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, string fullpath, int start, int end)
+srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMessage* r, string fullpath, int start, long end)
```
chundonglinlin, line 140: It is more appropriate to use int64_t instead of long.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_static.cpp
```diff
@@ -180,8 +180,8 @@ srs_error_t SrsVodStream::serve_mp4_stream(ISrsHttpResponseWriter* w, ISrsHttpMe
     fs->seek2(start);
     
     // send data
-    if ((err = copy(w, fs, r, (int)left)) != srs_success) {
-        return srs_error_wrap(err, "read mp4=%s size=%d", fullpath.c_str(), (int)left);
+    if ((err = copy(w, fs, r, left)) != srs_success) {
```
chundonglinlin, line 183: If only the `left` variable type is changed to `int64_t`, but the declaration of `copy(ISrsHttpResponseWriter* w, SrsFileReader* fs, ISrsHttpMessage* r, int64_t size)` should also be modified to `int64_t`, and the variable types inside the copy function should remain consistent, otherwise the modification will not be achieved.

`TRANS_BY_GPT3`

winlinvip: Could you please create a new PR? @CodeflowSun mentioned that he works with Java, so it may not be suitable for him to continue modifying the C++ code. Please make sure to maintain the markdown structure.

@CodeflowSun has already identified the problem, which is already very good. Please make sure to maintain the markdown structure.

Thank you very much to both of you. 🙏 Please make sure to maintain the markdown structure.

`TRANS_BY_GPT3`

chundonglinlin: Okay, I will create a new PR to address this issue. Thanks to @CodeflowSun for helping to identify the problem 🙏.

`TRANS_BY_GPT3`

chundonglinlin: When modifying the distribution of mp4 files, it was discovered that flv files also have this issue. `offset` is defined as `int` type. If you use `ffplay http://127.0.0.1:8080/test.flv`, the flv file will keep buffering after exceeding 2GB. After making modifications, it was found that flv does not support range operations. Summarizing the above observations:

1. When distributing mp4 and flv files, if they exceed 2GB, there will be overflow or connection timeout issues.
2. When distributing flv files, VLC player will exit or fail when dragging.

`TRANS_BY_GPT3`

chundonglinlin: > When modifying the distribution of mp4 files, it was found that flv also had this problem. `offset` is defined as `int` type. If `ffplay http://127.0.0.1:8080/test.flv` is used, the flv file will freeze indefinitely after exceeding 2GB. After modification, it was found that range operation is not supported for flv. Summarize the above phenomena: 1. When distributing mp4 and flv files, if they exceed 2GB, overflow or connection timeout may occur. 2. When distributing flv files, VLC will exit or fail when dragging during playback.

Phenomenon 2, when dragging to play flv, does not support the range parameter. Is it necessary? When distributing flv files using nginx, it supports range operation.

`TRANS_BY_GPT3`

winlinvip:
Fix by https://github.com/ossrs/srs/pull/2809 

Thanks again @CodeflowSun @chundonglinlin

trunk/src/app/srs_app_http_static.cpp
ellenWei11: Hi, I would like to ask if the drag-and-drop feature for FLV files has been resolved?

`TRANS_BY_GPT3`

ellenWei11: I compiled nginx with the following configure arguments: --prefix=/usr/local/nginx --with-http_stub_status_module --with-http_ssl_module --with-http_gzip_static_module --with-http_mp4_module --with-http_flv_module. 
At the same time, I used yamdi to add keyframe information to the recorded videos. 
yamdi -i source.flv -o dest.flv //input video output video
Then, I can use nginx to play the recorded files and enable dragging and dropping. 
So, the problem should be that the DVR recording does not include keyframe information, which is why dragging and dropping playback is not possible.

`TRANS_BY_GPT3`

Issue #2760, opened 2021-11-29
WebRTC AV1 推流失败 no found valid AV1 payload type


trswygg:
> 注意：提问前，请先看FAQ(Please read FAQ before file an issue) https://github.com/ossrs/srs/issues/2716

WebRTC AV1 推流，播放失败，使用，srs返回400，查看sdp发现sdp中 AV1X 变为了`a=rtpmap:35 AV1/90000`
使用自建服务器(局域网linux)
chrome:
Google Chrome | 96.0.4664.55 (正式版本) (x86_64)
操作系统 | macOS 版本12.0.1（版号21A559）
JavaScript | V8 9.6.180.12
用户代理 | Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.55 Safari/537.36

> 描述你遇到了什么问题(Please description your issue here)

1. SRS版本(Version): `SRS/5.0.19(Leo)`

1. SRS的日志如下(Log):
```
[2021-11-29 11:19:10.223][Trace][274637][373v3df6] Hybrid cpu=1.00%,161MB, cid=1,0, timer=62,0,0, clock=0,49,0,0,0,0,0,0,0
[2021-11-29 11:19:13.138][Trace][274637][67f446i5] HTTP #0 172.17.0.3:34462 POST http://ubuntu.pve/rtc/v1/publish/?codec=av1, content-length=6179
[2021-11-29 11:19:13.138][Trace][274637][67f446i5] RTC publish webrtc://ubuntu.pve/live/livestream?codec=av1, api=https://ubuntu.pve:443/rtc/v1/publish/?codec=av1, tid=2ba78c4, clientip=192.1.1.50, app=live, stream=livestream, offer=5678B, eip=, codec=av1
[2021-11-29 11:19:13.138][Trace][274637][67f446i5] ignore attribute=, value=
[2021-11-29 11:19:13.138][Warn][274637][67f446i5][11] RTC error code=5018 : create session : create session : add publisher : publish negotiate : no found valid AV1 payload type
thread [274637][67f446i5]: do_serve_http() [src/app/srs_app_rtc_api.cpp:443][errno=11]
thread [274637][67f446i5]: create_session() [src/app/srs_app_rtc_server.cpp:480][errno=11]
thread [274637][67f446i5]: do_create_session() [src/app/srs_app_rtc_server.cpp:497][errno=11]
[...]

PR #2784, opened 2021-12-10
RTC: Replace payload name AV1X with AV1 for WebRTC. (#2760)

Huachao:
Fix the WebRTC **"no found valid AV1 payload type"** issue

### DIFF
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -2904,6 +2904,11 @@ srs_error_t SrsRtcConnection::negotiate_publish_capability(SrsRtcUserConfig* ruc
         } else if (remote_media_desc.is_video() && ruc->codec_ == "av1") {
             std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1");
             if (payloads.empty()) {
+                // Be compatible with the Chrome M96, still check the AV1X encoding name
+                // @see https://bugs.chromium.org/p/webrtc/issues/detail?id=13166
+                payloads = remote_media_desc.find_media_with_encoding_name("AV1X");
+            }
+            if (payloads.empty()) {
                 return srs_error_new(ERROR_RTC_SDP_EXCHANGE, "no found valid AV1 payload type");
             }
 
@@ -3214,6 +3219,11 @@ srs_error_t SrsRtcConnection::negotiate_play_capability(SrsRtcUserConfig* ruc, s
             track_descs = source->get_track_desc("audio", "opus");
         } else if (remote_media_desc.is_video() && ruc->codec_ == "av1") {
             std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1");
+            if (payloads.empty()) {
+                // Be compatible with the Chrome M96, still check the AV1X encoding name
+                // @see https://bugs.chromium.org/p/webrtc/issues/detail?id=13166
+                payloads = remote_media_desc.find_media_with_encoding_name("AV1X");
+            }
             if (payloads.empty()) {
                 return srs_error_new(ERROR_RTC_SDP_EXCHANGE, "no found valid AV1 payload type");
             }
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -2902,7 +2902,7 @@ srs_error_t SrsRtcConnection::negotiate_publish_capability(SrsRtcUserConfig* ruc
                 break;
             }
         } else if (remote_media_desc.is_video() && ruc->codec_ == "av1") {
-            std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1X");
+            std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1");
             if (payloads.empty()) {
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -2902,7 +2902,7 @@ srs_error_t SrsRtcConnection::negotiate_publish_capability(SrsRtcUserConfig* ruc
                 break;
             }
         } else if (remote_media_desc.is_video() && ruc->codec_ == "av1") {
-            std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1X");
+            std::vector<SrsMediaPayloadType> payloads = remote_media_desc.find_media_with_encoding_name("AV1");
```
winlinvip, line 2905: It seems that the old version is AV1X, and the new one is AV1.

In that case, it's best to search for both, which means passing two parameters into it:

```
remote_media_desc.find_media_with_encoding_name("AV1X", "AV1");
```

Alternatively, search twice:

```
payloads  = remote_media_desc.find_media_with_encoding_name("AV1");
if (payloads.empty()) payloads  = remote_media_desc.find_media_with_encoding_name("AV1X");
```

This can avoid failures in older browsers, which is considered a compatibility change.

`TRANS_BY_GPT3`

Huachao: @winlinvip nice catch, I will fix it

Huachao:
Ping @winlinvip @chen-guanghua, please help to review.

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -3213,13 +3218,18 @@ srs_error_t SrsRtcConnection::negotiate_play_capability(SrsRtcUserConfig* ruc, s
...
+                // @see https://bugs.chromium.org/p/webrtc/issues/detail?id=13166
+                payloads = remote_media_desc.find_media_with_encoding_name("AV1X");
+            }
             if (payloads.empty()) {
                 return srs_error_new(ERROR_RTC_SDP_EXCHANGE, "no found valid AV1 payload type");
             }
 
             remote_payload = payloads.at(0);
-            track_descs = source->get_track_desc("video", "AV1X");
+            track_descs = source->get_track_desc("video", "AV1");
```
winlinvip, line 3232: Why isn't this place compatible with AV1X?

`TRANS_BY_GPT3`

Huachao: @winlinvip Because I found that the get_track_desc function only checks for audio, so I didn't add it. If we consider future compatibility, we can also add it here.
https://github.com/ossrs/srs/blob/7fcafef6c4c1bb5a78518b8c620204e6ed396818/trunk/src/app/srs_app_rtc_source.cpp#L663-L685

`TRANS_BY_GPT3`

winlinvip: 👍 I'm wondering if it's possible to only save the new AV1 format when saving, so we don't have to judge everywhere.

Regardless of AV1X or AV1, would it be better to only store AV1 in the source?

`TRANS_BY_GPT3`

Huachao: @winlinvip I don't know if my understanding of what you said is correct or not, but my understanding is that here, the source is always judged using AV1, regardless of the version of the code, both old and new.

If it's about this issue, my understanding is that it may not work because the source obtained at the streaming end should essentially be provided during the pushing process, that is, the code for obtaining the source in the play section below.
https://github.com/ossrs/srs/blob/c5a511bc84259028d90a456261c8187256721160/trunk/src/app/srs_app_rtc_conn.cpp#L3185-L3187
Therefore, if it is desired here that the `media_->name` of `source` is consistently AV1, it means that during the negotiation in the publish process, `encoding_name` needs to be stored as AV1 uniformly, rather than storing it based on the actual AV1 or AV1X. Furthermore, the source here is essentially related to the line of code negotiating the payload in the publish section below, which will ultimately be saved in `stream_desc` and then saved by the `fetch_or_create` method.
https://github.com/ossrs/srs/blob/c5a511bc84259028d90a456261c8187256721160/trunk/src/app/srs_app_rtc_conn.cpp#L2914 

Based on the analysis above, my initial idea was to directly modify the line of code for streaming. In the AV1 scenario, I wanted to hard code it as AV1, regardless of whether the SDP sent by the browser is AV1X or AV1. However, the problem is that making this change could potentially cause the streaming to fail because the local SDP returned to the browser would be AV1, while the SDP declared by the browser would be AV1X. This depends on whether older versions of Chrome can continue streaming after receiving AV1.

I'm not sure if my understanding above is correct, so I would appreciate it if someone knowledgeable could enlighten me. I am new to this and may have described the process incorrectly.

`TRANS_BY_GPT3`

winlinvip: This is a compatibility issue. If you have tried it and encountered problems, then it definitely won't work.

If you can't use AV1 entirely, then please make sure to use compatibility with AV1 and AV1X wherever necessary. I remember that this particular place is not yet compatible, right?

`TRANS_BY_GPT3`

Huachao: @winlinvip There is indeed a problem, so I also added compatibility checks here.

`TRANS_BY_GPT3`

winlinvip (approved):

winlinvip:
Thank you very much. 👍

Issue #2770, opened 2021-12-06
SRT: SRS will crash if `streamid` is too long


allanlei:
**描述(Description)**

> 描述你遇到了什么问题(Please description your issue here)

SRS will crash if the `streamid` is too long

1. SRS版本(Version): `v4.0-b0`

1. SRS的日志如下(Log):
```
$ docker run -it --rm --network=host -p 10080:10080/udp -v "$(pwd)/srs.conf:/usr/local/srs/conf/docker.conf:ro" ossrs/srs:v4.0-b0
WARNING: Published ports are discarded when using host network mode
[2021-12-06 13:23:59.780][Trace][1][n3pq1580] XCORE-SRS/4.0.201(Leo)
[2021-12-06 13:23:59.780][Trace][1][n3pq1580] config parse complete
[2021-12-06 13:23:59.781][Trace][1][n3pq1580] write log to console
[2021-12-06 13:23:59.781][Trace][1][n3pq1580][MAIN] SRS/4.0.201(Leo), MIT
[2021-12-06 13:23:59.781][Trace][1][n3pq1580] authors: https://github.com/ossrs/srs/blob/4.0release/trunk/AUTHORS.txt
[...]

PR #2786, opened 2021-12-11
SRT: Fix srt2rtmp crash when streamid is too long(#2770)

akanchi:


winlinvip:
@akanchi  Thanks a lot 
@runner365 Please review this PR, appreciate that.

runner365 (approved):

trunk/src/srt/srt_to_rtmp.cpp
```diff
@@ -15,6 +15,7 @@
 #include <srs_app_config.hpp>
 #include <srs_kernel_stream.hpp>
 #include <list>
+#include <sstream>
```
chundonglinlin, line 18: The `#include <sstream>` has already been imported in the `srs_app_config.hpp` header file, so there is no need to repeat `#include <sstream>`.

`TRANS_BY_GPT3`

chundonglinlin: The crash issue has been verified and there are no other problems by using stringstream instead of a char array. Thanks to @akanchi for the pull request.

`TRANS_BY_GPT3`

winlinvip:
Merged, thanks a lot @akanchi @runner365 @chundonglinlin

PR #2800, opened 2021-12-17
DVR: fix large file mp4 dvr failed.

xiaozhihong:


xiaozhihong:
DVR MP4 section, stco supports 32-bit offset, co64 supports 64-bit offset. Both encoding and decoding of these two boxes are implemented in SRS. However, only stco is used when writing. When the file is too large and the offset exceeds UINT32_MAX, the MP4 offset will be incorrect, resulting in playback failure for files beyond 4GB.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_mp4.cpp
```diff
@@ -4840,10 +4852,19 @@ srs_error_t SrsMp4SampleManager::write(SrsMp4MovieBox* moov)
...
+        SrsMp4FullBox* co = NULL;
+        // When sample offset less than UINT32_MAX, we use stco(support 32bit offset) box to save storage space.
+        if (samples.empty() || (*samples.rbegin())->offset < UINT32_MAX) {
+            // stco support 32bit offset.
+            co = new SrsMp4ChunkOffsetBox();
+            stbl->set_stco(static_cast<SrsMp4ChunkOffsetBox*>(co));
+        } else {
+            // When sample offset bigger than UINT32_MAX, we use co64(support 64bit offset) box to avoid overflow.
+            co = new SrsMp4ChunkLargeOffsetBox();
+            stbl->set_co64(static_cast<SrsMp4ChunkLargeOffsetBox*>(co));
```
winlinvip, line 4864: Can we use a 64-bit box instead of a 32-bit box?

`TRANS_BY_GPT3`

xiaozhihong: Okay, but for compatibility and to save N 4-byte storage considerations, let the program automatically choose what may be more appropriate.

`TRANS_BY_GPT3`

winlinvip: MP4 files, if only saving 4 bytes, the benefit is too small.

Please directly use the 64-bit one, it's simpler in terms of logic.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_mp4.cpp
```diff
@@ -4864,10 +4876,10 @@ srs_error_t SrsMp4SampleManager::write(SrsMp4MovieBox* moov)
         SrsMp4SampleSizeBox* stsz = new SrsMp4SampleSizeBox();
         stbl->set_stsz(stsz);
         
-        SrsMp4ChunkOffsetBox* stco = new SrsMp4ChunkOffsetBox();
-        stbl->set_stco(stco);
+        SrsMp4ChunkLargeOffsetBox* co64 = new SrsMp4ChunkLargeOffsetBox();
+        stbl->set_co64(co64);
         
-        if ((err = write_track(SrsFrameTypeAudio, stts, stss, ctts, stsc, stsz, stco)) != srs_success) {
+        if ((err = write_track(SrsFrameTypeVideo, stts, stss, ctts, stsc, stsz, co64)) != srs_success) {
```
chundonglinlin: The type here is `SrsFrameTypeAudio`.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_mp4.cpp
xiaozhihong: Temporary method to determine whether it is 32-bit or 64-bit. 
Because SRS has unit tests for parsing MP4, using co64 directly will cause the unit tests to fail. 
The SrsMp4Decoder uses stco, but actually needs to determine whether the box is stco or co64. 
This part of the code is only used in unit tests.

`TRANS_BY_GPT3`

winlinvip (approved):

winlinvip (approved):

PR #2803, opened 2021-12-20
RTC: Eliminate duplicate assignment for video packet frame type 

chundonglinlin:
when video_track call on_rtp function, duplicate assignment packet's frame_type.
```
pkt->frame_type = SrsFrameTypeVideo;
```

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -2516,8 +2516,6 @@ srs_error_t SrsRtcVideoRecvTrack::on_rtp(SrsRtcSource* source, SrsRtpPacket* pkt
 {
     srs_error_t err = srs_success;
 
-    pkt->frame_type = SrsFrameTypeVideo;
-
     pkt->set_avsync_time(cal_avsync_time(pkt->header.get_timestamp()));
 
     if ((err = source->on_rtp(pkt)) != srs_success) {


trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2516,8 +2516,6 @@ srs_error_t SrsRtcVideoRecvTrack::on_rtp(SrsRtcSource* source, SrsRtpPacket* pkt
 {
     srs_error_t err = srs_success;
 
-    pkt->frame_type = SrsFrameTypeVideo;
```
winlinvip, line 2519: I didn't see where it repeated. Can you please specify which function?

`TRANS_BY_GPT3`

chundonglinlin: Sorry, the code being called is not listed here. It has been assigned a value.
https://github.com/ossrs/srs/blob/c15fadb9d52214f7ddc497ad4582c593f6a3f1a9/trunk/src/app/srs_app_rtc_conn.cpp#L1450
However, when processing video_track->on_rtp(source, pkt), it is assigned a value again.

`TRANS_BY_GPT3`

winlinvip: 👍 After reviewing, it should indeed be deleted:

```
// frame_type is not set
SrsRtcAudioRecvTrack::on_rtp
// frame_type is set
SrsRtcVideoRecvTrack::on_rtp
```

It is better to set it uniformly on the periphery, for example:

```
SrsRtcPublishStream::do_on_rtp_plaintext
SrsRtcFromRtmpBridger::package_opus
SrsRtcFromRtmpBridger::package_stap_a
SrsRtcFromRtmpBridger::package_nalus
SrsRtcFromRtmpBridger::package_single_nalu
SrsRtcFromRtmpBridger::package_fu_a
```

So it is reasonable to delete in the track.

`TRANS_BY_GPT3`

winlinvip (approved):

winlinvip:
Thanks a lot @chundonglinlin

Issue #2837, opened 2022-01-01
希望在callback接口on_publish中增加stream_id属性


stormbirds:
基于现在的开发版本，我自己实现sip信令服务器的过程中，需要在邀请拉流之后获取stream信息，但是现在的on_publish回调中只有app和stream，没有stream_id，这时候我需要先通过http_api的api/v1/streams接口获取到stream列表，然后通过app和stream筛选出具体流信息，我觉得如果在on_publish回调中返回stream_id在业务端保存会更合适，这样可以基于stream_id直接获取对应的流信息。

winlinvip:
Welcome to submit a patch.'
Make sure to maintain the markdown structure.

`TRANS_BY_GPT3`

jinleileiking:
cid pubtime也有用' translates to 'cid pubtime is also useful' in English.

`TRANS_BY_GPT3`

winlinvip:
Please provide a detailed explanation of the specific application scenarios, instead of just giving a conclusion. Do not add any unnecessary information unless it is necessary.

`TRANS_BY_GPT3`

jinleileiking:
> @jinleileiking Please provide a detailed explanation of the application scenario, instead of just giving a conclusion. Do not include any unnecessary information unless it is necessary.

When there is a flash break, for example, when stream1 stops and stream2 starts, there is a rather troublesome case in distributed systems. That is, when sending callbacks or triggering transcoding tasks, these two events may be out of order. My experience is to use CID and pubtime for disorder handling. In the case of network jitter, these four messages (stream1 start, stop, stream2 start, stop) can sometimes be disordered, in extreme cases.

`TRANS_BY_GPT3`

winlinvip:
Thank you, described very clearly. 👍

`TRANS_BY_GPT3`

drchzy:
Curve-saving law: Use the client_id in the callback information to query the clients interface => can obtain the stream_id.

`TRANS_BY_GPT3`

28269890:
Not bad, not bad, I support it. My application scenario also needs this functionality.

Actual application situation: In the list, you can see the information of the stream currently being pushed, and you can also verify the running status of the stream.

To implement this application scenario, it is necessary to directly obtain the information of the stream. However, currently, the data submitted from the hook cannot be directly obtained from the API.

--------

The ultimate goal is to be able to directly obtain the relevant information of the stream through the API, such as modifying the API, for example, api/v1/vhost/app/stream.

[...]

PR #2838, opened 2022-01-03
在on_publish的http_callback接口回调中增加stream_id

stormbirds:
#2837 
在on_publish的回调事件中增加了流ID属性，暂时只实现了rtmp连接，rtc暂时未看明白流程。

### DIFF
--- a/trunk/src/app/srs_app_rtmp_conn.cpp
+++ b/trunk/src/app/srs_app_rtmp_conn.cpp
@@ -694,7 +694,8 @@ srs_error_t SrsRtmpConn::do_playing(SrsLiveSource* source, SrsLiveConsumer* cons
 
     // update the statistic when source disconveried.
     SrsStatistic* stat = SrsStatistic::instance();
-    if ((err = stat->on_client(_srs_context->get_id().c_str(), req, this, info->type)) != srs_success) {
+    std::string stream_id;
+    if ((err = stat->on_client(_srs_context->get_id().c_str(), req, this, info->type, stream_id)) != srs_success) {
         return srs_error_wrap(err, "rtmp: stat client");
     }
     
@@ -825,17 +826,18 @@ srs_error_t SrsRtmpConn::publishing(SrsLiveSource* source)
             return srs_error_wrap(err, "rtmp: referer check");
         }
     }
-    
-    if ((err = http_hooks_on_publish()) != srs_success) {
-        return srs_error_wrap(err, "rtmp: callback on publish");
-    }
-    
+    std::string stream_id = "vid-" + srs_random_str(7);
+
     // TODO: FIXME: Should refine the state of publishing.
-    if ((err = acquire_publish(source)) == srs_success) {
+    if ((err = acquire_publish(source, stream_id)) == srs_success) {
+
+        if ((err = http_hooks_on_publish(stream_id)) != srs_success) {
+            return srs_error_wrap(err, "rtmp: callback on publish");
+        }
         // use isolate thread to recv,
         // @see: https://github.com/ossrs/srs/issues/237
         SrsPublishRecvThread rtrd(rtmp, req, srs_netfd_fileno(stfd), 0, this, source, _srs_context->get_id());
-        err = do_publishing(source, &rtrd);
+        err = do_publishing(source, &rtrd, stream_id);
         rtrd.stop();
     }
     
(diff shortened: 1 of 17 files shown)


winlinvip:
👍 Take matters into your own hands, and you will have enough clothing and food.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -825,17 +826,18 @@ srs_error_t SrsRtmpConn::publishing(SrsLiveSource* source)
             return srs_error_wrap(err, "rtmp: referer check");
         }
     }
-    
-    if ((err = http_hooks_on_publish()) != srs_success) {
-        return srs_error_wrap(err, "rtmp: callback on publish");
-    }
-    
+    std::string stream_id = "vid-" + srs_random_str(7);
```
jinleileiking, line 829: stream_id has already been generated, right? There is no need to generate it again here.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -825,17 +826,18 @@ srs_error_t SrsRtmpConn::publishing(SrsLiveSource* source)
             return srs_error_wrap(err, "rtmp: referer check");
         }
     }
-    
-    if ((err = http_hooks_on_publish()) != srs_success) {
-        return srs_error_wrap(err, "rtmp: callback on publish");
```
jinleileiking, line 830: This part has been deleted, cannot be restored, and will not invoke this hook anymore?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
winlinvip: According to the URL, it is better to obtain the statistic stream object and then retrieve the ID.

`TRANS_BY_GPT3`

duiniuluantanqin:
The changes you made are too significant.
17 files and a series of interfaces have been modified. What if we need app_id/vhost_id/param_id in the future?
And this hasn't even involved rtc and gb yet.

Can we refer to the handling method of server_id?
```
obj->set("server_id", SrsJsonAny::str(stat->server_id().c_str()));
```

Alternatively, we can discuss other more reasonable approaches.

ps: 
https://github.com/ossrs/srs/issues/2837#issue-1091890023
After reviewing your requirements, can't the app and stream be combined into one continuous flow? I don't think the stream_id is necessary.

`TRANS_BY_GPT3`

stormbirds:
> This change you made is too big. 17 files, a series of interfaces have been modified. What if we need app_id/vhost_id/param_id in the future? And this hasn't even involved rtc and gb yet. 
> 
> Can we refer to the handling method of server_id?
> 
> ```
> obj->set("server_id", SrsJsonAny::str(stat->server_id().c_str()));
> ```
> 
> Or we can discuss other more reasonable approaches.
> 
> ps: [#2837 (comment)](https://github.com/ossrs/srs/issues/2837#issue-1091890023) I went back to look at your requirement, can't the app+stream determine the stream? I feel like stream_id is not necessary.

Yes, it can be determined. The problem is that every time I need this stream information, I have to request the streams API to fetch the entire list, and then compare and filter out the streams that meet the criteria. I think it would be better to directly retrieve the specific information of this stream using the streams?id=streamId method.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
stormbirds: > It is better to obtain the statistic stream object based on the URL and then retrieve the ID.

I just tested it and it is possible to obtain the statistic stream object through the URL. This method is actually more reasonable. Previously, I didn't think of it because I haven't looked at the SRS source code much. I will submit a new method here, so I suggest closing this PR. Also, thank you for providing the idea.

`TRANS_BY_GPT3`

stormbirds:
> This change you made is too big. 17 files, a series of interfaces have been modified. What if we need app_id/vhost_id/param_id in the future? And this hasn't even involved rtc and gb yet. 
> 
> Can we refer to the handling method of server_id?
> 
> ```
> obj->set("server_id", SrsJsonAny::str(stat->server_id().c_str()));
> ```
> 
> Or we can discuss other more reasonable approaches.
> 
> ps: [#2837 (comment)](https://github.com/ossrs/srs/issues/2837#issue-1091890023) I went back to look at your requirement, can't the app+stream determine the stream together? I feel like stream_id is not necessary.

Also, if multiple playback requests in gb28181 use the same app and stream, it seems that it is impossible to distinguish which specific stream.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
winlinvip: NICE 👍

PR #2844, opened 2022-01-04
HTTPCallback:Add the stream id attribute to the http on_publish callback

stormbirds:
## Summary

#2837

## Details

采取对代码最小的侵入性建议，使用SrsStatistic对象根据url来获取stream的方式获取到stream_id。
在SrsStatistic类中增加find_stream_by_url方法，基于url获取Statistic Stream对象。

PR #2859, opened 2022-01-09
Add the stream id attribute to the http on_publish callback

winlinvip:
For https://github.com/ossrs/srs/pull/2844

### DIFF
--- a/trunk/src/app/srs_app_rtmp_conn.cpp
+++ b/trunk/src/app/srs_app_rtmp_conn.cpp
@@ -825,20 +825,25 @@ srs_error_t SrsRtmpConn::publishing(SrsLiveSource* source)
             return srs_error_wrap(err, "rtmp: referer check");
         }
     }
-    
-    if ((err = http_hooks_on_publish()) != srs_success) {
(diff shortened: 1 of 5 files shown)


duiniuluantanqin:
Added a flag, adjusted the timing of http hook and stat, all just for one streamid, it doesn't seem worth it.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -825,20 +825,25 @@ srs_error_t SrsRtmpConn::publishing(SrsLiveSource* source)
             return srs_error_wrap(err, "rtmp: referer check");
         }
     }
-    
-    if ((err = http_hooks_on_publish()) != srs_success) {
-        return srs_error_wrap(err, "rtmp: callback on publish");
-    }
+    bool hooks_not_publish = true;
```
winlinvip, line 828: Why do we need to add this flag?

`TRANS_BY_GPT3`

winlinvip:
Fixed by 6a108fab6dddf5c4e0c61a88131f1ebc0ec89c30

PR #2870, opened 2022-01-11
bugfix: SrsMetaCache memleak; getaddrinfo use delete

bluestn:
## Summary

bugfix: SrsMetaCache memleak
bugfix: use freeaddrinfo instad of delete to release the result of getaddrinfo

## Details

...


trunk/src/app/srs_app_source.cpp
```diff
@@ -662,15 +662,15 @@ void SrsGopCache::clear()
         srs_freep(msg);
     }
     gop_cache.clear();
-    
+
```
winlinvip, line 665: Please change the editor and do not submit such blank modifications.

`TRANS_BY_GPT3`

winlinvip:
Dup to #2880

PR #2873, opened 2022-01-11
LiveSource: Refine fetch for external exposed interface.

chundonglinlin:
LiveSource's fetch function should be extended Parameters, because it update the request.

### DIFF
--- a/trunk/src/app/srs_app_source.hpp
+++ b/trunk/src/app/srs_app_source.hpp
@@ -450,7 +450,6 @@ public:
     virtual srs_error_t fetch_or_create(SrsRequest* r, ISrsLiveSourceHandler* h, SrsLiveSource** pps);
 public:
     // Get the exists source, NULL when not exists.
-    // update the request and return the exists source.
     virtual SrsLiveSource* fetch(SrsRequest* r);
 public:
     // dispose and cycle all sources.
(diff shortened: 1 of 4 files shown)


trunk/src/app/srs_app_source.hpp
```diff
@@ -451,7 +451,7 @@ class SrsLiveSourceManager : public ISrsHourGlass
 public:
     // Get the exists source, NULL when not exists.
     // update the request and return the exists source.
-    virtual SrsLiveSource* fetch(SrsRequest* r);
+    virtual SrsLiveSource* fetch(SrsRequest* r, bool update = true);
```
winlinvip: This modification is the simplest, but it lacks maintainability and does not solve the fundamental problem.

```
Functions like fetch, which are similar to get, can never update.
```

Adding a parameter cannot solve this problem. Therefore, the fetch function should be separated into an API.

`TRANS_BY_GPT3`

chundonglinlin: Remove the `update` operation from `fetch` and place `update` in the corresponding position where the `request` needs to be updated, such as in `fetch_or_create` requiring an update to the `request`.

`TRANS_BY_GPT3`

winlinvip:
NICE 👍

winlinvip (approved):

Issue #1399, opened 2019-06-25
srs配置文件是否支持类似nginx中的include指令？


weiyuefei:
## 背景

@winlinvip 你好，请问srs是否可以支持类似nginx中的include指令配置，因为如果vhost很多情况下，将所有vhost都保存到srs.conf中不方便阅读和管理，如果能支持include指令，就可以很方便将vhost分离到一个个小的文件中，比如1.vhost.conf, 2.vhost.conf等，然后通过在srs.conf中include vhost/*.conf即可将所有vhost加载进来。

## 场景一 **加载vhost配置**

```
### 定义文件 srs.com.conf, 目录 ./conf/vhosts/srs.com.conf
vhost srs.com {
    hls {
        enabled         on;
    }
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].flv;
    }
}

### 可通过如下方式加载对应配置
listen              1935;
max_connections     1000;
srs_log_tank        console;
daemon              off;

http_api {
    enabled         on;
    listen          1985;
}

http_server {
    enabled         on;
    listen          8080;
    dir             ./objs/nginx/html;
}

include             ./conf/vhosts/srs.com.conf;
```

## 场景二 **加载vhost中某一分块配置，如转码配置**

```
### 定义文件 engine.ffhd.conf, 目录 ./conf/engines/engine.ffhd.conf
engine ffhd {
    enabled         on;
    vcodec          libx264;
    vbitrate        1200;
    vfps            25;
    vwidth          1382;
    vheight         576;
    vthreads        6;
    vprofile        main;
    vpreset         medium;
    vparams {
    }
    acodec          libfdk_aac;
    abitrate        70;
    asample_rate    44100;
    achannels       2;
    aparams {
    }
    output          rtmp://127.0.0.1:[port]/[app]?vhost=[vhost]/[stream]_[engine];
}

### 可通过如下方式加载对应配置
listen              1935;
max_connections     1000;
srs_log_tank        console;
daemon              off;

http_api {
    enabled         on;
    listen          1985;
}

http_server {
    enabled         on;
    listen          8080;
    dir             ./objs/nginx/html;
}

vhost __defaultVhost__ {
    transcode {
        enabled     on;
        ffmpeg      ./objs/ffmpeg/bin/ffmpeg;

        include     ./conf/engines/engine.ffhd.conf;
    }
}
```

[...]

PR #2878, opened 2022-01-12
Support include directive for config file

mapengfei53:
## Summary

Support include import configuration

## Details

Add include to import multiple configuration files, similar to nginx.


winlinvip:
Thank you, 👍

This feature has been mentioned before, to support the nginx include directive, but no one has implemented it until today: #1399

Also, please make sure to add Utest coverage to verify that specific configurations are working correctly. We have previously added Utest testing to our configuration file parsing.

`TRANS_BY_GPT3`

mapengfei53:
Okay, I will try my best to improve the code. Thank you.

`TRANS_BY_GPT3`

winlinvip:
In addition, please branch from develop, submit to develop. Currently, version 4.0 has been released and no new features will be added, only bug fixes.

`TRANS_BY_GPT3`

mapengfei53:
Okay

`TRANS_BY_GPT3`

winlinvip:
麻烦重新提交个PR吧。

winlinvip:
You can continue using this PR. If you don't know how to operate it, you can contact me on WeChat.

`TRANS_BY_GPT3`

mapengfei53:
> You can continue using this PR. If you don't know how to operate, you can contact me on WeChat.

> I had something come up over the weekend, which delayed me from resubmitting. I apologize for the inconvenience.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -1013,19 +1014,50 @@ srs_error_t SrsConfDirective::parse_conf(SrsConfigBuffer* buffer, SrsDirectiveTy
...
-            srs_freep(err);
-            if ((err = directive->parse_conf(buffer, parse_block)) != srs_success) {
-                return srs_error_wrap(err, "parse dir");
+        if (args[0] == "include") {
+            if (args.size() < 2) {
+                return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "line %d: include is empty directive", buffer->line);
+            }
+
+            std::string file = args[1];
+            if (!srs_string_contains(file, "*", "?", "[")) {
```
winlinvip: Files that do not support special characters can return an error if there are any special characters.

This is relatively simple and doesn't have much impact.

`TRANS_BY_GPT3`

mapengfei53: I commented out the code that supports regular expression rules and re-added the method that only supports importing file names.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -1014,35 +1014,49 @@ srs_error_t SrsConfDirective::parse_conf(SrsConfigBuffer* buffer, SrsDirectiveTy
         }
         
         // build directive tree.
-        if (args[0] == "include") {
+        if (args.at(0) == "include") {
             if (args.size() < 2) {
                 return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "line %d: include is empty directive", buffer->line);
             }
+#if 0
```
winlinvip, line 1021: 死代码？请直接删除就好了，不要留着**以后用**的死代码，除了干扰没有其他的用处，想要实现功能总能找到最好的答案，代码并不是那么必要的。

winlinvip: Why is it `#if 0`? Dead code?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -1014,35 +1014,49 @@ srs_error_t SrsConfDirective::parse_conf(SrsConfigBuffer* buffer, SrsDirectiveTy
...
+                    srs_trace("config parse include %s", file.c_str());
+                    srs_freep(err);
+                    if ((err = conf->parse_include_file(file.c_str())) != srs_success) {
+                        return srs_error_wrap(err, "parse file");
+                    }
+                } else {
+                    unsigned long i = 0;
+                    glob_t glob_buf;
+
+                    if (glob((char *) file.c_str(), 0, NULL, &glob_buf) != 0) {
```
winlinvip: Do not support wildcards.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: Removed this section of code

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: Okay

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -2419,6 +2435,29 @@ srs_error_t SrsConfig::parse_file(const char* filename)
...
+        return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "empty include config");
+    }
+
+    SrsConfigBuffer buffer;
+
+    if ((err = buffer.fullfill(include_file.c_str())) != srs_success) {
+        return srs_error_wrap(err, "buffer fullfil");
+    }
+
+    if ((err = parse_include_buffer(&buffer)) != srs_success) {
```
winlinvip: This function does not need to be encapsulated once, it can be called directly because it is only called in this one place.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -2979,6 +3018,18 @@ srs_error_t SrsConfig::parse_buffer(SrsConfigBuffer* buffer)
     return err;
 }
 
+srs_error_t SrsConfig::parse_include_buffer(SrsConfigBuffer *buffer)
+{
+    srs_error_t err = srs_success;
+
+    // Parse root tree from buffer.
+    if ((err = root->parse(buffer, this)) != srs_success) {
```
winlinvip: Does 'include' only support top-level ones, such as: 

```
listen 1935;
include rtmp.conf rtc.conf;
```

And does not support this kind:

```
listen 1935;
rtc_server {
  include rtc.conf;
}
```

It seems that the sub-file is being parsed using root. However, `SrsConfDirective::parse_conf` can be recursively called, which can cause issues in the later cases. It would be helpful to test it using utest.

`TRANS_BY_GPT3`

mapengfei53: Because I used root to parse the subfiles, the missing support can only be provided in a top-level manner. I will think about how to support the following format at the same time. The missing consideration mentioned earlier was not comprehensive enough.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: Okay

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: If it is a configuration inside a block, root cannot be used to parse the subfile. This needs to be adjusted in the code. Thank you for the reminder.

`TRANS_BY_GPT3`

winlinvip: Without uTest, it will never be comprehensive. These are just the problems that can be identified, there are still many problems that cannot be seen.

Please add uTest.

`TRANS_BY_GPT3`

mapengfei53: I added the uTest feature, if you have time, please help me check if it works. Thank you.

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_config.cpp
```diff
@@ -3622,3 +3622,98 @@ VOID TEST(ConfigMainTest, CheckVhostConfig5)
     }
 }
 
+VOID TEST(ConfigMainTest, CheckIncludeConfig)
+{
+    srs_error_t err;
+
+    if (true) {
+        MockSrsConfig conf;
+        HELPER_ASSERT_SUCCESS(conf.parse("include ./conf/include_test/include.conf;"));
```
winlinvip, line 3715: Do not write the configuration of the test under the 'conf' directory, you need to modify the mock to support 'include'. That means reading this mock value from a memory string. You can set it, for example:

```
MockSrsConfig conf;
`conf.mock_include("./conf/include_test/include.conf", "configuration content");`
```

When reading this file, it is still recommended to use mock to read it.

The code logic may need to support mocking and should not directly read from the file.

`TRANS_BY_GPT3`

mapengfei53: I roughly understand the meaning, I will try to make some modifications.

`TRANS_BY_GPT3`

mapengfei53: I have resubmitted the modifications, thank you.

`TRANS_BY_GPT3`

winlinvip: It is better to write the test content to the conf file. Of course, this is not the practice of mocking. Mocking should not interfere with normal logic, otherwise it would be more harm than good.

All mocking logic should only be in `MockSrsConfig`, and should not appear in `SrsConfig`. It can be implemented by overriding C++ virtual functions.

If you carefully examine the implementation of `MockSrsConfig`, you will notice that it does not read files, but directly reads string content. The logic of `SrsConfig` has not changed, it still implements the normal file reading and parsing, only `MockSrsConfig` has changed the file reading logic.

What you need to do is similar.

`TRANS_BY_GPT3`

mapengfei53: Okay, I will study it.

`TRANS_BY_GPT3`

winlinvip: I didn't look at the code carefully, but the specific idea should be like this: 'include' mainly needs to be mocked, which means reading content from a specified file. So we need to mock this part of its ability. By default, it uses file reading, but it can be changed to read from a string using a mock method.

`TRANS_BY_GPT3`

mapengfei53: I have made some modifications to the code again, this should be almost done. Hahaha

`TRANS_BY_GPT3`

winlinvip: 👍 Very close, hahaha, great job,
Just need to make a few changes, will reply to you later.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -1013,19 +1013,54 @@ srs_error_t SrsConfDirective::parse_conf(SrsConfigBuffer* buffer, SrsDirectiveTy
...
+                if (type != parse_block) {
+                    if ((err = conf->parse_include_file(file.c_str())) != srs_success) {
+                        return srs_error_wrap(err, "parse file");
+                    }
+                } else {
+                    if (file.empty()) {
+                        return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "empty include config");
+                    }
+
+                    SrsConfigBuffer* config_buffer = conf->get_buffer_from_include_file(file.c_str());
```
winlinvip: This place, do not use this kind of release method that is prone to problems, but should be changed to:

```c++
SrsConfigBuffer* config_buffer = conf->get_buffer_from_include_file(file.c_str());
SrsAutoFree(SrsConfigBuffer, config_buffer);

if(config_buffer == NULL) {
    return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "empty include config");
} else {
    if ((err = parse_conf(config_buffer, parse_file, conf)) != srs_success) {
        return srs_error_wrap(err, "parse include buffer");
    }
}
```

Take a look and see if it's super simple, it won't release repeatedly, and it won't cause memory leaks.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -1013,19 +1013,54 @@ srs_error_t SrsConfDirective::parse_conf(SrsConfigBuffer* buffer, SrsDirectiveTy
...
+                std::string file = args.at(i);
+
+                srs_trace("config parse include %s", file.c_str());
+                srs_freep(err);
+                if (type != parse_block) {
+                    if ((err = conf->parse_include_file(file.c_str())) != srs_success) {
+                        return srs_error_wrap(err, "parse file");
+                    }
+                } else {
+                    if (file.empty()) {
```
winlinvip: This error check should be moved up a bit, this is how you are currently doing it: 

```c++
    if (args.at(0) == "include") {
            for (int i = 1; i < (int)args.size(); i++) {
                std::string file = args.at(i);
                if (type != parse_block) {
                    if ((err = conf->parse_include_file(file.c_str())) != srs_success) {
                } else {
                    if (file.empty()) {
                        // error
```

And in `parse_include_file`, you checked the `file` parameter again, returning an error if it is empty. It should have been checked together earlier.

```c++
    if (args.at(0) == "include") {
            for (int i = 1; i < (int)args.size(); i++) {
                std::string file = args.at(i);
                if (file.empty()) {
                    // error
                }

                if (type != parse_block) {
                    if ((err = conf->parse_include_file(file.c_str())) != srs_success) {
                } else {
```

Not only is it repetitive, but it also leads to an additional level of nesting.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -2419,6 +2454,45 @@ srs_error_t SrsConfig::parse_file(const char* filename)
...
+{
+    srs_error_t err = srs_success;
+
+    std::string file = filename;
+
+    if (file.empty()) {
+        return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "empty include config");
+    }
+
+    SrsConfigBuffer buffer;
```
winlinvip: This place also needs to create `SrsConfigBuffer`, and `get_buffer_from_include_file` also needs to create `SrsConfigBuffer`. Essentially, both require creation, so why split them into two functions? This part can actually be improved.

Or, to put it another way, `parse_include_file` is for use by `root`, while `get_buffer_from_include_file` is for use by `conf`. Actually, there is no difference between the two, they are both `SrsDirective`. There is no need to specifically prepare a logic for `root`. Just think about it.

Is it possible to just call one logic: `parse_conf(config_buffer, parse_file, root)`, instead of specifically preparing a logic for `root`? It seems unnecessary to do so.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: Okay, I'll organize it, thank you.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: learned

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
mapengfei53: This, let me organize it again, thank you for your support.

`TRANS_BY_GPT3`

winlinvip: 👍

trunk/src/app/srs_app_config.cpp
winlinvip: 👍

trunk/src/app/srs_app_config.cpp
winlinvip: 👍

trunk/src/app/srs_app_config.cpp
mapengfei53: The main reason for separating these two is currently for the inheritance implementation of the utest module.

`TRANS_BY_GPT3`

winlinvip:
Thank you very much for making so many improvements, hahaha.

I think this PR has been done very well, thank you very much.

However, there is still a slight gap from what I want. I will make some further improvements based on your work. 😄

`TRANS_BY_GPT3`

mapengfei53:
> 非常感谢，能改进这么多次，哈哈哈。
> 
> 我想这个PR也已经做到非常的好了，非常感谢。
> 
> 不过还是离我要的有一点点差距，我会在你的基础上再改进一点点。 😄



> 非常感谢，能改进这么多次，哈哈哈。
> 
> 我想这个PR也已经做到非常的好了，非常感谢。
> 
> 不过还是离我要的有一点点差距，我会在你的基础上再改进一点点。 😄

jinleileiking:
Not useful, now everyone uses Helm to do it.

`TRANS_BY_GPT3`

winlinvip (approved):

PR #2880, opened 2022-01-12
bugfix: SrsMetaCache memleak; getaddrinfo use delete.

bluestn:
## Summary

1. SrsMetaCache memleak.
2. use freeaddrinfo instread of delete to release addrinfo pointer.


trunk/src/protocol/srs_service_st.cpp
```diff
@@ -171,29 +171,32 @@ srs_error_t srs_tcp_connect(string server, int port, srs_utime_t tm, srs_netfd_t
     hints.ai_socktype = SOCK_STREAM;
     
     addrinfo* r  = NULL;
-    SrsAutoFree(addrinfo, r);
```
winlinvip, line 174: Why can't we use `SrsAutoFree`?

If you need to use `free` instead of `delete` for deallocation, you can use `SrsAutoFreeF`.

Automatic freeing can significantly improve code readability.

`TRANS_BY_GPT3`

bluestn:
When looking at the man page for getaddrinfo, it states that freeaddrinfo should be used to release the memory.
The memory allocated by getaddrinfo may not necessarily be allocated using new, so it is inappropriate to use delete with SrsAutoFree.
Furthermore, the addrinfo allocated within getaddrinfo is a linked list, so SrsAutoFree cannot release the entire addrinfo chain at once.
At 2022-01-13 12:20:28, "Winlin" ***@***.***> wrote:


@winlinvip commented on this pull request.


In trunk/src/protocol/srs_service_st.cpp:


> @@ -171,29 +171,32 @@ srs_error_t srs_tcp_connect(string server, int port, srs_utime_t tm, srs_netfd_t


     hints.ai_socktype = SOCK_STREAM;


     


     addrinfo* r  = NULL;


-    SrsAutoFree(addrinfo, r);






Why can't we use SrsAutoFree?
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.
You are receiving this because you authored the thread.Message ID: ***@***.***>

`TRANS_BY_GPT3`

winlinvip:
If it is not possible to release using free or delete, you must use freeaddrinfo to release. In that case, it can be changed like this: 

```
string do_srs_dns_resolve(string host, int& family, addrinfo*& r)
{
    addrinfo hints;
    memset(&hints, 0, sizeof(hints));
    hints.ai_family = family;

    if(getaddrinfo(host.c_str(), NULL, &hints, &r)) {
        return "";
    }
    
    char shost[64];
    memset(shost, 0, sizeof(shost));
    if (getnameinfo(r->ai_addr, r->ai_addrlen, shost, sizeof(shost), NULL, 0, NI_NUMERICHOST)) {
        return "";
    }

   family = r->ai_family;
   return string(shost);
}

string srs_dns_resolve(string host, int& family)
{
    addrinfo* r = NULL;
    string r0 = do_srs_dns_resolve(host, family, r);
    freeaddrinfo(r);

    return r0;
}
```

Transform the resource allocation and handling into a sub-function, and then release it once at the calling location.

If you find it troublesome, you can use the goto method in C, you can refer to FFmpeg for specific details.

`TRANS_BY_GPT3`

winlinvip:
There is another way, to improve `SrsAutoFree`, allowing to pass a release function into it.

`TRANS_BY_GPT3`

winlinvip:
I add a macro `SrsAutoFreeH` to allow release with hook: 3881c4c77

```
    addrinfo* r = NULL;
    SrsAutoFreeH(addrinfo, r, freeaddrinfo);
```

winlinvip:
@bluestn Could you please make the changes according to the new modifications? Can you also merge the 4.0 release branch?

`TRANS_BY_GPT3`

winlinvip:
@bluestn If you don't know how to operate git, you can contact me on WeChat.

There is no need to resubmit a new PR, you can continuously update this PR.

`TRANS_BY_GPT3`

PR #2886, opened 2022-01-22
bugfix: SrsAutoFree addrinfo

bluestn:
bugfix: getaddrinfo return cannot freed by delete


winlinvip (approved):

winlinvip:
@bluestn Good job! This patch is simple and stable enough. 👍

PR #2919, opened 2022-02-15
在full.conf 更新include指令的使用方式说明

mapengfei53:
## Summary

在full.conf 更新include指令的使用方式说明

## Details

更新include指令的使用方式说明，协助正确使用配置。具体可参考[https://github.com/ossrs/srs/issues/1399](url)

### DIFF
--- a/trunk/conf/full.conf
+++ b/trunk/conf/full.conf
@@ -2083,6 +2083,47 @@ vhost stream.transcode.srs.com {
 }
 
 #############################################################################################
+# Support include directive for config file
+#
+# In the config file, the include file can be anywhere in order to enhance the readability 
+# of the config file and enable the reuse of part of the config file. 
+# When using include files, make sure that the included files themselves have the correct SRS syntax, 
+# that is, config directives and blocks, and then specify the paths to these files.
+#
+# If load a config file.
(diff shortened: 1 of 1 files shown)


trunk/conf/full.conf
```diff
@@ -2082,6 +2082,47 @@ vhost stream.transcode.srs.com {
...
+# Example 2:
+#   vhost __defaultVhost__ {
+#       transcode {
+#           enabled     on;
+#           ffmpeg      ./objs/ffmpeg/bin/ffmpeg;
+#
+#           include     ./conf/engines/engine.ffhd.conf;
+#       }
+#   }
+#############################################################################################
```
winlinvip, line 2156: Consider implementing srs.conf using include.

```
include.conf
include.vhost.conf
```

Then you can write it simpler in full.conf.

```
# See include.conf and include.vhost.conf
include include.vhost.conf;
```

The main purpose of full.conf is to express the meaning and it will not be used directly, but it should also be close to actual usage.

Other configurations can be used directly, representing different scenarios, and should be able to be used directly in that scenario.

`full.conf` contains all our configurations, so it is convenient for searching, similar to an index.

`TRANS_BY_GPT3`

PR #2921, opened 2022-02-16
Fix Stream statistics bug

mapengfei53:
## Summary

Fix Stream statistics bug

## Details

Error number of clients is displayed after the stream is pushed again。[#2903](https://github.com/ossrs/srs/issues/2903)


mapengfei53:
The main reason for this issue is that when the streaming is disconnected, all the statistical information of this stream is cleared (previously, there was an issue in the code, but currently, the issue cannot be accessed, so it is unclear what problem this code segment is addressing). The current modification is to only clear the statistical information when the number of streams is 0. Previously, the code cleared the stream information but lacked the stream's free, which I believe also leads to a memory leak issue.

`TRANS_BY_GPT3`

winlinvip:
Fixed by https://github.com/ossrs/srs/commit/e9d6601f7d069c6762df7a107c56abefa69626f4

Issue #2682, opened 2021-10-16
WebRTC FLV


thegobot:
> 注意：不提供以下信息的Issue会被直接删除(Please follow issue template, or we will delete it)

> 注意：咨询和讨论请提交到SRS星球(Please ask question at) http://bbs.ossrs.net

**描述(Description)**

> 描述你遇到了什么问题(Please description your issue here)

1. SRS版本(Version): `4.0.178`
1. SRS的日志如下(Log):
```
xxxxxxxxxxxx
```
1. SRS的配置如下(Config):
```
listen              1940;
max_connections     1000;
daemon              off;
srs_log_tank        console;
srs_log_level verbose;

http_server {
    enabled         on;
    crossdomain     off;
    listen          8040;
    dir             ./objs/nginx/html;
}

http_api {
	crossdomain     off;
    enabled         on;
    listen          1945;
}
stats {
    network         0;
}
rtc_server {
    enabled on;
    listen 9999;
    # @see https://github.com/ossrs/srs/wiki/v4_CN_WebRTC#config-candidate
    candidate 178.76.25.21;
}


vhost webrtc {

    rtc {
        enabled     on;
        rtc_to_rtmp on;
    }

    http_remux {
        enabled     on;
        mount       /[app]/[stream].flv;

    }
}
```

**重现(Replay)**

> 重现Bug的步骤(How to replay bug?)

1. `Push WebRTC to server`
2. `Start playing FLV in Chrome with Flv.JS`

**期望行为(Expect)**
> 描述你期望发生的事情(Please describe your expectation)
Continuous playback, no pause

Some bugs
1. First, when downloading (stream.flv from SRS server), the server gives a 404 error, if you repeat it, the download will start. If you stop downloading and start snoring, then the server again gives a 404 error, after 5 times the server sends the file successfully
2. The video is interrupted approximately every 5 seconds. At the same time, errors like
`
net::ERR_INCOMPLETE_CHUNKED_ENCODING
[IOController] > Loader error, code = undefined, msg = Fetch stream meet Early-EOF
[TransmuxingController] > IOException: type = UnrecoverableEarlyEof, code = undefined, msg = Fetch stream meet Early-EOF
...
reload flv
....
net::ERR_FAILED 404
...
reload flv again
...
 net::ERR_INCOMPLETE_CHUNKED_ENCODING 200
[IOController] > Loader error, code = undefined, msg = Fetch stream meet Early-EOF
[...]

PR #2926, opened 2022-02-18
port  from config for transcoding

thegobot:
## Summary

port  from config for transcoding

// TODO: FIXME: use initialize to set req.

### DIFF
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -855,6 +855,13 @@ srs_error_t SrsOriginHub::initialize(SrsLiveSource* s, SrsRequest* r)
     req_ = r;
     source = s;
     
+    int port; string ip;
+    std::vector<std::string> listens = _srs_config->get_listens();
+    if(listens.size() > 0){
+	srs_parse_endpoint(listens[0], ip, port);
+	req->port = port;
+    }	
+	
     if ((err = format->initialize()) != srs_success) {
         return srs_error_wrap(err, "format initialize");
     }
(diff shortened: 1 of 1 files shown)


trunk/src/app/srs_app_source.cpp
```diff
@@ -855,6 +855,13 @@ srs_error_t SrsOriginHub::initialize(SrsLiveSource* s, SrsRequest* r)
     req_ = r;
     source = s;
     
+    int port; string ip;
+    std::vector<std::string> listens = _srs_config->get_listens();
+    if(listens.size() > 0){
+	srs_parse_endpoint(listens[0], ip, port);
+	req->port = port;
```
winlinvip: Thanks for patch 👍 

Please check the CR is failed: [https://github.com/ossrs/srs/runs/5241353852?check_suite_focus=true](https://github.com/ossrs/srs/runs/5241353852?check_suite_focus=true)

Seems to fix https://github.com/ossrs/srs/issues/2682#issuecomment-945363305 ?

It's because the variable substituion for transcoding, we should fix it there, rather than changing the req port.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -855,6 +855,13 @@ srs_error_t SrsOriginHub::initialize(SrsLiveSource* s, SrsRequest* r)
     req_ = r;
     source = s;
     
+    int port; string ip;
+    std::vector<std::string> listens = _srs_config->get_listens();
+    if(listens.size() > 0){
+	srs_parse_endpoint(listens[0], ip, port);
+	req_->port = port;
```
winlinvip, line 862: It seems that it's caused by the transcoding variable substitution, so we should fix the transcoding.

We MUST NOT change the port to listen port in Request, because the port is request.

Does that make sense?

winlinvip:

PR #2928, opened 2022-02-18
Fix parse_file function did not handle err

mapengfei53:
## Summary

Fix parse_file function did not handle err

## Details

配置文件 live.conf
```
listen              1935;
max_connections     1000;
srs_log_tank        console;
daemon              off;

http_api {
    enabled         on;
    listen          1985;
}

http_server {
    enabled         on;
    listen          8080;
    dir             ./objs/nginx/html;
}

vhost __defaultVhost__ {
    hls {
        enabled         on;
    }
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].flv;
    }
#} ----------------------------假定此花括号缺失

```
运行加载配置
```
./objs/srs -c ./conf/live.conf
```
运行并不会报错，提示配置文件异常。

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -2040,14 +2040,14 @@ srs_error_t SrsConfig::parse_options(int argc, char** argv)
         // we check it when user requires check config file.
         if (err == srs_success && (err = srs_config_transform_vhost(root)) == srs_success) {
             if ((err = check_config()) == srs_success) {
-                srs_trace("the configuration file %s syntax is ok", config_file.c_str());
-                srs_trace("configuration file %s test is successful", config_file.c_str());
+                srs_trace("the config file %s syntax is ok", config_file.c_str());
+                srs_trace("config file %s test is successful", config_file.c_str());
                 exit(0);
             }
         }
 
-        srs_trace("invalid configuration: %s in %s", srs_error_summary(err).c_str(), config_file.c_str());
-        srs_trace("configuration file %s test is failed", config_file.c_str());
+        srs_trace("invalid config%s in %s", srs_error_summary(err).c_str(), config_file.c_str());
+        srs_trace("config file %s test is failed", config_file.c_str());
         exit(srs_error_code(err));
     }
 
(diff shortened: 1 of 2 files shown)


trunk/src/app/srs_app_config.cpp
```diff
@@ -2033,29 +2033,27 @@ srs_error_t SrsConfig::parse_options(int argc, char** argv)
...
-        int ret = srs_error_code(err);
-        srs_freep(err);
-        exit(ret);
+    if ((err = parse_file(config_file.c_str())) != srs_success) {
+        return srs_error_wrap(err, "invalid config");
     }
-    
+
     // transform config to compatible with previous style of config.
     if ((err = srs_config_transform_vhost(root)) != srs_success) {
```
winlinvip: When testing the configuration: if an error occurs here, then the previous result is to print the error message and exit.

Now this logic is not like that anymore. There are still problems with this modification.

`TRANS_BY_GPT3`

mapengfei53: If this function returns an error, it will directly exit the main function, so there is no need to separately write an exit, right?

`TRANS_BY_GPT3`

winlinvip: This is how it should be: during testing, if there is a configuration error, it should not return such a complex error.

You can study the `-t` option of nginx to test the configuration file. Generally, it is recommended to test it before applying the configuration. Therefore, the printed message should not include a stack trace. At this point, the error is a configuration error, not a logical error.

Please try using the `-t` option of nginx, let's learn from it.

`TRANS_BY_GPT3`

mapengfei53: Okay, no problem, I'll make some adjustments.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_error.cpp
```diff
@@ -85,7 +85,9 @@ std::string SrsCplxError::summary() {
 
         SrsCplxError* next = this;
         while (next) {
-            ss << " : " << next->msg;
+            if (!next->wrapped) {
```
winlinvip: Do not make changes in a PR that are unrelated to yourself. This modification is for the public Error module and has nothing to do with the configuration file parsing.

`TRANS_BY_GPT3`

mapengfei53: Because when outputting errors, this function srs_error_summary is used, so I made some changes. I will adjust it and use a different method.

`TRANS_BY_GPT3`

winlinvip (approved):

PR #2939, opened 2022-02-28
fix coverity scan

fastfading:
CID 42517 (#2 of 2): Uninitialized scalar variable (UNINIT)
4. uninit_use_in_call: Using uninitialized value addr. Field addr.sin_zero is uninitialized when calling bind.

## Summary

## Details

Add more details about this PR.


winlinvip:
Thank you for your patch 👍 

SRS 2.0 is archived, so we don't update it anymore.

PR #2948, opened 2022-03-04
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/trunk/3rdparty/signaling/www/demos/js/srs.sdk.js
+++ b/trunk/3rdparty/signaling/www/demos/js/srs.sdk.js
@@ -154,7 +154,7 @@ function SrsRtcPublisherAsync() {
 
             return {
                 apiUrl: apiUrl, streamUrl: streamUrl, schema: schema, urlObject: urlObject, port: port,
-                tid: Number(parseInt(new Date().getTime()*Math.random()*100)).toString(16).substr(0, 7)
+                tid: Number(parseInt(new Date().getTime()*Math.random()*100)).toString(16).slice(0, 7)
             };
         },
         parse: function (url) {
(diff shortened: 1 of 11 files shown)


trunk/3rdparty/signaling/www/demos/js/srs.sdk.js
```diff
@@ -392,19 +392,19 @@ function SrsRtcPlayerAsync() {
                 .replace("rtc://", "http://");
 
             var vhost = a.hostname;
-            var app = a.pathname.substr(1, a.pathname.lastIndexOf("/") - 1);
-            var stream = a.pathname.substr(a.pathname.lastIndexOf("/") + 1);
+            var app = a.pathname.substring(1, a.pathname.lastIndexOf("/") - 1);
```
winlinvip, line 395: There is a bug:

```
'1234567890'.substr(1,3)
'234'

'1234567890'.substring(1,3)
'23'
```

The two are not equal.

CommanderRoot: You are absolutely right. Not sure why but I clearly didn't pay enough attention there.
I have now rechecked the changes and fixed the issues

winlinvip (approved):

winlinvip:
@CommanderRoot Thanks a lot. I have merged this PR, please test the demo of develop branch, and ensure that it works. 👍

PR #2960, opened 2022-03-08
SRT: Fix the cache and update for SequenceHeader.

dpdp2020:
## Summary

1. Problem Reproduction:

The camera pushes the audio and video streams through SRT to SRS. During the testing process, it was found that the browser playback end cannot last for a night and will be interrupted. Only by restarting SRS or the camera can it be restored.

2. Problem cause:

The cause is due to SequenceHeader updates and caching issues.

## Details

Audio Flow

Video Flow

---------

`TRANS_BY_GPT3`


dpdp2020:
Modify PR submitted to 4.0 release

`TRANS_BY_GPT3`

winlinvip:
NICE 👍 Thanks!

winlinvip:

PR #2967, opened 2022-03-14
fix crash in edge mode.

johzzy:
## Summary

fix crash in edge mode.

## Details
crash if run edge, use `conf/edge.conf` or this config

```
    cluster {
        mode remote;
        origin xxx;
    }
```


winlinvip:
Please provide more background information.

1. The code fix is for a problem with a call.
2. The configuration and reproduction steps are just a normal edge configuration.
3. It is not clear what exactly happened.

`TRANS_BY_GPT3`

PR #2970, opened 2022-03-17
UTest: Upgrade gtest to 1.11 and support clion.

winlinvip:
Upgrade gtest to 1.11 and support clion.

1. Upgrade gtest from 1.6 to 1.11, because mac build failed.
2. Use source code at `3rdparty/gtest-fit`, remove `3rdparty/gtest-1.6.0.zip`.
3. Support CLion build and debug.


winlinvip:
The new gtest relies on C++11, but CentOS7's gcc is 4.8, which does not support many of the features of C++11. Therefore, it was upgraded to GCC 7.

As a result, a bunch of issues were discovered. Fortunately, there is a docker image, `ossrs/srs:dev-gcc7`, but it was found that even docker cannot pass.

After making various changes, it was finally resolved. Sigh, my heart is exhausted. No one was willing to deal with this compilation failure issue, let alone find someone to work on it together.

C++11 can be used, but be careful and don't use all the features. Don't think that this thing is only left with good aspects.

Usually, after we submit the code, we don't have time to look at these compilation failure issues.

Even if everyone has time to manage, they probably don't want to look at these compilation failure issues.

`TRANS_BY_GPT3`

PR #2988, opened 2022-03-31
Release the dependency of jQuery

lin-calvin:
And can i speak Chinese here? my English is  bad


winlinvip:
https://github.com/ossrs/srs/pull/4367

Issue #2999, opened 2022-04-10
[Feature] HLS: Support delay cleanup for TS files.


winlinvip:
Some player start to play from the first ts file, which might be cleanup.

It's better to delay cleanup the ts files.

<img width="1296" alt="image" src="https://user-images.githubusercontent.com/2777660/162606651-fb17320d-60d5-4d38-a196-157cb65c673c.png">

mapengfei53:
Currently, the delay deletion feature is expected to be modified into either Method 1: `delay_cleanup on`; or Method 2: `delay_cleanup 60`; (unit: s). In the first method, users can only configure to turn it on or off, with a fixed internal duration. In the second method, users can customize the delay deletion duration.

winlinvip:
Seems not a very useful feature.

PR #3001, opened 2022-04-11
HLS: Added delay cleanup old expired ts files

mapengfei53:
Some player start to play from the first ts file, which might be cleanup.
It's better to delay cleanup the ts files. issues [#2999](https://github.com/ossrs/srs/issues/2999)

### DIFF
--- a/trunk/src/app/srs_app_fragment.cpp
+++ b/trunk/src/app/srs_app_fragment.cpp
@@ -5,6 +5,7 @@
 //
 
 #include <srs_app_fragment.hpp>
+#include <srs_app_hybrid.hpp>
 
 #include <srs_kernel_utility.hpp>
 #include <srs_kernel_log.hpp>
@@ -12,6 +13,7 @@
 
 #include <unistd.h>
 #include <sstream>
+
 using namespace std;
 
 SrsFragment::SrsFragment()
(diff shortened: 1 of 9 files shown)


trunk/src/app/srs_app_fragment.cpp
```diff
@@ -212,22 +212,49 @@ void SrsFragmentWindow::shrink(srs_utime_t window)
...
+
+        expired_fragments.clear();
+    } else {
+        srs_utime_t duration = 0;
+
+        int remove_index = -1;
+
+        for (int i = (int)expired_fragments.size() - 1; i >= 0; i--) {
+            SrsFragment* fragment = expired_fragments[i];
+            duration += fragment->duration();
```
winlinvip: Here, duration is used as the time for scheduled cleaning, which is difficult to understand, and there are too many if-else statements, making it very challenging for future maintenance.

A better solution is to send the task of deleting slice files to an asynchronous worker, which will handle recording the time and performing the cleaning.

It's something like this:

```cpp
void SrsFragmentWindow::clear_expired(bool delete_files, srs_utime_t delay_time)
{
  for (it = expired_fragments.begin(); it != expired_fragments.end(); ++it) {
    SrsFragment* fragment = *it;
    if (!delete_files) {
      continue;
    }
    
    if (delay_time) {
      err = cleanup_worker->unlink_file(filename);
    } else {
      err = fragment->unlink_file();
    }
    if (err != srs_success) {
      srs_warn("Unlink ts failed, %s", srs_error_desc(err).c_str());
      srs_freep(err);
    }
```

You can use a global `cleanup_worker`, which is relatively simple.

`TRANS_BY_GPT3`

mapengfei53: I added an asynchronous clock to handle expired ts files. If you have time, please take a look. Thank you.

`TRANS_BY_GPT3`

winlinvip:

PR #3010, opened 2022-04-15
SRT: Support Coroutine Native SRT over ST.

xiaozhihong:
1. Rewrite SRT to support Coroutine Native, based on ST.
2. Add SRT unit tests.
3. Error based handler.


trunk/conf/full.conf
```diff
@@ -364,11 +364,14 @@ srt_server {
...
     connect_timeout 4000;
-    peerlatency 300;
-    recvlatency 300;
-    # Default app for vmix, see https://github.com/ossrs/srs/pull/1615
-    # default: live
-    default_app live;
+    peerlatency 0;
+    recvlatency 0;
+    # TODO: FIXME: Add comments.
+    latency 0;
```
winlinvip, line 391: Please add explanations for these new configurations, including their purposes and default values, for example:

```
# Set the default value for peerlatency and recvlatency.
# Where 0 means xxxxxxxxx.
# Default: 0
latency 0;
```

> Note: Please also provide explanations for the previous parameters. It is important to keep the documentation up to date to facilitate future usage, maintenance, and parameter understanding. Even though some people may not read the documentation, having it available allows for filtering out those who do read it (as there are still some who do), guiding everyone to consult the documentation, and requesting everyone to refer to it. Without documentation, there is no next step, and no possibility for continuous development.

Additionally, I noticed that some default values for certain configurations have changed. It would be helpful if you could update the [Wiki](https://github.com/ossrs/srs/wiki/v5_CN_SRTParams) where these parameters were previously linked.

```
# For detail parameters, please read wiki:
#   https://github.com/ossrs/srs/wiki/v5_CN_SRTParams
#   https://github.com/ossrs/srs/wiki/v5_EN_SRTParams
```

We are now on the 5.0 version of the Wiki.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -6882,6 +6894,20 @@ int SrsConfig::get_srto_conntimeout() {
     return atoi(conf->arg0().c_str());
 }
 
+int SrsConfig::get_srto_peeridletimeout() {
+    static int DEFAULT = 10000;
```
winlinvip: All the time, basic types should not be used, but instead use `srs_utime_t`, which has a unified unit of us. Because there is no time type in C++, here `10000` is not known literally what unit it is, so it must be checked in the documentation when used.

Comparing the following two functions:

```
srs_utime_t SrsConfig::get_rtc_stun_timeout(string vhost)
int SrsConfig::get_srto_peeridletimeout() {
```

When we use it, if we use a primitive type, such as `int timeout`, we still cannot determine its unit. We have to refer to `SET_SRT_OPT` to know if it is used correctly. It is very easy to confuse the unit and cause problems.

```
    if ((err = srs_srt_set_peer_idle_timeout(listener_->fd(), _srs_config->get_srto_peeridletimeout())) != srs_success) {

srs_error_t srs_srt_set_peer_idle_timeout(SRTSOCKET srt_fd, int timeout)
{
    SET_SRT_OPT(srt_fd, SRTO_PEERIDLETIMEO, timeout);
    return srs_success;
}
```

Looking at the usage of `srs_utime_t`, we can see that it is converted to milliseconds using `srsu2msi`, or in the case of `is_alive`, since everyone's unit is `srs_utime_t`, there cannot be a problem of inconsistent units.

```
    session_timeout = _srs_config->get_rtc_stun_timeout(req_->vhost);

    srs_trace("RTC: session pub=%u, sub=%u, to=%dms connection established", publishers_.size(), players_.size(),
        srsu2msi(session_timeout));

bool SrsRtcConnection::is_alive()
{
    return last_stun_time + session_timeout > srs_get_system_time();
}
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -6756,46 +6764,46 @@ unsigned short SrsConfig::get_srt_listen_port()
     return (unsigned short)atoi(conf->arg0().c_str());
 }
 
-bool SrsConfig::get_srt_mix_correct() {
-    static bool DEFAULT = true;
+int SrsConfig::get_srto_maxbw() {
```
winlinvip: The code style of the function is inconsistent, please place the braces on the next line.

```
int SrsConfig::get_srto_maxbw() 
{
```

Please also make changes in other places.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.hpp
```diff
@@ -658,6 +662,11 @@ class SrsConfig
     virtual std::string get_default_app_name();
     // Get the mix_correct
     virtual bool get_srt_mix_correct();
+public:
+    SrsConfDirective* get_srt(std::string vhost);
```
winlinvip, line 664: The function `get_srt` should be private and only used in the config.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_srt_source.hpp
```diff
@@ -0,0 +1,210 @@
...
+    // previous source id.
+    SrsContextId _pre_source_id;
+    SrsRequest* req;
+    // To delivery packets to clients.
+    std::vector<SrsSrtConsumer*> consumers;
+    bool can_publish_;
+    std::vector<ISrsSrtSourceBridge*> bridgers_;
+};
+
+class SrsSrtFromRtmpBridge : public ISrsLiveSourceBridger, public ISrsStreamWriter
```
winlinvip: RTMP to SRT, this scenario is likely very niche. SRT is already used relatively less, primarily in streaming scenarios. The conversion from RTMP to SRT is probably even less common.

I suggest simply removing the RTMP to SRT functionality altogether. This way, it would save a significant amount of code and eliminate the need for writing bilingual documentation.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_service_st_srt.cpp
```diff
@@ -0,0 +1,830 @@
...
+        if (iter == fd_sockets_.end()) {
+            srs_assert(false);
+        }
+
+        SrsSrtSocket* srt_skt = iter->second;
+        srs_assert(srt_skt != NULL);
+
+        // notify error, don't notify read/write event.
+        if (event.events & SRT_EPOLL_ERR) {
+            srt_skt->notify_error();
```
winlinvip: The call chain is as follows. It seems that detailed error information has not been provided. Please check if it is possible to obtain the error.

```
            srt_skt->notify_error();

void SrsSrtSocket::notify_error() {
    has_error_ = true;

srs_error_t SrsSrtSocket::check_error() {
    if (has_error_) {
        return srs_error_new(ERROR_SRT_IO, "has error");
    }
```

For example, it is important to determine whether the IO error is caused by the server disconnecting actively (indicating an issue with the client's data) or by the client initiating the disconnection (such as sending a disconnect packet). In any case, providing more information would be helpful.

Sometimes, there are frequent mentions of this disconnection error, but the reason for the disconnection is unknown.

`TRANS_BY_GPT3`

winlinvip: There are two more related log issues:

```
19:09:44.950031!W:SRT.c: HSREQ/rcv: Agent did not set rcv-TSBPD - ignoring proposed latency from peer
```
Can we find a way to hook the SRT log into SRS? This way, we will have the connection ID and won't miss anything when using grep. If there are multiple sessions, we won't be able to distinguish which one it is.

```
[2022-06-07 19:09:44.954][Trace][49469][857ikont] accept srt client from 127.0.0.1:49765, fd=251271634
[2022-06-07 19:09:44.956][Trace][49469][0qg31q2s] @srt, streamid=#!::r=live/livestream?secret=efi3ne2r3r,m=publish, stream_url=/live/livestream, vhost=, app=live, stream=livestream, param=secret=efi3ne2r3r
[2022-06-07 19:09:44.957][Trace][49469][0qg31q2s] new srt source, stream_url=/live/livestream
```

The log ID for this SRT session should be 0qg31q2s, while the first message 857ikont is detached from any session. Let's see if there is a way to merge it so that grep won't miss it either.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_srt.cpp
```diff
@@ -0,0 +1,912 @@
...
+    if ((err = srs_srt_nonblock(srt_fd)) != srs_success) {
+        return srs_error_wrap(err, "nonblock");
+    }
+
+    if (srt_bind(srt_fd, r->ai_addr, r->ai_addrlen) == -1) {
+        return srs_error_new(ERROR_SOCKET_BIND, "bind");
+    }
+
+    if (srt_listen(srt_fd, 100) == -1) {
+        return srs_error_new(ERROR_SOCKET_LISTEN, "listen");
```
winlinvip, line 63:
```
int srt_close        (SRTSOCKET u);

SrsSrtListener::~SrsSrtListener()
{
    srs_freep(trd_);
    srs_freep(srt_skt_);
    srt_close(lfd_);
}
```

I have added a note to this, it's not easy to change because I don't know how to handle the error. I have given it to you.

```
    SRTSOCKET srt_fd = 0;
    if ((srt_fd = srt_create_socket()) < 0) {
        return srs_error_new(ERROR_SOCKET_CREATE, "create srt socket");


    if (srt_listen(srt_fd, 100) == -1) {
        return srs_error_new(ERROR_SOCKET_LISTEN, "listen");
    }
```

The error handling here is problematic. I have made a note about it, but I don't know how to handle it. Please make the necessary changes.

Especially, be very careful with the 'close' operation. For example, ST's 'st_netfd_close' actually returns an error. If the file descriptor is still being read or written by a coroutine when it is closed, problems may arise. So, if this is ignored, strange issues may occur. It is better to identify which coroutine is still performing read or write operations and which coroutine is being closed. This way, the problem can be identified on the spot. If not handled properly, the issue will be impossible to trace. Currently, there is an issue caused by this.

`TRANS_BY_GPT3`

winlinvip:
Merged by https://github.com/ossrs/srs/commit/1af30dea324d0f1729aabd22536ea62e03497d7d

PR #3040, opened 2022-05-16
bugfix: hls ts segment does not start with key frame for a stream with audio and video

peipeiguo:
bugfix: hls ts segment does not start with key frame for a stream with audio and video.

- my hls config as below:
```
vhost __defaultVhost__ {
    hls {
        enabled         on;
        hls_fragment    6;
        hls_td_ratio    1.0;
        hls_aof_ratio   1.0;
        hls_window      30;
        hls_m3u8_file   [app]/[stream]/[stream].m3u8;
        hls_ts_file     [app]/[stream]/[stream]-[2006][01][02][15][04]-[seq].ts;
        hls_ts_floor    off;
    }
}
```
- bug description as below:
For a stream with audio and video, some ts segment not start with key frame. 

- root cause is as below: 
`reap_segment` is triggered both in `write_audio` and `write_video`. 
`is_segment_absolutely_overflow` true when write audio. it is early then `is_segment_overflow` when write video.

- bug fix solution as below:
`is_segment_absolutely_overflow` should only check for  a stream with pure audio.


winlinvip:

PR #3067, opened 2022-06-15
HLS: Support config to disable HLS streaming.

duiniuluantanqin:
# Whether enable hls_ctx.

        # hls_ctx used to keep track of subsequent m3u8 requests.
        # if on, hls client info can be collected in stat module,
        # if on, on_play/on_stop can be called in http hook module.
        # otherwise not;

        # When request stopped, http hook or stat with special time effected, that is 2*hls_window, 

        # Default: off

### DIFF
--- a/trunk/conf/full.conf
+++ b/trunk/conf/full.conf
@@ -1371,6 +1371,19 @@ vhost hls.srs.com {
         # if off, do not write hls(ts and m3u8) when publish.
         # default: off
         enabled         on;
+
+        # Whether enable hls_ctx for HLS streaming, for which we create a "fake" connection for HTTP API and callback.
+        # For each HLS streaming session, we use a child m3u8 with a session identified by query "hls_ctx", it simply
+        # work as the session id.
+        # Once the HLS streaming session is created, we will cleanup it when timeout in 2*hls_window seconds. So it
+        # takes a long time period to identify the timeout.
+        # Now we got a HLS stremaing session, just like RTMP/WebRTC/HTTP-FLV streaming, we're able to stat the session
+        # as a "fake" connection, do HTTP callback when start playing the HLS streaming. You're able to do querying and
+        # authentication.
+        # Note that it will make NGINX edge cache always missed, so never enable HLS streaming if use NGINX edges.
+        # Default: on
+        hls_ctx on;
+
         # the hls fragment in seconds, the duration of a piece of ts.
         # default: 10
         hls_fragment    10;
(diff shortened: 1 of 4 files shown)


trunk/src/utest/srs_utest_http.cpp
```diff
@@ -1210,8 +1210,8 @@ VOID TEST(ProtocolHTTPTest, VodStreamHandlers)
         __MOCK_HTTP_EXPECT_STREQ(200, "Hello, world!", w);
     }
 
-    // should return "hls_ctx"
-    if (true) {
+    // TODO: should return "hls_ctx"
+    if (false) {
```
xiaozhihong: Are there still any issues with these two places?

`TRANS_BY_GPT3`

trunk/conf/full.conf
```diff
@@ -1460,6 +1460,14 @@ vhost hls.srs.com {
...
         # @remark It's optional.
         hls_key_url       https://localhost:8080;
+	# Whether enable hls_ctx.
+	# hls_ctx used to keep track of subsequent m3u8 requests.
+	# if on, hls client info can be collected in stat module,
+	# if on, on_play/on_stop can be called in http hook module.
+	# otherwise not;
+	# When request stopped, http hook or stat with special time effected, that is 2*hls_window, 
+	# Default: off
+	hls_ctx		on;
```
peipeiguo, line 1470: It says default off in above line, but value is `on` in this line. Suggest change the value as `off`.

peipeiguo (changes requested):

trunk/src/utest/srs_utest_http.cpp
duiniuluantanqin: This part is not easy to modify because the configuration file has `hls_ctx` disabled by default, so the test will fail here. Let me study how to dynamically modify the content of the configuration file.

`TRANS_BY_GPT3`

trunk/conf/full.conf
duiniuluantanqin: yeah, I will modifie it.

trunk/src/utest/srs_utest_http.cpp
winlinvip: It should be fine to change the default to open.

`TRANS_BY_GPT3`

winlinvip (approved):

winlinvip:
Merged as 2c569f568c3da9ecaf44cd29efb6e9912487f0cd 2c569f568c3da9ecaf44cd29efb6e9912487f0cd 2c569f568c3da9ecaf44cd29efb6e9912487f0cd

Issue #3041, opened 2022-05-18
RTMP推流RTC拉流时，黑屏只有声音，视频流全丢包了


zhOrange:
## 1. Description(描述)

在使用原始的rtmp2rtc.conf配置文件启动srs服务时，出现以下现象：
1. 使用obs进行rtmp时，rtmp拉流、rtc拉流均可以正常播放；
2. 第一次使用RTC推流时，rtmp拉流、rtc拉流也可以正常播放；
3. 但是，一旦使用了RTC推流后，关闭RTC推流网页的话，再次使用RTMP推流，RTC拉流就不能正常播放了，只有声音可以正常播放，视频只是黑屏。通过chrome://webrtc-internals查看视频流数据时，发现视频流数据还是可以正常接收，但全都Drop掉了，解码帧数始终为0。
4. 上述场景下，在RTC拉流时出现上述异常，但同时用RTMP拉流的话，可以正常播放；另外，如果重新换回RTC推流的话，RTMP拉流、RTC拉流也都可以正常播放。

**采用默认的srt2rtc.conf配置文件时，也会出现上述问题。**

## 3. SRS Log(日志):

```
[2022-05-14 20:59:15.164][Trace][221869][q8567159] TCP: clear zombies=1 resources, conns=3, removing=0, unsubs=0
[2022-05-14 20:59:15.164][Trace][221869][7173158y] TCP: disposing #0 resource(HttpConn)(0x557db80fab20), conns=3, disposing=1, zombies=0
[2022-05-14 20:59:15.420][Trace][221869][7173158y] RTC: session address init RTC拉流端公网IP地址:49305
[2022-05-14 20:59:15.420][Trace][221869][7173158y] RTC: session STUN done, waiting DTLS handshake.
[2022-05-14 20:59:15.421][Trace][221869][42u1778l] <- RTC RECV #12, udp 937, pps 8/14, schedule 937
[2022-05-14 20:59:15.434][Trace][221869][7173158y] DTLS: State Passive RECV, done=0, arq=0/0, r0=1, r1=0, len=157, cnt=22, size=144, hs=1
[2022-05-14 20:59:15.435][Trace][221869][7173158y] DTLS: State Passive SEND, done=0, arq=0/0, r0=-1, r1=2, len=681, cnt=22, size=82, hs=2
[...]

PR #3079, opened 2022-06-23
WebRTC: Fix no audio and video issue for Firefox.

johzzy:
#3041 

After the RTC starts streaming, the video payload type corresponding to the RTC source will change due to the SDP negotiation result.

If rtmp to rtc or srt to rtc occurs afterwards, the bridge constructs the video payload type of the rtp packet as the default value 102. This mismatch with the previous situation causes the webrtc receiver to not correctly obtain the video track.

Solution: Set the value of the video payload type for constructing the rtp packet in the bridge to match the video payload type of the rtc source.

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -711,17 +711,6 @@ srs_error_t SrsRtcSource::on_timer(srs_utime_t interval)
 
 #ifdef SRS_FFMPEG_FIT
 
-inline uint8_t srs_parse_video_payload_type_from(const std::vector<SrsRtcTrackDescription*>& descs) {
-    if (descs.empty()) {
-        return kVideoPayloadType;
-    }
-
-    SrsCodecPayload* media_ = descs.front()->media_;
-    // For publisher, pt should be equal to pt_of_publisher.
-    srs_assert(media_->pt_ == media_->pt_of_publisher_);
-    return media_->pt_;
-}
-
 SrsRtcFromRtmpBridge::SrsRtcFromRtmpBridge(SrsRtcSource* source)
 {
     req = NULL;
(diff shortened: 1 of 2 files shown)


trunk/src/app/srs_app_threads.cpp
```diff
@@ -62,18 +62,12 @@ extern SrsStageManager* _srs_stages;
...
 extern SrsRtcBlackhole* _srs_blackhole;
-extern SrsResourceManager* _srs_rtc_manager;
-
 extern SrsResourceManager* _srs_rtc_manager;
 extern SrsDtlsCertificate* _srs_rtc_dtls_certificate;
 #endif
 
 #include <srs_protocol_kbps.hpp>
 
-extern SrsPps* _srs_pps_snack2;
```
lipeng19811218: This deletion seems to have no relation to this commit.

`TRANS_BY_GPT3`

winlinvip: Removed.

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -710,6 +710,17 @@ srs_error_t SrsRtcSource::on_timer(srs_utime_t interval)
...
 
 #ifdef SRS_FFMPEG_FIT
+
+inline uint8_t srs_parse_video_payload_type_from(const std::vector<SrsRtcTrackDescription*>& descs) {
+    if (descs.empty()) {
+        return kVideoPayloadType;
+    }
+
+    SrsCodecPayload* media_ = descs.front()->media_;
+    assert(media_->pt_ == media_->pt_of_publisher_);
```
winlinvip: If the SDP is not standardized, will it cause a crash?

`TRANS_BY_GPT3`

PR #3101, opened 2022-07-05
Bugfix srshls memleak

bluestn:
in function SrsHlsMuxer::update_config, the following code will cause writer whose type is IFileWriter* pointer be overrided,
so need release writer before allocate a new instance for it.

    if(hls_keys) {
        writer = new SrsEncFileWriter();
    } else {
        writer = new SrsFileWriter();
    }


peipeiguo:
why this pr has so many files changed?

PR #3105, opened 2022-07-10
Kickoff publisher when stream is idle, which means no players. v6.0.31, v5.0.144

industriousonesoft:
For some use scenario, the publisher is invited when player want to view the stream:

1. Publisher connect to system, but does not publish any stream to SRS yet.
2. Player connect to system and start to request the stream.
3. System notifies publisher to publish stream to SRS.
4. Player play the stream from SRS.

Please notice that `system` means your business system, not SRS.

This is what we called `on-demand-live-streaming`, so when the last player stop to view the stream, what happends?

1. System needs to notify publisher to stop publish.
2. Or, SRS disconnect the publisher when idle(the last player stops playing).

This PR is for the solution 2, so that the cleanup is very simple, your system does not need to notify publisher to stop publish, because SRS has already disconnected the publihser.

### DIFF
--- a/trunk/conf/full.conf
+++ b/trunk/conf/full.conf
@@ -1343,6 +1343,7 @@ vhost publish.srs.com {
         try_annexb_first on;
         # The timeout time waiting for new consumers in ms. When the publisher is idle 
         # more than this config, it will be kicked off.
+        # Overwrite by env SRS_VHOST_PUBLISH_KICKOFF_FOR_IDLE for all vhosts.
         # default: 0 (0 or negative means disabled)
         kickoff_for_idle 0;
     }
--- a/trunk/conf/full.conf
+++ b/trunk/conf/full.conf
@@ -1341,10 +1341,11 @@ vhost publish.srs.com {
         # Overwrite by env SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST for all vhosts.
         # default: on
         try_annexb_first on;
-        # The timeout time waiting for new consumers in ms. When the publisher is idle 
-        # more than this config, it will be kicked off.
+        # The timeout in seconds to disconnect publisher when idle, which means no players.
+        # Note that 0 means no timeout or this feature is disabled.
+        # Note that this feature conflicts with forward, because it disconnect the publisher stream.
         # Overwrite by env SRS_VHOST_PUBLISH_KICKOFF_FOR_IDLE for all vhosts.
-        # default: 0 (0 or negative means disabled)
+        # default: 0
         kickoff_for_idle 0;
     }
 }
--- a/trunk/conf/full.conf
+++ b/trunk/conf/full.conf
@@ -1341,6 +1341,10 @@ vhost publish.srs.com {
         # Overwrite by env SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST for all vhosts.
         # default: on
         try_annexb_first on;
+        # The timeout time waiting for new consumers in ms. When the publisher is idle 
+        # more than this config, it will be kicked off.
+        # default: 0 (0 or negative means disabled)
+        kickoff_for_idle 0;
     }
 }
 
--- a/trunk/conf/rtmp.kickoff.conf
+++ b/trunk/conf/rtmp.kickoff.conf
@@ -8,6 +8,6 @@ daemon              off;
 srs_log_tank        console;
 vhost __defaultVhost__ {
     publish {
-        kickoff_for_idle     60000; # default is 60s, 0 means disable
+        kickoff_for_idle 60000;
     }
 }
--- a/trunk/conf/rtmp.kickoff.conf
+++ b/trunk/conf/rtmp.kickoff.conf
@@ -1,0 +1,13 @@
+# the config for srs to delivery RTMP with kicking off publish as no one watching.
+# @see https://github.com/ossrs/srs/wiki/v1_CN_SampleRTMP
+# @see full.conf for detail config.
+
+listen              1935;
+max_connections     1000;
+daemon              off;
+srs_log_tank        console;
+vhost __defaultVhost__ {
+    publish {
+        kickoff_for_idle     60000; # default is 60s, 0 means disable
+    }
+}
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -2559,7 +2559,7 @@ srs_error_t SrsConfig::check_normal_config()
                 for (int j = 0; j < (int)conf->directives.size(); j++) {
                     string m = conf->at(j)->name;
                     if (m != "mr" && m != "mr_latency" && m != "firstpkt_timeout" && m != "normal_timeout"
-                        && m != "parse_sps" && m != "try_annexb_first") {
+                        && m != "parse_sps" && m != "try_annexb_first" && m != "kickoff_for_idle") {
                         return srs_error_new(ERROR_SYSTEM_CONFIG_INVALID, "illegal vhost.publish.%s of %s", m.c_str(), vhost->arg0().c_str());
                     }
                 }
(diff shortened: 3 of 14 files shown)


trunk/conf/rtmp.kickoff.conf
```diff
@@ -0,0 +1,14 @@
...
+# @see full.conf for detail config.
+
+listen              1935;
+max_connections     1000;
+daemon              off;
+srs_log_tank        console;
+vhost __defaultVhost__ {
+    publish {
+        kickoff_enabled     on;
+        kickoff_timeout     60000;
```
winlinvip: The configuration can be merged into one.

```nginx
kickoff_timeout 0; // disabled
```

```nginx
kickoff_timeout 60000; // enabled.
```

The configuration item, `kickoff_timeout`, has been changed to `kickoff_for_idle` to kick off due to being idle (no player).

Additionally, please make changes to `full.conf` and add this configuration item. Please explain its purpose as well. The default value is 0, which means this option is not enabled.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -894,6 +898,15 @@ srs_error_t SrsRtmpConn::do_publishing(SrsLiveSource* source, SrsPublishRecvThre
             return srs_error_wrap(err, "rtmp: thread quit");
         }
 
+#ifdef SRS_PERF_KICKOFF_AS_NO_ONE_WATCHING
+        // Check if the source is expired or not as no one watching.
+        if (publish_kickoff_enabled && source->expired_as_no_one_watching(publish_kickoff_timeout))
+        {
```
winlinvip: Code style issue, the braces are at the end, please check the rest of the code.

```cpp
if (publish_kickoff_enabled && source->expired_as_no_one_watching(publish_kickoff_timeout)) {
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -894,6 +898,15 @@ srs_error_t SrsRtmpConn::do_publishing(SrsLiveSource* source, SrsPublishRecvThre
...
+            srs_trace("kickoff %s as no one watching for more than %d seconds.", req->tcUrl.c_str(), publish_kickoff_timeout / SRS_UTIME_SECONDS);
+            return srs_error_wrap(err, "rtmp: kickoff pusher as no one watching.");
```
winlinvip: There is no need to log, as the logs will be printed at the upper level. The variable information of the logs should be placed in the error.

```cpp
return srs_error_wrap(err, "rtmp: kickoff pusher as no one watching, url=%s, timeout=%ds", 
  req->tcUrl.c_str(), srsu2si(publish_kickoff_timeout));
```

> Note: Use `srsu2si` or `srsu2msi` to convert to int.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -2005,6 +2008,20 @@ bool SrsLiveSource::expired()
     return false;
 }
 
+#ifdef SRS_PERF_KICKOFF_AS_NO_ONE_WATCHING
+bool SrsLiveSource::expired_as_no_one_watching(srs_utime_t timeout)
```
winlinvip: This function name can be changed to:

```cpp
bool SrsLiveSource::idle_for(srs_utime_t timeout)
```

Please remove `SRS_PERF_KICKOFF_AS_NO_ONE_WATCHING`, there is no need for macro definition.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -2648,7 +2671,10 @@ srs_error_t SrsLiveSource::create_consumer(SrsLiveConsumer*& consumer)
     
     consumer = new SrsLiveConsumer(this);
     consumers.push_back(consumer);
-    
+#ifdef SRS_PERF_KICKOFF_AS_NO_ONE_WATCHING
+    no_one_wating_at = 0;
```
winlinvip: The variable name `idle_at` expresses the time point when the Source starts becoming idle.

`TRANS_BY_GPT3`

trunk/conf/full.conf
```diff
@@ -1341,6 +1341,10 @@ vhost publish.srs.com {
         # Overwrite by env SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST for all vhosts.
         # default: on
         try_annexb_first on;
+        # The timeout time waiting for new consumers in ms. When the publisher is idle 
+        # more than this config, it will be kicked off.
+        # default: 0 (0 or negative means disabled)
+        kickoff_for_idle 0;
```
winlinvip, line 1349: The configuration item needs to support environment variable settings, refer to `try_annexb_first` and `SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST`. 

```
        # When parsing SPS/PPS, whether try ANNEXB first. If not, try IBMF first, then ANNEXB.
        # Overwrite by env SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST for all vhosts.
        # default: on
        try_annexb_first on;
```

Please search for `SRS_VHOST_PUBLISH_TRY_ANNEXB_FIRST` and add the corresponding unit test to confirm that the parsing of this configuration item is correct.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -5340,6 +5340,30 @@ srs_utime_t SrsConfig::get_publish_normal_timeout(string vhost)
     return (srs_utime_t)(::atoi(conf->arg0().c_str()) * SRS_UTIME_MILLISECONDS);
 }
 
+srs_utime_t SrsConfig::get_publish_kickoff_timeout(std::string vhost)
+{
+    // the timeout to kickoff publish as no one watching,
+    // the default value is 3 minutes.
+    static srs_utime_t DEFAULT = 180 * SRS_UTIME_SECONDS;
```
winlinvip: The default value is `0`, right? It says 0 in `full.conf`.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.cpp
```diff
@@ -5340,6 +5340,30 @@ srs_utime_t SrsConfig::get_publish_normal_timeout(string vhost)
     return (srs_utime_t)(::atoi(conf->arg0().c_str()) * SRS_UTIME_MILLISECONDS);
 }
 
+srs_utime_t SrsConfig::get_publish_kickoff_timeout(std::string vhost)
```
winlinvip: The function name should also be changed accordingly so that it can be searched for.

```
get_publish_kickoff_for_idle
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -985,6 +985,7 @@ srs_error_t SrsRtmpConn::do_publishing(SrsLiveSource* source, SrsPublishRecvThre
     // initialize the publish timeout.
     publish_1stpkt_timeout = _srs_config->get_publish_1stpkt_timeout(req->vhost);
     publish_normal_timeout = _srs_config->get_publish_normal_timeout(req->vhost);
+    srs_utime_t publish_kickoff_timeout = _srs_config->get_publish_kickoff_timeout(req->vhost);
```
winlinvip: This variable name also needs to be changed to `kickoff_for_idle` so that it can be globally searched.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -1008,6 +1009,11 @@ srs_error_t SrsRtmpConn::do_publishing(SrsLiveSource* source, SrsPublishRecvThre
             return srs_error_wrap(err, "rtmp: thread quit");
         }
 
+        // Check if the source is expired or idle (no consumers or players).
+        if (source->idle_for(publish_kickoff_timeout)) {
+            return srs_error_wrap(err, "rtmp: kickoff publisher as no one watching for a while, url=%s, timeout=%ds", req->tcUrl.c_str(), srsu2si(publish_kickoff_timeout));
```
chundonglinlin: The log message suggests to display `no players` or `no subscribers`.
```log
return srs_error_wrap(err, "rtmp: kickoff publisher for no players over timeout=%ds, stream_url=%s", srsu2si(publish_kickoff_timeout), req->get_stream_url().c_str())
```

`TRANS_BY_GPT3`

winlinvip (approved):

trunk/src/app/srs_app_source.cpp
```diff
@@ -2690,7 +2707,8 @@ srs_error_t SrsLiveSource::create_consumer(SrsLiveConsumer*& consumer)
     
     consumer = new SrsLiveConsumer(this);
     consumers.push_back(consumer);
-    
+    idle_at = 0;
```
chundonglinlin: If `forward` is configured, but it is not a `consumer`, should it also be considered?
```config
vhost __defaultVhost__ {
    forward 127.0.0.1:19350;
    publish {
        kickoff_for_idle     60000; # default is 60s, 0 means disable
    }
}
```

`TRANS_BY_GPT3`

chundonglinlin: image

winlinvip: During the configuration check, an error will occur if both forward and kickoff_for_idle are enabled at the same time.

`TRANS_BY_GPT3`

winlinvip:
This PR is part of #3420

chundonglinlin (approved):

PR #3107, opened 2022-07-11
WebRTC: Fix missing type in track desc when backup H.264 payload type is chosen.

faicker:
When the backup H.264 payload type is selected, the track desc type is not set.


winlinvip (approved):

winlinvip:
Thanks 👍

PR #3126, opened 2022-07-31
HLS: restore HLS information when republish stream.(#3088). v7.0.57

chundonglinlin:
### Feature
HLS continuous mode: In this mode HLS sequence number is started from where it stopped last time. Old fragments are kept. Default is on.
### Configuration
```
vhost __defaultVhost__ {
    hls {
        enabled         on;
        hls_fragment    10;
        hls_window      60;
        hls_continuous  on;
    }
}
```

Contributed by AI:

* [AI: Refine and extract HLS recover.](https://github.com/ossrs/srs/pull/3126/commits/656e4e296d8f93ff7e27df5c18e1f26cf5afcf82)

---------


trunk/src/app/srs_app_hls.cpp
```diff
@@ -345,6 +345,170 @@ srs_error_t SrsHlsMuxer::update_config(SrsRequest* r, string entry_prefix,
...
+    SrsFileReader fr;
+    if ((err = fr.open(m3u8)) != srs_success) {
+        return srs_error_wrap(err, "open file");
+    }
+
+    int nb_fbuf = fr.filesize();
+    char* fbuf = new char[nb_fbuf];
+    SrsAutoFreeA(char, fbuf);
+    if ((err = fr.read(fbuf, nb_fbuf, NULL)) != srs_success) {
+        return srs_error_wrap(err, "read data");
```
duiniuluantanqin, line 1277: If the read fails here, `fr.close()` will not be executed.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_hls.cpp
```diff
@@ -345,6 +345,170 @@ srs_error_t SrsHlsMuxer::update_config(SrsRequest* r, string entry_prefix,
...
+    }
+    fr.close();
+
+    std::string body = fbuf;
+
+    // parse
+    if (body.empty()) {
+        return srs_error_wrap(err, "read empty m3u8");
```
duiniuluantanqin, line 378: 这里也是，fr.close()没有被执行到

duiniuluantanqin (approved):

trunk/src/app/srs_app_hls.cpp
```diff
@@ -407,16 +407,13 @@ srs_error_t SrsHlsMuxer::restore_stream()
     if ((err = fr.read(fbuf, nb_fbuf, NULL)) != srs_success) {
         return srs_error_wrap(err, "read data");
     }
-    fr.close();
```
duiniuluantanqin: Did `fr.close()` delete it directly? Doesn't that mean `fr` is not released?

`TRANS_BY_GPT3`

winlinvip:
Due to the HLS state recovery, it is necessary to parse the m3u8 files on the disk, which also involves the recovery of various states of HLS. From a broader perspective, it is not suitable for SRS to do, which is very similar to DVR.

Of course, SRS has also implemented DVR FLV and MP4, because sometimes it is necessary to have various basic functions and capabilities for convenience. Unfortunately, it is not possible to implement a more complete capability in SRS, because SRS is a C++ server, and it is difficult to customize the business.

The best solution is to use Go with SRS in the SRS Stack, just like the recording we implemented in the SRS Stack, which includes recording to local disk, DVR to cloud storage, and even a VoD cloud on-demand system.

In summary: HLS status recovery can be implemented in SRS, but only basic recovery capabilities are implemented to keep it simple, and full coverage testing is required. Complete HLS recovery capabilities are implemented in the SRS Stack.

`TRANS_BY_GPT4`

trunk/conf/hls.conf
```diff
@@ -17,5 +17,6 @@ vhost __defaultVhost__ {
         hls_path        ./objs/nginx/html;
         hls_fragment    10;
         hls_window      60;
+        hls_continuous  on;
```
winlinvip: Change to: `hls_recover` would be better, corresponding to the panic-recover concept in Go.

`TRANS_BY_GPT4`

trunk/src/app/srs_app_hls.cpp
```diff
@@ -383,6 +383,167 @@ srs_error_t SrsHlsMuxer::update_config(SrsRequest* r, string entry_prefix,
     return err;
 }
 
+srs_error_t SrsHlsMuxer::restore_stream()
```
winlinvip: The logic of this exception recovery, because it is a very commonly used path, must be verified and ensured for correctness through utest, otherwise it is very easy to make mistakes.

From the perspective of utest, it would be best to split this function into two functions, making it easier to mock `fr`, that is, to read the m3u8 content from memory instead of from the actual file. It's somewhat like this:

```cpp
srs_error_t SrsHlsMuxer::restore_stream() {
    SrsFileReader fr;
    if ((err = fr.open(m3u8)) != srs_success) {
        return srs_error_wrap(err, "open file");
    }

    return do_restore_stream(&fr);
```

In this way, a mock can be used in utest to avoid reading the actual file.

```cpp
VOID TEST() {
    MOCKFileReader fr;
    SrsHlsMuxer muxer;
    muxer.do_restore_stream(&fr);
````

The above is just an example, you can refer to the Config section, which also requires mocking file reading.


`TRANS_BY_GPT4`

File: trunk/src/app/srs_app_hls.cpp [hunk 6]
```diff
@@ -383,6 +383,167 @@ srs_error_t SrsHlsMuxer::update_config(SrsRequest* r, string entry_prefix,
...
+    // read m3u8
+    SrsFileReader fr;
+    if ((err = fr.open(m3u8)) != srs_success) {
+        return srs_error_wrap(err, "open file");
+    }
+
+    int nb_fbuf = fr.filesize();
+    char* fbuf = new char[nb_fbuf];
+    SrsAutoFreeA(char, fbuf);
+    if ((err = fr.read(fbuf, nb_fbuf, NULL)) != srs_success) {
```
winlinvip: I seem to remember there is an auxiliary function like `srs read all`, which can read all the content of a certain Reader until EOF.

It's a bit like the `ioutil.ReadAll(io.Reader)` function in Go.

`TRANS_BY_GPT4`

winlinvip (approved):

PR #3159, opened 2022-08-31
Fix the recursively included issue for srs_core_time.hpp

mingo-wu1:

1. In included file: main file cannot be included recursively when building a preamble
    clang(pp_including_mainfile_in_preamble)
    srs_core.hpp(43, 10): Error occurred here
    Circular reference to header files

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/src/core/srs_core_time.hpp
+++ b/trunk/src/core/srs_core_time.hpp
@@ -7,8 +7,15 @@
 #ifndef SRS_CORE_TIME_HPP
 #define SRS_CORE_TIME_HPP
 
+#if defined(_WIN32) && !defined(__MINGW32__) && (!defined(_MSC_VER) || _MSC_VER<1600) && !defined(__WINE__)
+#include <BaseTsd.h>
+typedef __int64 int64_t;
+#else
+#include <stdint.h>
+#endif
+
 // Time and duration unit, in us.
-typedef signed long int srs_utime_t;
+typedef int64_t srs_utime_t;
 
 // The time unit in ms, for example 100 * SRS_UTIME_MILLISECONDS means 100ms.
 #define SRS_UTIME_MILLISECONDS 1000
--- a/trunk/src/core/srs_core_time.hpp
+++ b/trunk/src/core/srs_core_time.hpp
@@ -7,10 +7,8 @@
 #ifndef SRS_CORE_TIME_HPP
 #define SRS_CORE_TIME_HPP
 
-#include <srs_core.hpp>
-
 // Time and duration unit, in us.
-typedef int64_t srs_utime_t;
+typedef signed long int srs_utime_t;
 
 // The time unit in ms, for example 100 * SRS_UTIME_MILLISECONDS means 100ms.
 #define SRS_UTIME_MILLISECONDS 1000


trunk/src/core/srs_core_time.hpp
```diff
@@ -7,10 +7,8 @@
 #ifndef SRS_CORE_TIME_HPP
 #define SRS_CORE_TIME_HPP
 
-#include <srs_core.hpp>
-
 // Time and duration unit, in us.
-typedef int64_t srs_utime_t;
+typedef signed long int srs_utime_t;
```
winlinvip: Why change this? Not causes the recursively including.

mingo-wu1: Here is a detailed analysis to indicate that I have a lack of thinking (please see in detail, I will make more rigorous changes):

1. int64_t is defined on my machine Ubuntu 20.04 Focal at /usr/include/x86_64-linux-gnu/bits/stdint-intn.h, line 27, with the code typedef __int64_t int64_t; If int64_t is needed, the header file <bits/stdint-intn.h> or a higher-level header file needs to be included.

2. When removing the inclusion of the header file <srs_core.hpp> due to header file circular referencing, the trunk/src/core/srs_core_time.hpp does not include any header files. At this point, int64_t will throw an error indicating that the definition cannot be found (this error will occur with both gcc and g++).

3. At this point, there are two solutions: <1> include the header files included in the header file or <2> modify int64_t to signed long int (on my machine, int64_t is defined by signed long int, but there is a lack of consideration here).

4. After my more rigorous analysis, I believe that replacing typedef int64_t srs_utime_t with typedef signed long int srs_utime_t is inadequate because I only considered my own machine.

5. Because the rule exists that short <= int <= long <= long long, and in some cases, the size of long may not be 64 bits, such as on a 32-bit machine where long is 32 bits.

6. I checked the IDEs of Visual Studio on Windows and VSCode on Ubuntu (the versions are not relevant here), and int64_t is included in <stdint.h>.

7. Therefore, I referred to the usage of the header file <stdint.h> in the srs project in trunk/src/app/srs_app_rtc_sdp.hpp, trunk/src/app/srs_app_uuid_cpp, and trunk/src/protocol/srs_protocol_http_stack.hpp.
<1>  trunk/src/app/srs_app_rtc_sdp.hpp
#include <stdint.h>
<2>   trunk/src/app/srs_app_uuid_cpp
#ifdef HAVE_STDINT_H
#include <stdint.h>
#else
#ifdef HAVE_INTTYPES_H
#include <inttypes.h>
#endif
#endif
<3>  trunk/src/protocol/srs_protocol_http_stack.hpp
#if defined(_WIN32) && !defined(__MINGW32__) && \
  (!defined(_MSC_VER) || _MSC_VER<1600) && !defined(__WINE__)
#include <BaseTsd.h>
typedef __int8 int8_t;
typedef unsigned __int8 uint8_t;
typedef __int16 int16_t;
typedef unsigned __int16 uint16_t;
typedef __int32 int32_t;
typedef unsigned __int32 uint32_t;
typedef __int64 int64_t;
typedef unsigned __int64 uint64_t;
#else
#include <stdint.h>
#endif

It can be inferred that some environments do not include <stdint.h>.

8. So I used gcc and g++ to create simple test files to test the above three referencing strategies: gcc -o test test.cpp and g++ -o test test.cpp (the second one will throw an error: unterminated). Therefore, I believe a more rigorous approach is to consider the system environment more comprehensively.
#if defined(_WIN32) && !defined(__MINGW32__) && \
  (!defined(_MSC_VER) || _MSC_VER<1600) && !defined(__WINE__)
#include <BaseTsd.h>
typedef __int64 int64_t;
#else
#include <stdint.h>
#endif

9. Summary:
So far, we have considered the operating system's bitness, different header file dependencies of the operating system, and the IDE. After conducting compilation tests, the inclusion of int64_t should be done using the following approach: adopting the strategy of trunk/src/protocol/srs_protocol_http_stack.hpp and making modifications to it.
#if defined(_WIN32) && !defined(__MINGW32__) && (!defined(_MSC_VER) || _MSC_VER<1600) && !defined(__WINE__)
#include <BaseTsd.h>
typedef __int64 int64_t;
#else
#include <stdint.h>
#endif

If you have any doubts, please raise them. It is possible that any oversight on my part may have caused them.

`TRANS_BY_GPT3`

winlinvip: Good job! Thanks.

winlinvip:
I make a small change, directly define the srs_utine_t, do not define int64_t, because it might cause redefine issue.

```
// Time and duration unit, in us.
#if defined(_WIN32) && !defined(__MINGW32__) && (!defined(_MSC_VER) || _MSC_VER<1600) && !defined(__WINE__)
#include <BaseTsd.h>
typedef __int64 srs_utime_t;
#else
#include <stdint.h>
typedef int64_t srs_utime_t;
#endif
```

PR #3232, opened 2022-11-09
Limit max frames in gop cache by gop_cache_max_frames

bluestn:
set limits for gop cahe max frames, to keep srs from run out of memory while stream pusher pushes no video key frames for a long time.

### DIFF
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -37,7 +37,7 @@ vhost __defaultVhost__ {
         rtc_to_rtmp off;
     }
 
-	play{
-		gop_cache_max_frames 100;
+    play{
+        gop_cache_max_frames 100;
 	}
 }
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -36,4 +36,8 @@ vhost __defaultVhost__ {
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtc-to-rtmp
         rtc_to_rtmp off;
     }
+
+	play{
+		gop_cache_max_frames 100;
+	}
 }
--- a/trunk/src/app/srs_app_source.hpp
+++ b/trunk/src/app/srs_app_source.hpp
@@ -231,7 +231,7 @@ private:
     // to limit the max gop cache frames
     // without this limit, if ingest stream always has no IDR frame
     // it will cause srs run out of memory
-    int gop_cache_max_frames;
+    int gop_cache_max_frames_;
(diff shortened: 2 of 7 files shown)


trunk/src/app/srs_app_source.hpp
```diff
@@ -251,6 +255,8 @@ class SrsGopCache
     virtual void dispose();
     // To enable or disable the gop cache.
     virtual void set(bool v);
+    // max gop frames
+    virtual void set_max_frames(int m);
```
winlinvip: This name should be consistent so that it can be found during searches. It should be changed to:
```cpp
void set_gop_cache_max_frames(int v);
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.hpp
```diff
@@ -589,6 +595,7 @@ class SrsLiveSource : public ISrsReloadHandler
     virtual srs_error_t consumer_dumps(SrsLiveConsumer* consumer, bool ds = true, bool dm = true, bool dg = true);
     virtual void on_consumer_destroy(SrsLiveConsumer* consumer);
     virtual void set_cache(bool enabled);
+    virtual void set_cache_max_frames(int max);
```
winlinvip: This name should be consistent so that it can be found during searches. It should be changed to:
```cpp
void set_gop_cache_max_frames(int v);
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.hpp
```diff
@@ -228,6 +228,10 @@ class SrsGopCache
     // The client will wait for the next keyframe for h264,
     // and will be black-screen.
     bool enable_gop_cache;
+    // to limit the max gop cache frames
+    // without this limit, if ingest stream always has no IDR frame
+    // it will cause srs run out of memory
+    int gop_cache_max_frames;
```
winlinvip: According to the new specifications, change it to:
```cpp
int gop_cache_max_frames_;
```

`TRANS_BY_GPT3`

trunk/conf/srs.conf
```diff
@@ -36,4 +36,8 @@ vhost __defaultVhost__ {
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtc-to-rtmp
         rtc_to_rtmp off;
     }
+
+	play{
```
winlinvip: Do not use `\t`, use spaces to avoid formatting issues.

`TRANS_BY_GPT3`

PR #3236, opened 2022-11-10
Live: Limit cached max frames by gop_cache_max_frames

bluestn:
Use gop_cache_max_frames to limit the max frames which can be cached by gop cache, to keep srs from run out of memory while some malicious stream pusher never send any Key frames.

### DIFF
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -5,7 +5,7 @@ listen              1935;
 max_connections     1000;
 #srs_log_tank        file;
 #srs_log_file        ./objs/srs.log;
-daemon              on;
+daemon              off;
 http_api {
     enabled         on;
     listen          1985;
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -37,7 +37,7 @@ vhost __defaultVhost__ {
         rtc_to_rtmp off;
     }
 
-	play{
-		gop_cache_max_frames 100;
+    play{
+        gop_cache_max_frames 100;
 	}
 }
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -38,6 +38,6 @@ vhost __defaultVhost__ {
     }
 
     play{
-        gop_cache_max_frames 100;
+        gop_cache_max_frames 250;
     }
 }
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -39,5 +39,5 @@ vhost __defaultVhost__ {
 
     play{
         gop_cache_max_frames 100;
-	}
+    }
 }
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -36,4 +36,8 @@ vhost __defaultVhost__ {
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtc-to-rtmp
         rtc_to_rtmp off;
     }
+
+	play{
+		gop_cache_max_frames 100;
+	}
 }
--- a/trunk/conf/srs.conf
+++ b/trunk/conf/srs.conf
@@ -5,7 +5,7 @@ listen              1935;
 max_connections     1000;
 #srs_log_tank        file;
 #srs_log_file        ./objs/srs.log;
-daemon              off;
+daemon              on;
 http_api {
     enabled         on;
     listen          1985;
--- a/trunk/src/app/srs_app_config.hpp
+++ b/trunk/src/app/srs_app_config.hpp
@@ -553,9 +553,8 @@ public:
     // @return true when gop_cache is ok; otherwise, false.
     // @remark, default true.
     virtual bool get_gop_cache(std::string vhost);
-    // to set the max frames limit for gop cache
+    // to set the max frames limit in gop cache
     // @return gop cache frames limit
-    // @remark, default 250 (10s for 25fps)
     virtual int get_gop_cache_max_frames(std::string vhost);
     // Whether debug_srs_upnode is enabled of vhost.
     // debug_srs_upnode is very important feature for tracable log,
--- a/trunk/src/app/srs_app_config.hpp
+++ b/trunk/src/app/srs_app_config.hpp
@@ -553,8 +553,7 @@ public:
     // @return true when gop_cache is ok; otherwise, false.
     // @remark, default true.
     virtual bool get_gop_cache(std::string vhost);
-    // to set the max frames limit in gop cache
-    // @return gop cache frames limit
+    // Get the limit max frames for gop cache.
     virtual int get_gop_cache_max_frames(std::string vhost);
     // Whether debug_srs_upnode is enabled of vhost.
     // debug_srs_upnode is very important feature for tracable log,
--- a/trunk/src/app/srs_app_config.hpp
+++ b/trunk/src/app/srs_app_config.hpp
@@ -553,6 +553,10 @@ public:
     // @return true when gop_cache is ok; otherwise, false.
     // @remark, default true.
     virtual bool get_gop_cache(std::string vhost);
+    // to set the max frames limit for gop cache
+    // @return gop cache frames limit
+    // @remark, default 250 (10s for 25fps)
+    virtual int get_gop_cache_max_frames(std::string vhost);
     // Whether debug_srs_upnode is enabled of vhost.
     // debug_srs_upnode is very important feature for tracable log,
     // but some server, for instance, flussonic donot support it.
--- a/trunk/src/app/srs_app_rtmp_conn.cpp
+++ b/trunk/src/app/srs_app_rtmp_conn.cpp
@@ -557,10 +557,12 @@ srs_error_t SrsRtmpConn::stream_service_cycle()
     srs_assert(source != NULL);
 
     bool enabled_cache = _srs_config->get_gop_cache(req->vhost);
-    srs_trace("source url=%s, ip=%s, cache=%d, is_edge=%d, source_id=%s/%s",
-        req->get_stream_url().c_str(), ip.c_str(), enabled_cache, info->edge, source->source_id().c_str(), source->pre_source_id().c_str());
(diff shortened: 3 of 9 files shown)


trunk/conf/srs.conf
```diff
@@ -36,4 +36,8 @@ vhost __defaultVhost__ {
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtc-to-rtmp
         rtc_to_rtmp off;
     }
+
+    play{
+        gop_cache_max_frames 100;
+	}
```
winlinvip: There are also issues with these two lines of indentation, please take a look at the page.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.hpp
```diff
@@ -251,6 +255,8 @@ class SrsGopCache
     virtual void dispose();
     // To enable or disable the gop cache.
     virtual void set(bool v);
+    // max gop frames
+    virtual void set_max_frames(int v);
```
winlinvip: Keep the functions and variables consistent for easy searching. Change to:

```cpp
void set_gop_cache_max_frames(int v);
```

In addition, the comments for the functions are invalid, please remove them.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.hpp
```diff
@@ -589,6 +595,7 @@ class SrsLiveSource : public ISrsReloadHandler
     virtual srs_error_t consumer_dumps(SrsLiveConsumer* consumer, bool ds = true, bool dm = true, bool dg = true);
     virtual void on_consumer_destroy(SrsLiveConsumer* consumer);
     virtual void set_cache(bool enabled);
+    virtual void set_cache_max_frames(int v);
```
winlinvip: Keep the functions and variables consistent for easy searching. Change to:

```cpp
void set_gop_cache_max_frames(int v);
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -641,6 +653,11 @@ srs_error_t SrsGopCache::cache(SrsSharedPtrMessage* shared_msg)
         return err;
     }
     
+    if (cached_video_count > gop_cache_max_frames_){
+        srs_warn("too much frames in the gop cache");
```
winlinvip: This `gop_cache_max_frames_` is configured, so `cached_video_count` may be a variable and should be added to the print statement.

```cpp
srs_warn("too much frames in the gop cache, frames=%d, videos=%d, audio=%d", 
    (int)gop_cache.size(), cached_video_count, audio_after_last_video_count);
```

> Note: Additionally, upon further inspection, there is also the number of audio after the last video, `audio_after_last_video_count`, which can also be printed.

To avoid the situation where we don't know how many frames have been lost when problems occur.

`TRANS_BY_GPT3`

winlinvip:
最后，麻烦给出验证方式，能触发这个机制，确保没有隐藏问题。

trunk/src/app/srs_app_source.cpp
```diff
@@ -597,6 +598,17 @@ void SrsGopCache::set(bool v)
...
+
+void SrsGopCache::set_max_frames(int v)
+{
+    gop_cache_max_frames_ = v;
+
+    if (cached_video_count > gop_cache_max_frames_){
+        srs_warn("too much frames in the gop cache");
+        clear();
+    }
+}
```
winlinvip, line 604: This function should not call clear because it is a set function, and it is generally expected to only set variables without dropping frames.

After setting, the next incoming frame will trigger the frame dropping logic, but it is not necessary to do it here.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -641,6 +653,11 @@ srs_error_t SrsGopCache::cache(SrsSharedPtrMessage* shared_msg)
         return err;
     }
     
+    if (cached_video_count > gop_cache_max_frames_){
+        srs_warn("too much frames in the gop cache");
+        clear();
+    }
+
     // clear gop cache when got key frame
     if (msg->is_video() && SrsFlvVideo::keyframe(msg->payload, msg->size)) {
```
winlinvip: It would be better to drop frames after this.

```cpp
if (msg->is_video() && SrsFlvVideo::keyframe(msg->payload, msg->size)) {}

if (cached_video_count > gop_cache_max_frames_){}
```

Priority should be given to keyframes, followed by a fallback strategy for video frames.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -641,6 +653,11 @@ srs_error_t SrsGopCache::cache(SrsSharedPtrMessage* shared_msg)
         return err;
     }
     
+    if (cached_video_count > gop_cache_max_frames_){
```
winlinvip: The length of the queue `gop_cache.size()` should be used instead of the variable `cached_video_count`, as it only represents video frames.

This situation usually occurs as an exception, and continuously sending data messages can also cause OOM (Out of Memory).

`TRANS_BY_GPT3`

trunk/src/app/srs_app_config.hpp
```diff
@@ -553,6 +553,10 @@ class SrsConfig
     // @return true when gop_cache is ok; otherwise, false.
     // @remark, default true.
     virtual bool get_gop_cache(std::string vhost);
+    // to set the max frames limit for gop cache
+    // @return gop cache frames limit
+    // @remark, default 250 (10s for 25fps)
```
winlinvip: Do not write the magic number 250 here, as it has already been written in `full.conf`. Having constants in two places can lead to forgetting to update one when modifying the other.

If it is necessary to write, it can be written in `full.conf`. Configuration files and documentation are as effective as code.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -560,6 +560,7 @@ srs_error_t SrsRtmpConn::stream_service_cycle()
     srs_trace("source url=%s, ip=%s, cache=%d, is_edge=%d, source_id=%s/%s",
         req->get_stream_url().c_str(), ip.c_str(), enabled_cache, info->edge, source->source_id().c_str(), source->pre_source_id().c_str());
     source->set_cache(enabled_cache);
+    source->set_cache_max_frames(_srs_config->get_gop_cache_max_frames(req->vhost));
```
winlinvip: There is a print log in front, and this configuration should be printed out. If it is configured as 1, it will cause a problem here.

```cpp
int gcf = _srs_config->get_gop_cache_max_frames(req->vhost);
srs_trace("source url=%s, ip=%s, cache=%d/%d, is_edge=%d, source_id=%s/%s",
        req->get_stream_url().c_str(), ip.c_str(), enabled_cache, gcf, info->edge, source->source_id().c_str(), source->pre_source_id().c_str());

source->set_cache_max_frames(gcf);
```

> Note: It can be printed after `cache`.

`TRANS_BY_GPT3`

bluestn:
OK, modified as you pointed out. thank u.

How to verify this patch can work?

1. find a flv video which is long enough encoded with avc/aac.
2. use the flv key frame remove tool (https://github.com/bluestn/flv_remove_keyframe) to remove the flv key frames.
```bash
git clone https://github.com/bluestn/flv_remove_keyframe.git
mkdir -p flv_remove_keyframe/build
cd flv_remove_keyframe/build
cmake ..
make
cp ~/git/srs/trunk/doc/source.flv .
./flv_remove_keyframe source.flv
```
3. use command to push the flv file to srs server.
```
ffmpeg -re -i source_no_keyframe.flv -c copy -f flv rtmp://localhost/live/livestream 
```
4. without this patch, and when gop_cache is on, we can easily see that srs server will cache more and more frames in the gop_cache, and finally oom sooner or later.
    and after this patch applied, gop_cache will be kept no more than the frames set by gop_cache_max_frames.

winlinvip:
Verified, it works! 👍 

Before this patch, the gop cache increases:
`dispatch cached gop success. count=4244, duration=62760`

After this patch, we will clear the gop cache:
`Gop cache exceed max frames=250, total=251, videos=93, aalvc=1`

And the memory is ok now.

trunk/src/app/srs_app_source.cpp
```diff
@@ -656,16 +655,13 @@ srs_error_t SrsGopCache::cache(SrsSharedPtrMessage* shared_msg)
...
     }
 
-
     // cache the frame.
     gop_cache.push_back(msg->copy());
 
+    // Clear gop cache if exceed the max frames.
     if (gop_cache.size() > (size_t)gop_cache_max_frames_) {
-        srs_warn("too many frames in the gop cache,  %d video frames & %d audio frames dropped, "
-                 "audio_after_last_video_count: %d.",
```
winlinvip: Do not break the string in general logs, as it may cause search issues. This is a coding convention in JavaScript, and I personally think it makes a lot of sense.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -598,7 +598,6 @@ void SrsGopCache::set(bool v)
     }
 }
 
-
```
winlinvip: Do not have extra blank lines, as blank lines have meaning. Refer to the Go coding guidelines.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_source.cpp
```diff
@@ -575,7 +575,7 @@ SrsGopCache::SrsGopCache()
     cached_video_count = 0;
     enable_gop_cache = true;
     audio_after_last_video_count = 0;
-    gop_cache_max_frames_ = 250;
+    gop_cache_max_frames_ = 0;
```
winlinvip, line 578: Due to the default configuration of 250 as a magic number, it should not be used again here as it would cause interference.

`TRANS_BY_GPT3`

Issue #3249, opened 2022-11-19
SRT: Config file does not enable srt for srt2rtc.conf


dev-clavis:
> Note: Please read FAQ before file an issue, see 2716

**Description**

> I'd like to use srt to serve low latency content therefore I followed your guide (https://ossrs.io/lts/en-us/docs/v5/doc/sample-srt) which uses the develop of the repo and crashs because of memory leaks (which I think is already known). Because of that I used the commit of tag v5.0.63 (currently the newest) and followed again your guide _(with ./configure --srt=on)_, which all works fine then I start the server with the srt2rtc.conf file and try to connect from obs (other pc) like I did in srs v4 with ''srt://<ip>:10080?streamid=#!::r=live/livestream,m=publish'' but then obs immediately  stops with an error msg (see img) while the log of the server mentions that srt is disabled (which shouldn't be the case)
![image](https://user-images.githubusercontent.com/17059861/202861026-ecb0bfe8-6e81-4876-9c8f-315b34406920.png)

1. SRS Version: `v5.0.63`

1. SRS Log:

```
[2022-11-19 16:20:08.524][Trace][81042][1v4d1c72] XCORE-SRS/5.0.63(Bee)
[2022-11-19 16:20:08.524][Trace][81042][1v4d1c72] config parse complete
[2022-11-19 16:20:08.524][Trace][81042][1v4d1c72] write log to console
[2022-11-19 16:20:08.524][Trace][81042][1v4d1c72][MAIN] SRS/5.0.63(Bee), MIT
[2022-11-19 16:20:08.524][Trace][81042][1v4d1c72] authors: Winlin<winlin@vip.126.com> ZhaoWenjie<zhaowenjie@tal.com> ShiWei<shiwei05@kuaishou.com> XiaoZhihong<hondaxiao@tencent.com> WuPengqiang<pengqiang.wpq@alibaba-inc.com> XiaLixin<xialixin@kanzhun.com> LiPeng<mozhan.lp@alibaba-inc.com> ChenGuanghua<jinxue.cgh@alibaba-inc.com> ChenHaibo<nmgchenhaibo@foxmail.com> Johzzy<hellojinqiang@gmail.com> Zhouxiaojun2008<279686030@qq.com> Pengfei.ma<pengfei.ma.chn@outlook.com> ZhangJunqin<zhangjunqin@jd.com> and https://github.com/ossrs/srs/blob/develop/trunk/AUTHORS.md#contributors
[...]

Issue #3104, opened 2022-07-08
srs v5: srt serve error code=6006 : srt disabled (even when config says enabled)


ufou:
> Note: Please read FAQ before file an issue, see 2716

> Note: 提问前，请先看FAQ, 即 2716

**Description(描述)**

> I expect SRS v5 to allow SRT connections when config has srt_server enabled, but I get this srt disabled error?

1. SRS Version(版本): `5.0.33`

1. SRS Log(日志):

```
[2022-07-08 13:48:20.430][Trace][17][0883s53l] SRT client ip=10.0.10.10:56389, fd=18649528
[2022-07-08 13:48:20.430][Trace][17][0883s53l] SRT: before dispose resource(srt-ts-conn)(0x7fcd7c03f890), conns=1, zombies=0, ign=0, inz=0, ind=0
[2022-07-08 13:48:20.430][Error][17][0883s53l][104] srt serve error code=6006 : srt disabled, vhost=
thread [17][0883s53l]: do_cycle() [src/app/srs_app_srt_conn.cpp:260][errno=104](Connection reset by peer)
[2022-07-08 13:48:20.430][Trace][17][4k32hzj8] SRT: clear zombies=1 resources, conns=1, removing=0, unsubs=0
[2022-07-08 13:48:20.430][Trace][17][0883s53l] SRT: disposing #0 resource(srt-ts-conn)(0x7fcd7c03f890), conns=1, disposing=1, zombies=0
```

1. SRS Config(配置):

```
listen              1935;
max_connections     10000;
srs_log_tank        console;
srs_log_level       trace;
daemon              off;
http_api {
    enabled         on;
    listen          1985;
}
http_server {
    enabled         on;
    listen          1995;
    dir             ./objs/nginx/html;
}
srt_server {
    enabled         on;
    listen          10080;
    maxbw           1000000000;
    connect_timeout 4000;
    peerlatency     300;
    recvlatency     300;
}
vhost __defaultVhost__ {
    http_hooks {
        enabled     on;
        on_connect http://localhost:8089/srs/on_connect/;
        on_publish http://localhost:8089/srs/on_publish/;
        on_unpublish http://localhost:8089/srs/on_unpublish/;
        on_play http://localhost:8089/srs/on_play/;
    }
    publish {
        # 1st packet timeout in ms for encoder.
        # default: 20000
        firstpkt_timeout    30000;
        # normal packet timeout in ms for encoder.
        # default: 5000
        normal_timeout      15000;
    }
[...]

PR #3250, opened 2022-11-19
SRT: Config file does not enable srt for srt2rtc.conf

dev-clavis:
Changed the default config srt2rtc.conf because otherwise (in v5)  srs throws ''srt serve error code 6006''
see #3249 and #3104

### DIFF
--- a/trunk/conf/srt2rtc.conf
+++ b/trunk/conf/srt2rtc.conf
@@ -44,9 +44,9 @@ vhost __defaultVhost__ {
     rtc {
         enabled     on;
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtmp-to-rtc
-        rtmp_to_rtc off;
+        rtmp_to_rtc on;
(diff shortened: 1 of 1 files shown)


trunk/conf/srt2rtc.conf
```diff
@@ -38,12 +38,15 @@ rtc_server {
...
 # @doc https://github.com/ossrs/srs/issues/1147#issuecomment-577607026
 vhost __defaultVhost__ {
+    srt {
+        enabled     on;
+    }
     rtc {
         enabled     on;
         # @see https://ossrs.net/lts/zh-cn/docs/v4/doc/webrtc#rtmp-to-rtc
-        rtmp_to_rtc off;
+        rtmp_to_rtc on;
```
winlinvip, line 47: Please note that when coverting SRT to WebRTC, SRS actually converts SRT to RTMP, then converts RTMP to WebRTC, so we must enable `rtmp_to_rtc` here.

winlinvip (approved):

PR #3258, opened 2022-11-20
Feature/set asan report path

chen-guanghua:
Asan: support set asan report log path by conf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@
 
 .idea
 .DS_Store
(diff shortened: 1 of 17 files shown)


winlinvip:
It is not recommended to use separate logs for asan, instead it should be directly written into SRS logs. You can refer to the practices of FFmpeg API and SRT logs.

This way, users do not need to separately search for log files.

`TRANS_BY_GPT3`

PR #3264, opened 2022-11-21
Asan: Try to fix st_memory_leak for asan check

chen-guanghua:
Asan: fix st_memory_leak for asan check


winlinvip:
See https://github.com/ossrs/state-threads/commit/843e74b038b53bde526dbf0ef35d008d1f3ff85b

winlinvip:
See 6fa17aa3f

PR #3282, opened 2022-11-25
API: Bugfix: Clean stat when hooks fail

### LINKED ISSUE #3231: API: Failed to kickoff when authentication by on_play error. Some clients cannot be kicked off when on_play authentication fails.
> Note: Please read FAQ before file an issue, see 2716

> Note: Before asking a question, please refer to the FAQ, specifically 2716.

**Description**
> Please describe your issue here.
I am using SRS5 deployed on Tencent Cloud TKE. I have set up a public CLB for port mapping. When playing HLS streams, it triggers the on_play hook for authentication. If the authentication is successful, the clients work normally and the playback is smooth. However, if the authentication fails, a bunch of strange clients appear (4 additional clients for each failure), and kicking the stream results in a 400 error.

In the screenshot below, I manually stopped the stream, and it shows "No stream" with 4 online users. This display persists and I am unable to kick these inexplicable clients.
image

image

This issue can cause errors in the code's judgment. Originally, the stream would automatically disconnect when no one was watching, based on the clients' judgment. But now, even if there is an error with the clients, the stream remains disconnected.

1. SRS Version: `xxxxxx`
develog 
commit 21b9345387a66106dbc9e0afdf5a2911baf4ea0d 

1. SRS Log:

image

1. SRS Config:

```
# conf/http_hls.conf
listen              1935;
max_connections     10000
srs_log_level       trace;
# Log rotation needs to install the logrotate plugin: [logrotate](https://ossrs.net/lts/zh-cn/docs/v5/doc/log-rotate)
pid                  /data/logs/srslog/srs.pid;
ff_log_dir           /data/logs/srslog;
srs_log_file         /data/logs/srslog/srs.log;

http_api {
    enabled         on;
    listen          8008;
    crossdomain     on;
}

heartbeat {
    enabled         on;
    interval        9.3;
    url             https://*****/invoke/heartbeat;
    device_id       "srs1";
    summaries       off;
}

rtc_server {
    enabled on;
    listen 8000;
    tcp {
        enabled on;
        listen 8000;
    }
    protocol tcp;
    candidate *;
}

http_server {
    enabled         on;
    listen          8008;
    dir             /run/media;
}

vhost __defaultVhost__ {
    hls {
        enabled         on;
        hls_path        /run/media;
        hls_fragment    4;
        hls_window      10;
    }

    rtc {
        enabled on;
        nack on;
        twcc on;
        stun_timeout 30;
        dtls_role passive;
        rtmp_to_rtc on;
        rtc_to_rtmp off;
    }

    http_remux {
        enabled     on;
    }

    http_hooks {
	enabled        on;
	on_hls         https://***/invoke/onHls;
        on_play        https://***/invoke/onPlay;
        on_unpublish   https://***/invoke/onUnPublish;
        on_publish     https://***/invoke/onPublish;
        on_stop        https://***invoke/onStop;
   }
}

```

**Replay**
> Please describe how to replay the bug? (Steps to reproduce the bug)

Translate to English:
'Play using HLS, if the on_play returns a code other than 0, the bug can be reproduced.

Translate to English:
'**Expect (Expected Behavior)**

Translate to English:
'> Please describe your expectation (Describe what you expect to happen)
I hope that when on_play returns a code other than 0, clients can count normally.

`TRANS_BY_GPT3`

duiniuluantanqin:
fix issue https://github.com/ossrs/srs/issues/3231

### DIFF
--- a/trunk/src/app/srs_app_http_static.cpp
+++ b/trunk/src/app/srs_app_http_static.cpp
@@ -150,6 +150,7 @@ srs_error_t SrsHlsStream::serve_new_session(ISrsHttpResponseWriter* w, ISrsHttpM
 
     // We must do hook after stat, because depends on it.
     if ((err = http_hooks_on_play(req)) != srs_success) {
+        stat->on_disconnect(ctx);
         return srs_error_wrap(err, "HLS: http_hooks_on_play");
     }
 


trunk/src/app/srs_app_http_static.cpp
```diff
@@ -150,6 +150,7 @@ srs_error_t SrsHlsStream::serve_new_session(ISrsHttpResponseWriter* w, ISrsHttpM
 
     // We must do hook after stat, because depends on it.
     if ((err = http_hooks_on_play(req)) != srs_success) {
+        stat->on_disconnect(ctx);
```
winlinvip, line 153: Before calling this `serve_new_session`, the `alive` function must have been called, which will invoke `stat->on_disconnect` on timeout. Why is it being called separately here?

Or in other words, why is there no timeout cleanup for the session?

`TRANS_BY_GPT3`

duiniuluantanqin: The reasons for not having a timeout cleanup are probably as follows:
```
alive(ctx)                   // At this time, ctx (referred to as A) is empty, put it into map_ctx_info_, and start polling on_timer.
        |
ctx = srs_random_str(8)      // Generate a new ctx (referred to as B) here
        |
stat->on_client(ctx)         // Use the previously generated ctx (B) here
        |
http_hooks_on_play           // If successful, B will also be placed in map_ctx_info_ and can be cleaned up by timeout.
                             // But if it fails, map_ctx_info_ only has A without B, so stat cannot be cleaned up by on_timer.

```
   

`TRANS_BY_GPT3`

winlinvip: Take more time to consider the plan, no rush.

`TRANS_BY_GPT3`

PR #3286, opened 2022-11-28
H265: Demux sps for log print and statistic streams.(#3271)

### LINKED ISSUE #3271: HEVC: The codec information is incorrect.
**Description**

For HEVC, the stream description is incorrect.

1. SRS Version: `6.0.2`

1. SRS Log:

```
2457B video sh,  codec(12, profile=Other, level=Other, 0x0, 0kbps, 0.0fps, 0.0s)
```

1. SRS Config: `conf/hevc.flv.conf`

**Replay**:

> Please describe how to replay the bug? (Steps to reproduce the bug)

1. `./objs/srs -c conf/hevc.flv.conf`
1. `./ffmpeg -re -i ~/srs/doc/source.flv -acodec copy -vcodec libx265 -f flv rtmp://localhost/live/livestream`
1. Check log of SRS.

**Expect (Expected behavior)**

Should show correct metadata about stream, such as the profile, resolution and fps, etc.

The HTTP API stream information is also incorrect: http://localhost:1985/api/v1/streams/

`TRANS_BY_GPT3`

chundonglinlin:
When you experience `./configure --h265=on`, print detail codec infomation and statistic streams. Likely,
```Log
[2022-11-28 13:22:54.854][INFO][18226][05w7b88u] 2088B video sh, codec(12, profile=Main, level=3.1, 1280x720, 0kbps, 0.0fps, 0.0s)
[2022-11-28 13:22:54.854][INFO][18226][05w7b88u] 7B audio sh, codec(10, profile=LC, 2channels, 0kbps, 44100HZ), flv(16bits, 2channels, 44100HZ)
```
curl "http://127.0.0.1:1985/api/v1/streams/"
```Streams
{
    "code": 0, 
    "server": "vid-5x95fw5", 
    "streams": [
        {
            "id": "vid-8upx759", 
            "name": "livestream", 
            "vhost": "vid-4034os0", 
            "app": "live", 
            "tcUrl": "rtmp://127.0.0.1:1935/live", 
            "url": "/live/livestream", 
            "live_ms": 1669641839460, 
            "clients": 1, 
            "frames": 0, 
            "send_bytes": 4425, 
            "recv_bytes": 295733, 
            "kbps": {
                "recv_30s": 0, 
                "send_30s": 0
            }, 
            "publish": {
                "active": true, 
                "cid": "75167s02"
            }, 
            "video": {
                "codec": "HEVC", 
                "profile": "Main", 
                "level": "3.1", 
                "width": 1280, 
                "height": 720
            }, 
            "audio": {
                "codec": "AAC", 
                "sample_rate": 44100, 
                "channel": 2, 
                "profile": "LC"
            }
        }
    ]
}
```

### DIFF
--- a/trunk/src/kernel/srs_kernel_buffer.hpp
+++ b/trunk/src/kernel/srs_kernel_buffer.hpp
@@ -174,6 +174,13 @@ public:
 public:
     bool empty();
     int8_t read_bit();
+    int bits_left();
+    void skip_bits(int n);
+    int32_t read_bits(int n);
+    int8_t read_8bits();
+    int16_t read_16bits();
+    int32_t read_32bits();
+    int32_t read_bits_ue();
 };
 
 #endif
--- a/trunk/src/kernel/srs_kernel_buffer.hpp
+++ b/trunk/src/kernel/srs_kernel_buffer.hpp
@@ -174,6 +174,7 @@ public:
 public:
     bool empty();
     int8_t read_bit();
+    bool require(int required_size);
     int bits_left();
     void skip_bits(int n);
     int32_t read_bits(int n);
--- a/trunk/src/kernel/srs_kernel_buffer.hpp
+++ b/trunk/src/kernel/srs_kernel_buffer.hpp
@@ -174,8 +174,8 @@ public:
 public:
     bool empty();
     int8_t read_bit();
-    bool require(int required_size);
-    int bits_left();
+    bool require_bits(int n);
+    int left_bits();
     void skip_bits(int n);
     int32_t read_bits(int n);
     int8_t read_8bits();
--- a/trunk/src/kernel/srs_kernel_codec.cpp
+++ b/trunk/src/kernel/srs_kernel_codec.cpp
@@ -1076,14 +1076,14 @@ srs_error_t SrsFormat::hevc_demux_sps(SrsHevcHvccNalu* nal)
     // for NALU, ITU-T H.265 7.3.2.2 Sequence parameter set RBSP syntax
     // T-REC-H.265-202108-I!!PDF-E.pdf, page 61.
     if (!stream.require(1)) {
-        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "decode SPS");
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "decode hevc sps requires 1 only %d bytes", stream.left());
     }
     int8_t nutv = stream.read_1bytes();
 
     // forbidden_zero_bit shall be equal to 0.
     int8_t forbidden_zero_bit = (nutv >> 7) & 0x01;
     if (forbidden_zero_bit) {
-        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "forbidden_zero_bit shall be equal to 0");
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "hevc forbidden_zero_bit=%d shall be equal to 0", forbidden_zero_bit);
     }
 
     // 7.4.1 NAL unit semantics
@@ -1091,7 +1091,7 @@ srs_error_t SrsFormat::hevc_demux_sps(SrsHevcHvccNalu* nal)
     // nal_unit_type specifies the type of RBSP data structure contained in the NAL unit as specified in Table 7-1.
     SrsHevcNaluType nal_unit_type = (SrsHevcNaluType)((nutv >> 1) & 0x3f);
     if (nal_unit_type != SrsHevcNaluType_SPS) {
-        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "for sps, nal_unit_type shall be equal to 33");
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "hevc sps nal_unit_type=%d shall be equal to 33", nal_unit_type);
     }
 
     // nuh(nuh_layer_id + nuh_temporal_id_plus1)
@@ -1158,7 +1158,7 @@ srs_error_t SrsFormat::hevc_demux_sps_rbsp(char* rbsp, int nb_rbsp)
      *      }
      */
     if (!stream.require(2)) {
-        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "sps shall atleast 3bytes");
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "sps requires 2 only %d bytes", stream.left());
     }
     uint8_t nutv = stream.read_1bytes();
     (void)nutv;
@@ -1228,8 +1228,8 @@ srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfi
     srs_error_t err = srs_success;
 
     if (profile_resent_flag) {
-        if (!bs->require(88)) {
-            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl general profile tier shall atleast 88bits");
+        if (!bs->require_bits(88)) {
+            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl profile requires 88 only %d bits", bs->left_bits());
         }
 
         ptl->general_profile_space = bs->read_bits(2);
@@ -1277,8 +1277,8 @@ srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfi
         }
     }
 
-    if (!bs->require(8)) {
-        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl general level idc shall atleast 8bits");
+    if (!bs->require_bits(8)) {
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl level requires 8 only %d bits", bs->left_bits());
     }
 
     // level
@@ -1287,16 +1287,16 @@ srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfi
     ptl->sub_layer_profile_present_flag.resize(max_sub_layers_minus1);
     ptl->sub_layer_level_present_flag.resize(max_sub_layers_minus1);
     for (int i = 0; i < max_sub_layers_minus1; i++) {
-        if (!bs->require(2)) {
-            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl sub layer present_flag shall atleast 2bits");
+        if (!bs->require_bits(2)) {
+            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl present_flag requires 2 only %d bits", bs->left_bits());
         }
         ptl->sub_layer_profile_present_flag[i] = bs->read_bit();
         ptl->sub_layer_level_present_flag[i]   = bs->read_bit();
     }
 
     for (int i = max_sub_layers_minus1; max_sub_layers_minus1 > 0 && i < 8; i++) {
-        if (!bs->require(2)) {
-            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl reserved_zero shall atleast 2bits");
+        if (!bs->require_bits(2)) {
+            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl reserved_zero requires 2 only %d bits", bs->left_bits());
         }
         ptl->reserved_zero_2bits[i] = bs->read_bits(2);
     }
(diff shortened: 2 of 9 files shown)


trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1024,344 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    // FIXME: Support for multiple SPS, then pick the first non-empty one.
+    char* sps = (char*)(&nal->nal_data_vec[0].nal_unit_data[0]);
+    int nb_sps = nal->nal_data_vec[0].nal_unit_length;
+
+    SrsBuffer stream(sps, nb_sps);
+
+    // for NALU, ITU-T H.265 7.3.2.2 Sequence parameter set RBSP syntax
+    // T-REC-H.265-202108-I!!PDF-E.pdf, page 61.
+    if (!stream.require(1)) {
+        return srs_error_new(ERROR_HLS_DECODE_ERROR, "decode SPS");
```
winlinvip: Previously, only HLS required NALU parsing, so the error code used was `ERROR_HLS_DECODE_ERROR`. However, later on, this error code became inaccurate because the API also requires SPS parsing.

So it would be better to create a new error code, starting with `ERROR_HEVC_`.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1024,344 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    (void)sps_video_parameter_set_id;
+    (void)sps_max_sub_layers_minus1;
+    (void)sps_temporal_id_nesting_flag;
+
+    SrsBitBuffer bs(&stream);
+
+    // profile tier level...
+    SrsHevcSpsProfileTierLevel profile_tier_level;
+    memset(&profile_tier_level, 0, sizeof(SrsHevcSpsProfileTierLevel));
+    hevc_demux_sps_rbsp_ptl(&bs, &profile_tier_level, 1, sps_max_sub_layers_minus1);
```
winlinvip: This function `hevc_demux_sps_rbsp_ptl` has a return error, so it is necessary to check the error code.

1. Even if this function does not currently return an error, there is no guarantee that it won't return an error in the future.
2. If an error is returned, it signifies abnormal logic. Continuing to proceed with unknown behavior is extremely dangerous.
3. Even if there are no apparent issues, there may still be memory leakage problems. The error object is a complex object, not an integer.

Please check if other functions have this issue.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_buffer.hpp
```diff
@@ -174,6 +174,13 @@ class SrsBitBuffer
 public:
     bool empty();
     int8_t read_bit();
+    int bits_left();
+    void skip_bits(int n);
+    int32_t read_bits(int n);
+    int8_t read_8bits();
+    int16_t read_16bits();
+    int32_t read_32bits();
+    int32_t read_bits_ue();
```
winlinvip, line 184: This SrsBitBuffer utility class is relatively easy to validate with unit tests. The previous functions have coverage.
```cpp
VOID TEST(KernelUtility, AvcUev)
```

Please validate the newly added functions with unit tests to avoid introducing any issues.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,349 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    (void)sps_temporal_id_nesting_flag;
+
+    SrsBitBuffer bs(&stream);
+
+    // profile tier level...
+    SrsHevcSpsProfileTierLevel profile_tier_level;
+    memset(&profile_tier_level, 0, sizeof(SrsHevcSpsProfileTierLevel));
+
+    if ((err = hevc_demux_sps_rbsp_ptl(&bs, &profile_tier_level, 1, sps_max_sub_layers_minus1)) != srs_success) {
+        return srs_error_wrap(err, "sps rbsp ptl demux failed");
```
winlinvip: Here we need to consider: what information do we need to know when an error occurs.

This function has four parameters, the first three are output parameters, and the last one is an input parameter. The error is very likely caused by this last variable parameter, or we need to know its data when troubleshooting.

A better approach is to add this new variable at this level to the error.

```cpp
    if ((err = hevc_demux_sps_rbsp_ptl(&bs, &profile_tier_level, 1, sps_max_sub_layers_minus1)) != srs_success) {
        return srs_error_wrap(err, "sps rbsp ptl sps_max_sub_layers_minus1=%d", sps_max_sub_layers_minus1);
```

Note that since this parameter has already been added here, there is no need to add this variable again when an error occurs in the function `hevc_demux_sps_rbsp_ptl`. It only focuses on the variables **newly generated** at its own level.

By extension, when an error occurs, we are aware of the variables at each level. Even if we don't have a coredump, we can still understand what happened, and we can quickly identify the problem. It is easy for everyone to see the error logs, but obtaining the complete stack trace can be difficult because most of the time there is no coredump available.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,349 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    ptl->sub_layer_profile_present_flag.resize(max_sub_layers_minus1);
+    ptl->sub_layer_level_present_flag.resize(max_sub_layers_minus1);
+    for (i = 0; i < max_sub_layers_minus1; i++) {
+        ptl->sub_layer_profile_present_flag[i] = bs->read_bit();
+        ptl->sub_layer_level_present_flag[i]   = bs->read_bit();
+    }
+
+    if (max_sub_layers_minus1 > 0) {
+        for (i = max_sub_layers_minus1; i < 8; i++)
+        {
```
winlinvip: This place, put the curly braces on the next line after a layer:
```cpp
for (xxx) {
```

Additionally, here we can have one less level of nesting. Having one less level of nesting means having one less level of understanding. It can be changed to:

```cpp
for (i = max_sub_layers_minus1; max_sub_layers_minus1 > 0 && i < 8; i++) {
```

Reducing indentation can reduce the difficulty of understanding.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,349 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    return err;
+}
+
+srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfileTierLevel* ptl, int profile_resent_flag, int max_sub_layers_minus1)
+{
+    srs_error_t err = srs_success;
+
+    int i = 0;
+    if (profile_resent_flag) {
+        ptl->general_profile_space = bs->read_bits(2);
```
winlinvip, line 1240: This function does not perform any checks and directly calls various `bs->read_xxx` functions, while this function directly calls assert.

```cpp
int32_t SrsBitBuffer::read_bits(int n)
{
    srs_assert(n <= bits_left());
```

In other words, if an attacker forges a packet, they can bring the server down.

If the code can only run under normal circumstances and not under exceptional circumstances, it is of low quality.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,349 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    // 7.3.2.11 RBSP trailing bits syntax
+
+    return err;
+}
+
+srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfileTierLevel* ptl, int profile_resent_flag, int max_sub_layers_minus1)
+{
+    srs_error_t err = srs_success;
+
+    int i = 0;
```
winlinvip: Do not declare variables in advance, as this is a major drawback in C. Try to delay variable declaration and reduce its scope as much as possible.

For example, if we declare variables inside a for loop, then each for loop will have its own local variables without any connection:

```cpp
for (int i = 0; i < max_sub_layers_minus1; i++) {

for (int i = max_sub_layers_minus1; max_sub_layers_minus1 > 0 && i < 8; i++) {
```

We can quickly determine that these two "i" variables have no connection; they only take effect within their respective for loops. However, with the current approach, we need to carefully consider the lifespan of this "i" variable, as it remains effective throughout the entire function and may be reused.

```cpp
int i = 0;

// xxxx a lot of code

for (i = 0; i < max_sub_layers_minus1; i++) {

// xxxx a large amount of code

if (max_sub_layers_minus1 > 0) {
        for (i = max_sub_layers_minus1; i < 8; i++) {
// Here, we need to be very careful in judging whether we have used the previous value.
```

This has brought about complexity in maintenance and understanding that originally did not exist. What was originally a simple matter has ended up becoming complicated.

If this bad habit is developed, and this kind of code is everywhere, it will be very poor quality code.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,349 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+        for (i = max_sub_layers_minus1; i < 8; i++)
+        {
+            ptl->reserved_zero_2bits[i] = bs->read_bits(2);
+        }
+    }
+    ptl->sub_layer_profile_space.resize(max_sub_layers_minus1);
+    ptl->sub_layer_tier_flag.resize(max_sub_layers_minus1);
+    ptl->sub_layer_profile_idc.resize(max_sub_layers_minus1);
+    ptl->sub_layer_profile_compatibility_flag.resize(max_sub_layers_minus1);
+    for (int j = 0; j < max_sub_layers_minus1; j++) {
```
winlinvip: Here, you can use "i" instead of "j" because that "i" is locally valid.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,371 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    return err;
+}
+
+srs_error_t SrsFormat::hevc_demux_sps_rbsp_ptl(SrsBitBuffer* bs, SrsHevcSpsProfileTierLevel* ptl, int profile_resent_flag, int max_sub_layers_minus1)
+{
+    srs_error_t err = srs_success;
+
+    if (profile_resent_flag) {
+        if (!bs->require(88)) {
+            return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl general profile tier shall atleast 88bits");
```
winlinvip: There is an implicit variable here, which is the actual remaining bits. So the error message should include this information.

```cpp
return srs_error_new(ERROR_HEVC_DECODE_ERROR, "ptl requires 88 only %d bits", bs->bits_left());
```

> Note: There is no need to write so many constants. The errors already include line numbers and functions, so we only need a descriptive constant. Therefore, instead of `ptl general profile tier shall atleast`, it can be directly changed to `ptl requires` or `ptl profile requires`.

Please also make changes in other places.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_buffer.hpp
```diff
@@ -174,6 +174,14 @@ class SrsBitBuffer
 public:
     bool empty();
     int8_t read_bit();
+    bool require(int required_size);
+    int bits_left();
```
winlinvip: It would be better to change it to `left_bits()`, as other functions have bits as a suffix.

It would be better to change `require` to `require_bits` to differentiate it from `SrsBuffer::require`.

The newly added function should have a unit test to test the scenario where there are not enough bits.

The unit test should consider exceptional situations, that is, problematic or abnormal scenarios, and not just focus on normal readings.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,371 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    if (forbidden_zero_bit) {
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "forbidden_zero_bit shall be equal to 0");
+    }
+
+    // 7.4.1 NAL unit semantics
+    // ISO_IEC_14496-10-AVC-2012.pdf, page 61.
+    // nal_unit_type specifies the type of RBSP data structure contained in the NAL unit as specified in Table 7-1.
+    SrsHevcNaluType nal_unit_type = (SrsHevcNaluType)((nutv >> 1) & 0x3f);
+    if (nal_unit_type != SrsHevcNaluType_SPS) {
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "for sps, nal_unit_type shall be equal to 33");
```
winlinvip: There are variables here, please check all error generations, and variables must be brought in.

If nal_unit_type is not SPS, it will throw an error. What is the error when it occurs? I will be completely confused if there is an error.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_codec.cpp
```diff
@@ -989,10 +1026,371 @@ srs_error_t SrsFormat::hevc_demux_hvcc(SrsBuffer* stream)
...
+    }
+    int8_t nutv = stream.read_1bytes();
+
+    // forbidden_zero_bit shall be equal to 0.
+    int8_t forbidden_zero_bit = (nutv >> 7) & 0x01;
+    if (forbidden_zero_bit) {
+        return srs_error_new(ERROR_HEVC_DECODE_ERROR, "forbidden_zero_bit shall be equal to 0");
+    }
+
+    // 7.4.1 NAL unit semantics
```
winlinvip: Are the referenced documents and chapters incorrect? Comments should be treated as code.

`TRANS_BY_GPT3`

PR #3289, opened 2022-12-01
Live: Fix bug for gop cache limits. v5.0.99

bluestn:
1. set rtmp gop cache limit while using srt to push stream & bridge to rtmp source.
2. set rtmp gop cache limit while http_stream playing the live stream.

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -8,6 +8,7 @@ The changelog for SRS.
 
 ## SRS 5.0 Changelog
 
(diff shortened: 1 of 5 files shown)


winlinvip:
I see that you only changed SRT to RTMP, is there a problem with RTC to RTMP? What about GB to RTMP?

Additionally, this place can also be modified:
```cpp
    if (gop_cache.size() > (size_t)gop_cache_max_frames_) {
```

If gop_cache_max_frames_ is 0, do not enable this capability without initialization. For example, modify it to:
```cpp
    if (gop_cache_max_frames_ && gop_cache.size() > (size_t)gop_cache_max_frames_) {
```

`TRANS_BY_GPT3`

PR #3302, opened 2022-12-13
Build: Fix build warnings

winlinvip:
1. Avoid default guess for expression.
2. Force to void* for memset.


chundonglinlin (approved):
LGTM

PR #3306, opened 2022-12-13
FLV: Drop packet if header flag is not matched. v5.0.109

### LINKED ISSUE #939: HTTP: FLV or TS stream outputs the correct header. 浏览器下下载的Http-flv流和ffmepg下拉取的流flv头不同
我的网络摄像头是没有音频的，然后我用下面的命令将摄像机的`rtsp`流推到`srs`服务器上，配置文件采用的是[SRS-HTTP-FLV部署实例](https://github.com/ossrs/srs/wiki/v2_CN_SampleHttpFlv)。

- `ffmpeg -i rtsp://** -vcodec libx264 -acodec aac -f flv  rtmp://192.168.31.221/live/livestream`

之后我在`ffmpeg`上使用下面命令拉取`http-flv`流到`srs.flv`文件中：

- `ffmpeg.exe -i http://192.168.31.221:8080/live/livestream.flv -vcodec copy -acodec copy -f flv srs.flv`
然后我用`FlvParse`查看`srs.flv`结构，`has audio=0`，没有`Audio Tag`，和实际情况是一样的。但是另外一方面，我把链接`http://192.168.31.221:8080/live/livestream.flv`直接输入到浏览器中，下载的文件`has audio=1`，但是没有`Audio tag`，这是什么原因呢？

winlinvip:
1. Ignore audo or video packets if FLV header disable it.
2. Reset has_audio if got some video frames but no audio frames.
3. Reset has_video if got some audio frames but no video frames.
4. Note that audio/video frames are not sequence header.

### DIFF
--- a/trunk/src/app/srs_app_http_stream.cpp
+++ b/trunk/src/app/srs_app_http_stream.cpp
@@ -289,6 +289,11 @@ srs_error_t SrsFlvStreamEncoder::write_metadata(int64_t timestamp, char* data, i
     return enc->write_metadata(SrsFrameTypeScript, data, size);
 }
 
+void SrsFlvStreamEncoder::set_drop_if_not_match(bool v)
+{
+    enc->set_drop_if_not_match(v);
+}
+
 bool SrsFlvStreamEncoder::has_cache()
 {
     // for flv stream, use gop cache of SrsLiveSource is ok.
@@ -343,7 +348,7 @@ srs_error_t SrsFlvStreamEncoder::write_header(bool has_video, bool has_audio)
             return srs_error_wrap(err, "write header");
         }
 
-        srs_trace("FLV: write header audio=%d, video=%d", has_audio, has_video);
+        srs_trace("FLV: write header audio=%d, video=%d, dinm=%d", has_audio, has_video, enc->drop_if_not_match());
     }
 
     return err;
@@ -563,12 +568,15 @@ srs_error_t SrsLiveStream::do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMess
     
     string enc_desc;
     ISrsBufferEncoder* enc = NULL;
-    
+
     srs_assert(entry);
+    bool drop_if_not_match = _srs_config->get_vhost_http_remux_drop_if_not_match(req->vhost);
+
     if (srs_string_ends_with(entry->pattern, ".flv")) {
         w->header()->set_content_type("video/x-flv");
         enc_desc = "FLV";
         enc = new SrsFlvStreamEncoder();
+        ((SrsFlvStreamEncoder*)enc)->set_drop_if_not_match(drop_if_not_match);
     } else if (srs_string_ends_with(entry->pattern, ".aac")) {
         w->header()->set_content_type("audio/x-aac");
         enc_desc = "AAC";
@@ -638,8 +646,8 @@ srs_error_t SrsLiveStream::do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMess
     }
 
     srs_utime_t mw_sleep = _srs_config->get_mw_sleep(req->vhost);
-    srs_trace("FLV %s, encoder=%s, mw_sleep=%dms, cache=%d, msgs=%d", entry->pattern.c_str(), enc_desc.c_str(),
-        srsu2msi(mw_sleep), enc->has_cache(), msgs.max);
+    srs_trace("FLV %s, encoder=%s, mw_sleep=%dms, cache=%d, msgs=%d, dinm=%d", entry->pattern.c_str(), enc_desc.c_str(),
+        srsu2msi(mw_sleep), enc->has_cache(), msgs.max, drop_if_not_match);
 
     // TODO: free and erase the disabled entry after all related connections is closed.
     // TODO: FXIME: Support timeout for player, quit infinite-loop.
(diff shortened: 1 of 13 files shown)


trunk/src/app/srs_app_http_stream.cpp
```diff
@@ -307,18 +307,31 @@ srs_error_t SrsFlvStreamEncoder::write_tags(SrsSharedPtrMessage** msgs, int coun
 
     // For https://github.com/ossrs/srs/issues/939
```
xiaozhihong: Is it possible that calling this function only dumps to a single package (audio, video) and causes misjudgment?

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_flv.cpp
```diff
@@ -481,17 +490,19 @@ srs_error_t SrsFlvTransmuxer::write_tags(SrsSharedPtrMessage** msgs, int count)
...
-    }
+    do {
+        int nn_might_iovss = 3 * count;
+        if (nb_iovss_cache < nn_might_iovss) {
+            srs_freepa(iovss_cache);
+
+            nb_iovss_cache = nn_might_iovss;
+            iovss = iovss_cache = new iovec[nn_might_iovss];
+        }
+    } while (false);
```
xiaozhihong, line 514: This do while seems to have no effect, can we just remove it?

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_flv.cpp
```diff
@@ -444,6 +449,8 @@ srs_error_t SrsFlvTransmuxer::write_metadata(char type, char* data, int size)
 srs_error_t SrsFlvTransmuxer::write_audio(int64_t timestamp, char* data, int size)
 {
     srs_error_t err = srs_success;
+
+    if (!has_audio_) return err;
```
xiaozhihong: The single line is concise, but the code style doesn't match well with the overall SRS.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_flv.cpp
winlinvip: The main purpose is to hide the scope of `nn_might_iovss`. It would cause issues if this variable is used later on.

The variable that should be used later on is `nn_real_iovss`.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_http_stream.cpp
winlinvip: If there is a delay in the flow, such as when converting RTC to RTMP, it sometimes happens. That is definitely a misjudgment, and there is no way to guess it accurately.

If the gop cache is disabled, it still cannot be guessed accurately. In this case, there is only a Sequence Header, and in some exceptional scenarios, there may be AV Sequence Header even though there are no audio packets. In such cases, it is also inaccurate.

Therefore, this issue still requires support for parameters or configuration settings in order to be fully resolved. [Link to the issue](https://github.com/ossrs/srs/issues/939#issuecomment-1348740526)

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_flv.cpp
winlinvip: If it can be done in one line, it is preferable not to make it three lines. Having fewer lines is more friendly for reading code.

For example, this function:

```cpp
srs_error_t SrsFlvTransmuxer::write_tags(SrsSharedPtrMessage** msgs, int count) {
    for (int i = 0; i < count; i++) {
        if (msg->is_audio()) {
            if (!has_audio_) continue; // Ignore audio packets if no audio stream.
            cache_audio(msg->timestamp, msg->payload, msg->size, cache);
        } else if (msg->is_video()) {
            if (!has_video_) continue; // Ignore video packets if no video stream.
            cache_video(msg->timestamp, msg->payload, msg->size, cache);
```

If it is changed to multiple lines, it would be like this:

```cpp
srs_error_t SrsFlvTransmuxer::write_tags(SrsSharedPtrMessage** msgs, int count) {
    for (int i = 0; i < count; i++) {
        if (msg->is_audio()) {
            // Ignore audio packets if no audio stream.
            if (!has_audio_) {
                continue;
            }
            cache_audio(msg->timestamp, msg->payload, msg->size, cache);
        } else if (msg->is_video()) {
            // Ignore video packets if no video stream.
            if (!has_video_) {
                continue;
            }
            cache_video(msg->timestamp, msg->payload, msg->size, cache);
```

In this exceptional case, it is preferable to write it in one line. However, I don't think there is any problem with continuing to write it as before.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_flv.cpp
```diff
@@ -509,38 +520,41 @@ srs_error_t SrsFlvTransmuxer::write_tags(SrsSharedPtrMessage** msgs, int count)
...
-    iovec* iovs = iovss;
+    // Now all caches are ok, start to write all messages.
+    iovec* iovs = iovss; int nn_real_iovss = 0;
     for (int i = 0; i < count; i++) {
         SrsSharedPtrMessage* msg = msgs[i];
         
-        // cache all flv header.
+        // Cache FLV packet header.
         if (msg->is_audio()) {
+            if (!has_audio_) continue; // Ignore audio packets if no audio stream.
```
winlinvip: If the header does not have an audio or video tag, discard the packet. It feels like this should have an additional configuration, as sometimes the packet should not be discarded.

`TRANS_BY_GPT3`

xiaozhihong (approved):

PR #3308, opened 2022-12-14
Improve file writer performance by fwrite with cache.  v5.0.133

bluestn:
I found srs dvr bottleneck while doing dvr stress test with many concurrent streams pushing to srs.
with perf flame graph's help, I found there are so many file write ops which consumed so many cpu time.
And I tracked the dvr module code, and found the SrsFileWriter use unix system file operation functions including open/write/lseek/close, it has no buffer mechanism in it, and every write op will lead to a new system call.
With this analysis, I used corresponding libc  file operation functions to implement the SrsFileWriter, and use setvbuf with 65536 bytes buffer to enable the libc buffer mechanism, as i wished, the cpu usage rate drop down very obviously.

the following is the graph before and after the performance tuning.
before tuning:
before_tuning

after tuning:

after_tuning

### DIFF
--- a/trunk/src/kernel/srs_kernel_file.cpp
+++ b/trunk/src/kernel/srs_kernel_file.cpp
@@ -46,7 +46,6 @@ SrsFileWriter::~SrsFileWriter()
     srs_freepa(buf_);
 }
 
-
 srs_error_t SrsFileWriter::set_iobuf_size(int size)
 {
     srs_error_t err = srs_success;
@@ -56,24 +55,14 @@ srs_error_t SrsFileWriter::set_iobuf_size(int size)
     }
 
     srs_freepa(buf_);
-
-    int ret;
-
-    if (size > 0) {
-        buf_ = new char[size];
-        ret = _srs_setvbuf_fn(fp_, buf_, _IOFBF, size);
-    }
-    else {
-        ret = _srs_setvbuf_fn(fp_, NULL, _IONBF, size);
-    }
-
-    if (ret != 0) {
-        return srs_error_new(ERROR_SYSTEM_FILE_SETVBUF, "file %s set vbuf error", path_.c_str());
+    buf_ = size > 0 ? new char[size] : NULL;
+    int r0 = _srs_setvbuf_fn(fp_, buf_, _IOFBF, size);
+    if (r0) {
+        return srs_error_new(ERROR_SYSTEM_FILE_SETVBUF, "setvbuf err, file=%s, r0=%d", path_.c_str(), r0);
     }
 
     return err;
 }
-
 
 srs_error_t SrsFileWriter::open(string p)
 {
--- a/trunk/src/kernel/srs_kernel_file.cpp
+++ b/trunk/src/kernel/srs_kernel_file.cpp
@@ -121,16 +121,17 @@ bool SrsFileWriter::is_open()
 
 void SrsFileWriter::seek2(int64_t offset)
 {
+    srs_assert(is_open());
+
     int r0 = fseek(fd_, (long)offset, SEEK_SET);
     srs_assert(r0 != -1);
 }
 
 int64_t SrsFileWriter::tellg()
 {
-    fpos_t pos;
-    srs_assert(fgetpos(fd_, &pos) == 0);
-        
-    return (int64_t)pos.__pos;
+    srs_assert(is_open());
+
+    return ftell(fd_);
 }
 
 srs_error_t SrsFileWriter::write(void* buf, size_t count, ssize_t* pnwrite)
@@ -176,16 +177,14 @@ srs_error_t SrsFileWriter::writev(const iovec* iov, int iovcnt, ssize_t* pnwrite
 
 srs_error_t SrsFileWriter::lseek(off_t offset, int whence, off_t* seeked)
 {
+    srs_assert(is_open());
+
     if (fseek(fd_, (long)offset, whence) == -1) {
         return srs_error_new(ERROR_SYSTEM_FILE_SEEK, "seek file");
     }
 
     if (seeked) {
-        fpos_t pos;
-        if (fgetpos(fd_, &pos)){
-            return srs_error_new(ERROR_SYSTEM_FILE_SEEK, "get pos");
-        }
-        *seeked = pos.__pos;
+        *seeked = ftell(fd_);
     }
 
     return srs_success;
--- a/trunk/src/kernel/srs_kernel_file.cpp
+++ b/trunk/src/kernel/srs_kernel_file.cpp
@@ -28,30 +28,57 @@ srs_close_t _srs_close_fn = ::close;
 
 SrsFileWriter::SrsFileWriter()
 {
-    fd = -1;
+    fd_  = NULL;
+    buf_ = NULL;
 }
 
 SrsFileWriter::~SrsFileWriter()
 {
     close();
+    srs_freepa(buf_);
 }
+
+
+srs_error_t SrsFileWriter::srs_set_iobuf_size(int size)
+{
+    srs_error_t err = srs_success;
+
+    if (fd_ == NULL){
+        return srs_error_new(ERROR_SYSTEM_FILE_NOT_OPEN, "file %s is not opened", path_.c_str());
+    }
+
+    srs_freepa(buf_);
+
+    int ret;
+
+    if (size > 0){
+        buf_ = new char[size];
+        ret = setvbuf(fd_, buf_, _IOFBF, size);
+    }else{
+        ret = setvbuf(fd_, NULL, _IONBF, size);
+    }
+
+    if (ret != 0){
+        return srs_error_new(ERROR_SYSTEM_FILE_SETVBUF, "file %s set vbuf error", path_.c_str());
+    }
+
+    return err;
+}
+
 
 srs_error_t SrsFileWriter::open(string p)
 {
     srs_error_t err = srs_success;
     
-    if (fd > 0) {
+    if (fd_ != NULL) {
         return srs_error_new(ERROR_SYSTEM_FILE_ALREADY_OPENED, "file %s already opened", p.c_str());
     }
     
-    int flags = O_CREAT|O_WRONLY|O_TRUNC;
-    mode_t mode = S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH;
-    
-    if ((fd = _srs_open_fn(p.c_str(), flags, mode)) < 0) {
+    if ((fd_ = fopen(p.c_str(), "wb")) == NULL) {
         return srs_error_new(ERROR_SYSTEM_FILE_OPENE, "open file %s failed", p.c_str());
     }
     
-    path = p;
+    path_ = p;
     
     return err;
 }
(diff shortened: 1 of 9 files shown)


trunk/src/kernel/srs_kernel_file.cpp
```diff
@@ -28,30 +28,57 @@ srs_close_t _srs_close_fn = ::close;
...
+    }
+
+    srs_freepa(buf_);
+
+    int ret;
+
+    if (size > 0){
+        buf_ = new char[size];
+        ret = setvbuf(fd_, buf_, _IOFBF, size);
+    }else{
```
xiaozhihong: Format issue`} else {`

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_file.cpp
```diff
@@ -135,30 +162,31 @@ srs_error_t SrsFileWriter::writev(const iovec* iov, int iovcnt, ssize_t* pnwrite
     for (int i = 0; i < iovcnt; i++) {
         const iovec* piov = iov + i;
         ssize_t this_nwrite = 0;
-        if ((err = write(piov->iov_base, piov->iov_len, &this_nwrite)) != srs_success) {
-            return srs_error_wrap(err, "write file");
+        if ((err = write(piov->iov_base, piov->iov_len, &this_nwrite)) != NULL) {
```
xiaozhihong: `!= NULL` is it a typo? Should it be `!= srs_success`?

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_file.hpp
```diff
@@ -26,12 +26,18 @@ class SrsFileReader;
 class SrsFileWriter : public ISrsWriteSeeker
 {
 private:
-    std::string path;
-    int fd;
+    std::string path_;
+    FILE *fd_;
```
xiaozhihong: The code style of SRS is `FILE* fd_;`
Additionally, fd is the notation for descriptor, and the common name for FILE* is fp, which is easier to understand.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_file.hpp
```diff
@@ -26,12 +26,18 @@ class SrsFileReader;
...
+    char *buf_;
 public:
     SrsFileWriter();
     virtual ~SrsFileWriter();
 public:
+    /**
+     * set io buf size
+    */
+   virtual srs_error_t srs_set_iobuf_size(int size);
```
xiaozhihong: This function does not need to be prefixed with 'srs_', usually only single-purpose functions need to be prefixed with 'srs_'.

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
```diff
@@ -4273,13 +4228,60 @@ VOID TEST(KernelFileTest, ReadWriteCase)
...
+
+	SrsFileReader r;
+	HELPER_EXPECT_SUCCESS(r.open(filepath.c_str()));
+
+	ssize_t nn = 0;
+	HELPER_EXPECT_SUCCESS(w.write((void*)"Hello", 5, &nn));
+	EXPECT_EQ(5, nn);
+
+    // over 4g file test
+    long seek_pos = 0x100000002l;
```
xiaozhihong: Indentation is problematic, do not use tabs, use spaces

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_file.cpp
```diff
@@ -60,70 +87,70 @@ srs_error_t SrsFileWriter::open_append(string p)
...
-        return srs_error_new(ERROR_SYSTEM_FILE_WRITE, "write to file %s failed", path.c_str());
+
+    if (fd_ == NULL) {
+        return srs_error_new(ERROR_SYSTEM_FILE_NOT_OPEN, "file %s is not opened", path_.c_str());
     }
     
-    if (pnwrite != NULL) {
-        *pnwrite = nwrite;
+    size_t n = fwrite(buf, 1, count, fd_);
+    if (n != count){
```
xiaozhihong: `) {` Add a space in the middle

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_file.cpp
```diff
@@ -60,70 +87,70 @@ srs_error_t SrsFileWriter::open_append(string p)
...
+        return srs_error_new(ERROR_SYSTEM_FILE_WRITE, "write to file %s failed", path_.c_str());
     }
-    
+
+    if (pnwrite != NULL){
```
xiaozhihong: `) {` Add a space in the middle

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
```diff
@@ -4111,20 +4111,6 @@ VOID TEST(KernelFileWriterTest, WriteSpecialCase)
 		HELPER_EXPECT_FAILED(f.open_append("/dev/null"));
 	}
 
-	// Always fail.
```
xiaozhihong, line 4126: Why were these tests removed?

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
```diff
@@ -4154,37 +4140,6 @@ VOID TEST(KernelFileWriterTest, WriteSpecialCase)
 #endif
 	}
 
-	// Always fail.
```
xiaozhihong, line 4157: 这些test为啥去掉了

trunk/src/kernel/srs_kernel_file.cpp
xiaozhihong: Is it better to not handle size <= 0? man setvbuf, _IONBF should mean no buffer, so it's the same as write?

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
bluestn: The previous test used MockSystemIO to set _srs_open_fn as the mock_open function. After using the fwrite function, _srs_open_fn is no longer used, so the failed test cases will definitely not pass.

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
bluestn: 之前的测试是用MockSystemIO设置_srs_write_fn为mock_write函数，用了fwrite函数以后就不用_srs_write_fn不上了，每次失败测试用例肯定不能通过的，所以去掉了这个测试用例。

trunk/src/kernel/srs_kernel_file.cpp
bluestn: size<=0, calling setvbuf(fp_, NULL, _IONBF, size); means to disable the write buffer, providing the upper-level application with the choice of whether to enable or disable the cache. Although in our DVR, we still need to set size>0 to enable the cache functionality.

`TRANS_BY_GPT3`

trunk/src/utest/srs_utest_kernel.cpp
winlinvip: That requires a mock that supports fopen, if there is no utest coverage, we won't know if there are any issues.

`TRANS_BY_GPT3`

winlinvip:
## By Winlin

I remember why "write" is used instead of "fwrite". It's because "st_write" can also write files, and it is a non-blocking operation.

So actually, using "fwrite" from libc has better performance, but there are two issues that need to be discussed:

1. For servers, non-blocking is more crucial, and "write" can be easily changed to "st_write".

2. It is generally recommended to use memory disks, and whether "write" will have performance similar to "fwrite" for writing to disks.

## By John

`st_write` cannot perform non-blocking file writing, right?

## By Bluestn

`epoll` does not support files.

nginx does not use epoll to support file read/write, but instead uses aio.

## By Chen Yu

It's not that epoll doesn't support it, it's that even ext4 hasn't implemented the poll callback, so it can't be achieved.

The implementation of poll is required for it to be used with select/poll/epoll.

Another issue with `fwrite` is that it cannot implement `writev`.

The advantages of `aio` don't seem very obvious.

## By Winlin

The main purpose of `writev` is to reduce syscalls. In fact, if `fwrite` has a user space buffer, it can also work without supporting `writev`.

Implementing it yourself is also a solution.

Creating a thread is also considered as a final solution. However, you can also use fwrite within the thread.

The main issue with threads currently is that cygwin64 SRT crashes. This problem needs to be resolved before creating threads.

`TRANS_BY_GPT3`

winlinvip:
Benchmark data:

1. macOS: MacBook Pro, 16-inch, 2019, 12CPU(2.6 GHz 6-Core Intel Core i7), 16GB memory(16 GB 2667 MHz DDR4).
1. v5.0.132 without this PR: RTMP to HLS, **200 streams**, SRS **CPU 87%, 740MB**
1. v5.0.133 with this PR: RTMP to HLS, **200 streams**, SRS **CPU 56%, 618MB**
1. v5.0.132 without this PR: DVR RTMP to FLV, **500 streams**, SRS **CPU 83%, 759MB**
1. v5.0.133 with this PR: DVR RTMP to FLV, **500 streams**, SRS **CPU 35%, 912MB**
1. v5.0.133 with this PR: DVR RTMP to FLV, **1200 streams**, SRS **CPU 79%, 1590MB**

Run SRS for HLS:

```bash
env SRS_LISTEN=1935 SRS_DAEMON=off SRS_SRS_LOG_TANK=console SRS_HTTP_API_ENABLED=on \
  SRS_VHOST_HLS_ENABLED=on ./objs/srs -e
```

Run SRS for DVR:

```bash
env SRS_LISTEN=1935 SRS_DAEMON=off SRS_SRS_LOG_TANK=console SRS_HTTP_API_ENABLED=on \
  SRS_VHOST_DVR_ENABLED=on ./objs/srs -e
```

Publish 100 streams by [srs-bench](https://github.com/ossrs/srs-bench):

```bash
./objs/sb_rtmp_publish  -i doc/source.200kbps.768x320.flv -c 100 -r rtmp://127.0.0.1:1935/live/livestream_{i}
```

Publish another 200 streams:

```bash
./objs/sb_rtmp_publish -i doc/source.200kbps.768x320.flv -c 200 -r rtmp://127.0.0.1:1935/live/livestream2_{i}
```

winlinvip (approved):

trunk/src/kernel/srs_kernel_file.cpp
```diff
@@ -26,32 +26,68 @@ srs_read_t _srs_read_fn = ::read;
...
+        ret = _srs_setvbuf_fn(fp_, buf_, _IOFBF, size);
+    }
+    else {
+        ret = _srs_setvbuf_fn(fp_, NULL, _IONBF, size);
+    }
+
+    if (ret != 0) {
+        return srs_error_new(ERROR_SYSTEM_FILE_SETVBUF, "file %s set vbuf error", path_.c_str());
+    }
+
```
winlinvip, line 63: I made some changes to this section:

```cpp
    srs_freepa(buf_);
    buf_ = size > 0 ? new char[size] : NULL;
    int r0 = _srs_setvbuf_fn(fp_, buf_, _IOFBF, size);
    if (!r0) {
        return srs_error_new(ERROR_SYSTEM_FILE_SETVBUF, "setvbuf err, file=%s, r0=%d", path_.c_str(), r0);
    }
```

Special attention should be paid to the fact that the previous value of "ret" was lost.

`TRANS_BY_GPT3`

winlinvip (approved):

PR #3321, opened 2022-12-17
GB: Refine lazy object GC. v5.0.114

winlinvip:
1. Remove gc_set_creator_wrapper, pass by resource constructor.
2. Remove SRS_LAZY_WRAPPER_GENERATOR macro, use template directly.
3. Remove interfaces ISrsGbSipConn and ISrsGbSipConnWrapper.
4. Remove ISrsGbMediaConn and ISrsGbMediaConnWrapper.

### DIFF
--- a/trunk/src/app/srs_app_gb28181.cpp
+++ b/trunk/src/app/srs_app_gb28181.cpp
@@ -70,42 +70,11 @@ std::string srs_sip_state(SrsGbSipState ostate, SrsGbSipState state)
     return srs_fmt("%s->%s", srs_gb_sip_state(ostate).c_str(), srs_gb_sip_state(state).c_str());
 }
 
-ISrsGbSipConn::ISrsGbSipConn()
+SrsLazyGbSession::SrsLazyGbSession(SrsLazyObjectWrapper<SrsLazyGbSession>* wrapper_root)
 {
-}
-
-ISrsGbSipConn::~ISrsGbSipConn()
-{
-}
-
-ISrsGbSipConnWrapper::ISrsGbSipConnWrapper()
-{
-}
-
-ISrsGbSipConnWrapper::~ISrsGbSipConnWrapper()
-{
-}
-
-ISrsGbMediaConn::ISrsGbMediaConn()
-{
-}
-
-ISrsGbMediaConn::~ISrsGbMediaConn()
-{
-}
-
-ISrsGbMediaConnWrapper::ISrsGbMediaConnWrapper()
-{
-}
-
-ISrsGbMediaConnWrapper::~ISrsGbMediaConnWrapper()
-{
-}
-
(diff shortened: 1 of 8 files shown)


trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -1248,8 +1218,9 @@ ISrsPsPackHandler::~ISrsPsPackHandler()
 {
 }
 
-SrsLazyGbMediaTcpConn::SrsLazyGbMediaTcpConn()
+SrsLazyGbMediaTcpConn::SrsLazyGbMediaTcpConn(SrsLazyObjectWrapper<SrsLazyGbMediaTcpConn>* wrapper_root)
 {
+    wrapper_root_ = wrapper_root;
```
lipeng19811218, line 1223: 这里应该是wrapper_root->copy()吧？
要不外面删除wrapper_root，里面的wrapper_root_也是非法指针了

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -522,8 +491,9 @@ srs_error_t SrsGbListener::on_tcp_client(ISrsListener* listener, srs_netfd_t stf
     return err;
 }
 
-SrsLazyGbSipTcpConn::SrsLazyGbSipTcpConn()
+SrsLazyGbSipTcpConn::SrsLazyGbSipTcpConn(SrsLazyObjectWrapper<SrsLazyGbSipTcpConn>* wrapper_root)
 {
+    wrapper_root_ = wrapper_root;
```
lipeng19811218, line 496: Shouldn't it be wrapper_root->copy() here?
Otherwise, if wrapper_root is deleted outside, wrapper_root_ inside will also become an invalid pointer.

`TRANS_BY_GPT3`

lipeng19811218 (commented):
FYI

trunk/src/app/srs_app_gb28181.cpp
winlinvip: This is the wrapper object passed during construction, which means it is actually an object that no one will delete, so here it cannot be copied but directly referenced and deleted.

This is similar to how an RTC connection will delete itself after a timeout. It belongs to a situation where no one will manage it, so it can only manage itself. Therefore, the wrapper is used to manage the resource.

If this wrapper is managed by someone, then it needs to be copied.

`TRANS_BY_GPT3`

lipeng19811218 (approved):

trunk/src/app/srs_app_gb28181.cpp
winlinvip: This can be improved by copying everywhere.

`TRANS_BY_GPT3`

lipeng19811218 (approved):

PR #3352, opened 2023-01-01
Asan: Disable asan for CentOS and use statically link if possible.

### LINKED ISSUE #3347: Asan: Disable for CentOS7 because LD_PRELOAD issue.
**Description**

For CentOS7, we got error as bellow, see https://github.com/google/sanitizers/issues/796 for detail:

```
==48458==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```

1. SRS Version: 5.0.119

2. SRS Log:

```
==48458==ASan runtime does not come first in initial library list; you should either link 
runtime to your application or manually preload it with LD_PRELOAD.
```

3. SRS Config:

```
./objs/srs -c conf/srs.conf
```

**Replay**

CentOS 7.

Step 1: Link asan by shared library.

```
./configure --sanitizer=on --sanitizer-static=off && make
```

Step 2: Run SRS

```
./objs/srs -c conf/srs.conf

#==48458==ASan runtime does not come first in initial library list; you should either link 
# runtime to your application or manually preload it with LD_PRELOAD.
```

There is a workaround:

```
env LD_PRELOAD=/lib64/libasan.so.5 ./objs/srs -c conf/srs.conf
```

chen-guanghua:
disable asan for centos7, fix #3347

### DIFF
--- a/trunk/auto/options.sh
+++ b/trunk/auto/options.sh
@@ -68,7 +68,7 @@ SRS_LOG_LEVEL_V2=YES
 # Experts options.
 SRS_USE_SYS_SSL=NO # Use system ssl(-lssl) if required.
 SRS_VALGRIND=NO
-SRS_SANITIZER=YES
+SRS_SANITIZER=NO
 SRS_SANITIZER_STATIC=NO
 SRS_SANITIZER_LOG=NO
 SRS_BUILD_TAG= # Set the object files tag name.
@@ -501,6 +501,13 @@ function apply_auto_options() {
     # Enable FFmpeg fit for RTC to transcode audio from AAC to OPUS, if user enabled it.
     if [[ $SRS_RTC == YES && $SRS_FFMPEG_FIT == RESERVED ]]; then
         SRS_FFMPEG_FIT=YES
+    fi
+
+    # Enable asan, but disable for Centos
+    # @see https://github.com/ossrs/srs/issues/3347
+    if [[ $SRS_SANITIZER == NO && $OS_IS_CENTOS != YES ]]; then
+        echo "Enable asan by auto options."
+        SRS_SANITIZER=YES
     fi
 
     # If enable gperf, disable sanitizer.
--- a/trunk/configure
+++ b/trunk/configure
@@ -197,11 +197,6 @@ if [[ $SRS_STATIC == YES ]]; then
     SrsLinkOptions="${SrsLinkOptions} -static-libstdc++";
 fi
 
-# For asan(Google Address Sanitizer)
-if [[ $SRS_SANITIZER == YES && $OS_IS_X86_64 == YES && $SRS_SANITIZER_STATIC == YES ]]; then
-    SrsLinkOptions="${SrsLinkOptions} -static-libasan";
-fi
-
 # For coverage.
 if [[ $SRS_GCOV == YES ]]; then
     SrsLinkOptions="${SrsLinkOptions} ${SrsGcov}";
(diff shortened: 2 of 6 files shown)


trunk/auto/options.sh
```diff
@@ -144,6 +144,11 @@ function apply_system_options() {
     if [[ $OS_IS_LINUX == YES || $OS_IS_CYGWIN == YES ]]; then
         SRS_JOBS=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || echo 1)
     fi
+
+    # Set centos version
+    if [[ $OS_IS_CENTOS == YES ]]; then
```
winlinvip: No need to check the version, let's disable Asan by default for all CentOS. But simply turning it off is not enough, we still need to support users to enable it themselves.

First, mark SRS_SANITIZER as RESERVED, and then set the initial value based on different systems.

Additionally, by default, enable asan static linking if possible.

`TRANS_BY_GPT3`

trunk/configure
```diff
@@ -197,11 +197,6 @@ if [[ $SRS_STATIC == YES ]]; then
     SrsLinkOptions="${SrsLinkOptions} -static-libstdc++";
 fi
 
-# For asan(Google Address Sanitizer)
-if [[ $SRS_SANITIZER == YES && $OS_IS_X86_64 == YES && $SRS_SANITIZER_STATIC == YES ]]; then
-    SrsLinkOptions="${SrsLinkOptions} -static-libasan";
```
winlinvip, line 202: Why remove this link option?

`TRANS_BY_GPT3`

trunk/auto/options.sh
```diff
@@ -512,6 +512,13 @@ function apply_auto_options() {
         SRS_SANITIZER=NO
     fi
 
+    # For centos, default disable asan
+    # @see https://github.com/ossrs/srs/issues/3347
+    if [[ $SRS_SANITIZER == RESERVED && $OS_IS_CENTOS != YES && $OS_IS_X86_64 == YES ]]; then
+        echo "Disable asan for Centos"
+        SRS_SANITIZER=YES
```
winlinvip: I see the messages and comments, all of them are about disabling ASAN, why is it YES here?

Also, should we stop checking `OS_IS_X86_64`?

`TRANS_BY_GPT3`

trunk/configure
chen-guanghua: Repeated at line 217

`TRANS_BY_GPT3`

trunk/auto/options.sh
chen-guanghua: By default, it is RESERVED. If it is not CentOS, set it as SRS_SANITIZER=YES.

`TRANS_BY_GPT3`

trunk/auto/options.sh
```diff
@@ -506,6 +506,13 @@ function apply_auto_options() {
         SRS_FFMPEG_FIT=YES
     fi
 
+    # Enable asan, but disable for Centos
+    # @see https://github.com/ossrs/srs/issues/3347
+    if [[ $SRS_SANITIZER == RESERVED && $OS_IS_CENTOS != YES ]]; then
+        echo "Enable asan by auto options."
+        SRS_SANITIZER=YES
```
winlinvip, line 510: If it is CentOS, then `SRS_SANITIZER=RESERVED` is actually not very good.

It should be explicitly set to NO.

`TRANS_BY_GPT3`

winlinvip (approved):

PR #3363, opened 2023-01-06
HTTP: Add CORS Header for private network access

toss156:
Chrome will start sending a CORS preflight request ahead of any [private network request](https://developer.chrome.com/blog/private-network-access-preflight/#what-is-private-network-access-pna) for a subresource, which asks for explicit permission from the target server. This preflight request will carry a new header, `Access-Control-Request-Private-Network: true`, and the response to it must carry a corresponding header, `Access-Control-Allow-Private-Network: true`.

The aim is to protect users from [cross-site request forgery (CSRF) attacks](https://portswigger.net/web-security/csrf) targeting routers and other devices on private networks. These [attacks have affected hundreds of thousands of users](https://securityaffairs.co/wordpress/22743/cyber-crime/soho-pharming-attack.html), allowing attackers to redirect them to malicious servers.

Reference content: 
https://developer.chrome.com/blog/private-network-access-preflight/
#2787 RTC: Ensure private network requests are made from secure contexts

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -8,6 +8,7 @@ The changelog for SRS.
 
 ## SRS 6.0 Changelog
 
+* v6.0, 2023-01-06, Merge [#3363](https://github.com/ossrs/srs/issues/3363): HTTP: Add CORS Header for private network access. v6.0.13
 * v6.0, 2023-01-04, Merge [#3362](https://github.com/ossrs/srs/issues/3362): SRT: Upgrade libsrt from 1.4.1 to 1.5.1. v6.0.12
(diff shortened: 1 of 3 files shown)


winlinvip (approved):

kirakiseki:
Hi! I'm currently facing the same CORS problem, after searching for documents, I think the header added in this PR should be set to "Access-Control-**Allow**-Private-Network" instead of "Access-Control-**Request**-Private-Network" which is used in preflight requests.

You can also refer to the same documents as those in the description of this PR.

If there are still any other problems, plz feel free to reach me out! Thanks:)

darik07 (commented):
>

PR #3380, opened 2023-01-12
Console: Not needed, just a number is enough for EN.

simon1tan1:
Not needed, just a number is enough


duiniuluantanqin:
你代码没更新吧？

duiniuluantanqin (approved):

winlinvip (approved):

PR #3382, opened 2023-01-14
Rewrite research/api-server code by Go, remove Python.

panda1986:

1. rewrite python api-server to golang.


winlinvip:
Please simplify:


1. Use only one `main.go` file to avoid multiple go files, which makes searching difficult.
2. Add a `go.mod` file by initializing with `go mod init api-server`.
3. Remove `proxy.go` as it is rarely used for CDN preheating HLS.
4. Remove `server.go` as the server list is now managed by heartbeat. It is no longer needed for the API and console.
5. Remove the `src` directory and place the files directly under `api-server`.

`TRANS_BY_GPT3`

trunk/research/api-server/server.go
```diff
@@ -0,0 +1,841 @@
...
+}
+
+func (v *SrsForwardRequest) IsOnForward() bool {
+	return v.Action == "on_forward"
+}
+
+func main() {
+	var port int
+	var ffmpegPath string
+	flag.IntVar(&port, "p", 8085, "use -p to specify listen port, default is 8085")
```
winlinvip: The previous way to start Python was:

```
./api-server.py 8085
````

The `-p 8085` option can be kept to ensure compatibility with the previous method.

`TRANS_BY_GPT3`

File: trunk/research/api-server/server.go [hunk 2]
```diff
@@ -0,0 +1,841 @@
...
+				go func() {
+					if err := sj.cmd.Wait(); err != nil {
+						log.Println(fmt.Sprintf("snapshot %v cmd wait failed, err is %v", streamTag, err))
+					} else { // choose the best quality image
+						bestFileSize := int64(0)
+						for i := 1; i <= vframes; i++ {
+							pic := path.Join(StaticDir, sj.App, fmt.Sprintf("%v-%03d.png", sj.Stream, i))
+							fi, err := os.Stat(pic)
+							if err != nil {
+								log.Println(fmt.Sprintf("stat pic:%v failed, err is %v", pic, err))
```
winlinvip: There is a logic with 10 levels of indentation here, which is too complicated. We should consider how to simplify it.

`TRANS_BY_GPT3`

trunk/research/api-server/server.go
```diff
@@ -0,0 +1,843 @@
...
+	return fmt.Sprintf("%v/%v/%v", v.Vhost, v.App, v.Stream)
+}
+
+func (v *SnapshotJob) Abort() {
+	v.cancelFunc()
+	log.Println(fmt.Sprintf("cancel snapshot job %v", v.Tag()))
+}
+
+/*
+./objs/ffmpeg/bin/ffmpeg -i rtmp://127.0.0.1/live?vhost=__defaultVhost__/panda -vf fps=1 -vcodec png -f image2 -an -y -vframes 5 -y static-dir/live/panda-%03d.png
```
chundonglinlin: Suggested streaming address for this location: `rtmp://127.0.0.1/live?vhost=__defaultVhost__/livestream`

`TRANS_BY_GPT3`

trunk/research/api-server/server.go
```diff
@@ -0,0 +1,843 @@
...
+		v.snapshots.Delete(streamUrl)
+		log.Println(fmt.Sprintf("set stream:%v to destroy, update abort", sm.Stream))
+	} else {
+		log.Println(fmt.Sprintf("cannot find stream:%v in snapshot worker", streamUrl))
+	}
+	return
+}
+
+/*
+handle the forward requests: dynamic forward url.
```
chundonglinlin, line 487: Let's standardize the comments, the comments above are not aligned with the margin.

`TRANS_BY_GPT3`

trunk/research/api-server/server.go
winlinvip: It is better to use the new URL format, with vhost at the end.

```
rtmp://127.0.0.1/live/livestream?vhost=__defaultVhost__
```

Alternatively, removing `__defaultVhost__` is also possible.

`TRANS_BY_GPT3`

chen-guanghua (approved):

winlinvip (approved):

PR #3386, opened 2023-01-17
SRT: fix crash when srt_to_rtmp off

xiaozhihong:
When srt_to_rtmp disable, bridge_ is null ptr, fix crash.

### DIFF
--- a/trunk/src/app/srs_app_srt_source.cpp
+++ b/trunk/src/app/srs_app_srt_source.cpp
@@ -768,7 +768,7 @@ srs_error_t SrsSrtSource::on_publish()
         return srs_error_wrap(err, "source id change");
     }
 
-    if ((err = bridge_->on_publish()) != srs_success) {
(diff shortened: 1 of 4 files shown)


winlinvip (approved):

trunk/src/app/srs_app_srt_source.cpp
```diff
@@ -787,7 +787,9 @@ void SrsSrtSource::on_unpublish()
 
     can_publish_ = true;
 
-    bridge_->on_unpublish();
+    if (bridge_) {
+        bridge_->on_unpublish();
+    }
     srs_freep(bridge_);
```
chundonglinlin, line 793: Can `srs_freep(bridge_);` be placed inside the `if` condition? However, `srs_freep` also checks for null.

`TRANS_BY_GPT3`

chundonglinlin (approved):

PR #3392, opened 2023-01-28
Support composited bridges for 1:N protocols converting. v6.0.41

winlinvip:
In SRS 5.0, the bridge is attached to source, used to covert from one protocol to another one. For example, when configured RTMP to RTC:

```nginx
vhost __defaultVhost__ {
    rtc { enabled on;  rtmp_to_rtc on; }
}
```

When publish stream to SRS by RTMP client, a bridge will be created and attached to source:

```cpp
srs_error_t SrsRtmpConn::acquire_publish(SrsLiveSource* source) {
    SrsRtmpToRtcBridge* bridge = new SrsRtmpToRtcBridge(rtc);
    source->set_bridge(bridge);
}
```

Then all RTMP packets will be fed to bridge and allowd to be devliered by WebRTC. Similarly, a `SrsSrtToRtmpBridge` is used to convert SRT to RTMP stream, and a `SrsRtcToRtmpBridge` is for coverting RTC to RTMP.

It works well for `SRT => RTMP`, `RTMP => RTC`, `RTC => RTMP`. But, what happends if need `SRT => RTC`? Well, it actually combiles two bridges, first `SRT => RTMP` then `RTMP => RTC`, this is complex and an extra layer is introduced, this PR is to refine the bridge to allow directly and simultaneously coverting, such as:

```text
SRT => RTMP
SRT => RTC
```

In this PR, we need to:

- [x] Rename bridge from `YyyFromXxxBridge` to `XxxToYyyBridge`, which is better name, for example, rename `SrsRtcFromRtmpBridge` to `SrsRtmpToRtcBridge`
- [x] Use frame `SrsSharedPtrMessage` as message between all bridges.
- [x] Assemble packet, such as RTP for RTC, or TS for SRT, to frame before passed to bridge.
- [x] Support composite bridges to covert SRT to RTMP and RTC.
- [x] Should pass-by audio when converting SRT to RTMP, because both are AAC. https://github.com/ossrs/srs/pull/3392#pullrequestreview-1306493599
- [x] When bridge is disabled, each source is able to be published by different clients. https://github.com/ossrs/srs/pull/3392#issuecomment-1443835994
- [x] #3491


trunk/src/app/srs_app_srt_source.cpp
```diff
@@ -985,9 +1015,12 @@ void SrsSrtSource::on_unpublish()
     can_publish_ = true;
 
     if (bridge_) {
+        frame_builder_->on_unpublish();
```
chundonglinlin, line 1018: If `ifndef SRS_FFMPEG_FIT` is not defined, could `frame_builder_` be empty?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_stream_bridge.cpp
```diff
@@ -0,0 +1,236 @@
...
+}
+
+srs_error_t SrsCompositeBridge::on_frame(SrsSharedPtrMessage* frame)
+{
+    srs_error_t err = srs_success;
+
+    for (vector<ISrsStreamBridge*>::iterator it = bridges_.begin(); it != bridges_.end(); ++it) {
+        ISrsStreamBridge* bridge = *it;
+        if ((err = bridge->on_frame(frame)) != srs_success) {
+            return err;
```
xiaozhihong, line 224: Do you want to print a warn log and continue here?

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -1197,7 +1197,7 @@ srs_error_t SrsRtcPublishStream::initialize(SrsRequest* r, SrsRtcSourceDescripti
         // especially for stream merging.
         rtmp->set_cache(false);
 
-        SrsRtmpFromRtcBridge *bridge = new SrsRtmpFromRtcBridge(rtmp);
+        SrsFrameToRtmpBridge* bridge = new SrsFrameToRtmpBridge(rtmp);
```
xiaozhihong: Should we change it to here?
```
        SrsCompositeBridge* bridge = new SrsCompositeBridge();
        bridge->append(new SrsFrameToRtmpBridge(rtmp));
```
It should be consistent with the logic in srs_app_srt_conn.cpp.

`TRANS_BY_GPT3`

xiaozhihong:
The current structure may have one issue. If all bridges are closed, any protocol can push the same named stream. For example, rtmp can push live/livestream, srt can push live/livestream, rtc can push live/livestream. However, these three streams have different content.

In the future, it may be possible to create an independent source that is not tied to any specific protocol, which can be used to manage streams.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_srt_source.cpp
winlinvip: Frame builder only relies on FFmpeg when transcoding is required. RTC to RTMP relies on it, while SRT to RTMP does not. Therefore, it is necessary to consider the case of passing through AAC here.

`TRANS_BY_GPT3`

winlinvip:
> The current structure may have one problem. If all bridges are closed, any protocol can push the same named stream. For example, rtmp pushes live/livestream, srt pushes live/livestream, rtc pushes live/livestream, but these three streams have different contents.
> 
> In the future, it may be possible to create a protocol-independent source to manage streams.

> Confirmed, currently it checks if the stream exists regardless of whether the bridge is enabled or not, but there is a non-atomic operation issue. Created a separate issue to address it. #3491

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
winlinvip: It can be improved. Also, I will change the bridge from RTMP to RTC to a composite bridge.

```
        SrsCompositeBridge* bridge = new SrsCompositeBridge();
        bridge->append(new SrsFrameToRtcBridge(rtc));
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_stream_bridge.cpp
winlinvip: This involves error handling, such as SRT to RTMP and RTC, with two strategies for error handling:

1. After an error occurs, it returns directly, causing SRT to re-push. Currently, this is the strategy being used.
2. After an error occurs, it is ignored and processing continues, SRT stream remains normal, but some RTMP or RTC may be abnormal.

If the target is RTMP or RTC and they must be available, then strategy 1 is effective. If SRT is primarily used and RTMP and RTC can be ignored, then strategy 2 is effective.

This is similar to converting RTMP to HLS and FLV. How should errors in converting to HLS be handled? Previously, there was a configuration to choose different strategies, but it was not good enough.

`TRANS_BY_GPT3`

winlinvip: I will create a separate issue to track this problem. Regarding error handling for protocol conversion, we may need a good mechanism. https://github.com/ossrs/srs/issues/3437

`TRANS_BY_GPT3`

xiaozhihong (approved):
ok

trunk/src/app/srs_app_srt_source.cpp
winlinvip: I confirmed it:

1. frame_builder_ does not rely on FFmpeg, it only composes frames.
2. Audio transcoding is only available in SrsRtcRtpBuilder and SrsRtcFrameBuilder, it will not be present between RTMP and SRT.

So both of them are fine.

`TRANS_BY_GPT3`

chundonglinlin (approved):

PR #3394, opened 2023-01-29
UTest: Fix crash for stack overflow, allocate object on heap.

winlinvip:


chundonglinlin (approved):

duiniuluantanqin (approved):

trunk/src/kernel/srs_kernel_codec.hpp
```diff
@@ -490,6 +490,22 @@ struct SrsHevcHvccNalu {
...
+    // @see F.7.3.2.1: vps_video_parameter_set_id is u(4).
+    // @doc ITU-T-H.265-2021.pdf, page 464.
+    SrsHevcMax_VPS_COUNT = 16,
+    // @see 7.4.3.2.1: sps_seq_parameter_set_id is in [0, 15].
+    // @doc ITU-T-H.265-2021.pdf, page 95.
+    SrsHevcMax_SPS_COUNT = 16,
+    // @see 7.4.3.3.1: pps_pic_parameter_set_id is in [0, 63].
+    // @doc ITU-T-H.265-2021.pdf, page 102.
+    SrsHevcMax_PPS_COUNT = 64,
+};
```
winlinvip: These are not a type of data, so it is not suitable to make them an enum. They should be made as const.

`TRANS_BY_GPT3`

PR #3400, opened 2023-02-02
Config: Error when both HLS and HTTP-TS enabled.

### LINKED ISSUE #3391: Should return error when both HLS and HTTP-TS are enabled
If enable HLS and HTTP-TS, SRS can't identify the protocol(HLS or HTTP-TS) when viewer accesses the url `http://x/y/z.ts`, because it might be HTTP-TS or HLS file.

So, we should detect the conflict for HLS and HTTP-TS, which should return error when enable both.

See https://github.com/ossrs/srs/pull/3389#issuecomment-1407263428

chundonglinlin:
If enable HLS and HTTP-TS, SRS can't identify the protocol(HLS or HTTP-TS). So when SRS start, we detect the conflict for HLS and HTTP-TS.
### Test1
```config
vhost __defaultVhost__ {
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].ts;
    }
    hls {
        enabled         on;
        hls_fragment    10;
        hls_window      60;
    }
}
```
### Test2
```config
vhost __defaultVhost__ {
    http_remux {
        enabled     off;
        mount       [vhost]/[app]/[stream].ts;
    }
    hls {
        enabled         on;
        hls_fragment    10;
        hls_window      60;
    }
}
```
### Test3
```config
vhost __defaultVhost__ {
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].ts;
    }
    hls {
        enabled         off;
        hls_fragment    10;
        hls_window      60;
    }
}
```
### Test4
```config
vhost __defaultVhost__ {
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].ts;
    }
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].flv;
    }
}
```
### Test5
```config
vhost __defaultVhost__ {
    http_remux {
        enabled     off;
        mount       [vhost]/[app]/[stream].ts;
    }
    http_remux {
        enabled     on;
        mount       [vhost]/[app]/[stream].flv;
    }
}
```

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -2503,9 +2503,6 @@ srs_error_t SrsConfig::check_normal_config()
     for (int n = 0; n < (int)vhosts.size(); n++) {
         SrsConfDirective* vhost = vhosts[n];
 
-        bool hls_enabled = false;
-        bool http_remux_ts = false;
-        int http_remux_cnt = 0;
         for (int i = 0; vhost && i < (int)vhost->directives.size(); i++) {
             SrsConfDirective* conf = vhost->at(i);
             string n = conf->name;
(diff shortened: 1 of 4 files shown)


xiaozhihong (approved):

trunk/src/app/srs_app_config.cpp
```diff
@@ -2673,6 +2692,16 @@ srs_error_t SrsConfig::check_normal_config()
                 }
             }
         }
+
+        // check valid http-remux count
+        if (http_remux_cnt > 1) {
```
winlinvip, line 2720: The logic before is all about checking if the key in the configuration is correct.

However, this check for conflicts between HTTPS, TS, and HLS is actually a completely different logic.

And these data are scattered in different places, which will be a big problem for subsequent understanding.

A better approach is to have a separate logic just to check if there is a conflict between this HTTP TS and HLS in the vhost.

This way, maintenance will be very easy.

`TRANS_BY_GPT3`

winlinvip (approved):

PR #3409, opened 2023-02-10
SRT: Reduce latency to 200ms of srt2rtc.conf

xiaozhihong:
peerlatency=0 can reduce 300ms of latency of srt to rtc.

### DIFF
--- a/trunk/conf/srt2rtc.conf
+++ b/trunk/conf/srt2rtc.conf
@@ -20,8 +20,8 @@ srt_server {
     listen 10080;
     maxbw 1000000000;
     connect_timeout 4000;
-    peerlatency 300;
-    recvlatency 300;
+    peerlatency 0;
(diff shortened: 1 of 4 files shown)


chundonglinlin (approved):

chundonglinlin:
When the latency is set to 300, the delay is above 500ms.
```
srt_server {
    enabled on;
    listen 10080;
    maxbw 1000000000;
    connect_timeout 4000;
    peerlatency 300;
    recvlatency 300;
}
```

When the latency is set to 0, the delay is above 200ms.
```
srt_server {
    enabled on;
    listen 10080;
    maxbw 1000000000;
    connect_timeout 4000;
    peerlatency 0;
    recvlatency 0;
}
```

`TRANS_BY_GPT3`

winlinvip (approved):

PR #3458, opened 2023-03-08
API: Support HTTP basic authentication for API. v6.0.4, v5.0.152

### LINKED ISSUE #1640: API: Secure HTTP API for Authentication. 1985端口没有任何安全认证
**描述(Description)**

> 目前采用程序实现了通过API踢掉超时的推流客户端，但是发现，任何人如果稍微熟悉SRS，都可以查看，甚至踢掉合法使用的客户。有什么方式可以避免这个问题？
  我看到前面有帖子，说可以把1985端口监听到127.0.0.1 ，然后通过另外的http代理过来，请问在docker部署环境下，怎么实现以上方法？

1. SRS版本(Version): `v3`
2. 部署方式 k8s

### LINKED ISSUE #2684: API: Exposed for WebRTC without any access control
With WebRTC we need to expose http_api to public.
Why there isn't any access control for http_api?
User can send HTTP to /rtc/v1/play/ - which is OK.
But... there is also /api/v1/streams/ /api/v1/clients/ accessible for everyone in internet...

We need to setup reverse-proxy between client and SRS server?

duiniuluantanqin:
### The 'Basic' HTTP Authentication Scheme
Refer to [RFC7617](https://www.rfc-editor.org/rfc/rfc7617), currently only authentication for HTTP API has been implemented.
**Configuration**
```
http_api {
    enabled on;
    listen 8080;
    auth {
        # default: off
        enabled         on;
        username        admin;
        password        admin;
    }
}
```

**Testing**
```
curl http://admin:admin@ip:1985/api/v1/clients/
```

Alternatively, accessing it directly through the browser will prompt for entering a username and password.

**Modification Plan**
The original process is as follows:
```mermaid
graph LR
    A(SrsHttpConn) --> B[SrsHttpCorsMux] --> C[SrsHttpServeMux]
```
The new process adds `SrsHttpAuthMux` as follows:
```mermaid
graph LR
    A(SrsHttpConn) --> B[SrsHttpCorsMux] --> C[SrsHttpAuthMux] --> D[SrsHttpServeMux]
    style C fill: #09f, stroke: #333
```

**Note:**
1. Considering that `API is for operations and maintenance, while streaming media is for users, these two passwords may be different.`, authentication is only performed for `/api/v1/`, while other endpoints such as RTC signaling `/rtc/v1/` are not authenticated.
2. Of course, static pages will also not be authenticated, even if they share the same port with the `API`.

---------

`TRANS_BY_GPT3`


winlinvip:
After enabling authentication, will the Console prompt this popup? Check if there is a popup in the Console, as many people have reported that there is no authentication in the Console. Therefore, it would be good to have a popup, not having one might not be ideal.

Does this also apply to the RTC API? The RTC API belongs to the streaming media part, so if RTC is enabled, should HTTP-FLV also be enabled for consistency? The API is for operations and maintenance, while streaming media is for users, so the passwords for these two may be different.

The HTTP server and API can reuse the same port, so we may need to check if it will function properly.

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
```diff
@@ -1303,3 +1303,54 @@ int srs_chunk_header_c3(int perfer_cid, uint32_t timestamp, char* cache, int nb_
     return (int)(p - cache);
 }
 
+
+inline bool is_base64(unsigned char c)
+{
+    return (std::isalnum(c) || (c == '+') || (c == '/'));
+}
+
+std::string base64_decode(const std::string &encoded_string)
```
xiaozhihong: Functions in the kernel, it is recommended to add unit tests for all of them.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -920,11 +919,84 @@ srs_error_t SrsHttpCorsMux::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessag
...
+
+    std::string base64 = base64_decode(token);
+
+    // should be 'username:password'
+    std::vector<std::string> user_pwd = srs_string_split(base64, ":");
+    if (user_pwd.size() != 2) {
+        return srs_error_new(SRS_CONSTS_HTTP_Unauthorized, "invalid token");
+    }
+
+    srs_warn("get auth %s/%s, expect %s/%s", user_pwd[0].c_str(), user_pwd[1].c_str(), username.c_str(), password.c_str());
```
xiaozhihong: Should this log be placed inside the if statement below?

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.hpp
```diff
@@ -482,22 +482,43 @@ class SrsHttpServeMux : public ISrsHttpServeMux
...
+    virtual srs_error_t serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessage* r);
+};
+
+// The filter http mux, directly serve the http AUTH requests,
+// while proxy to the worker mux for services.
+// @see: https://www.rfc-editor.org/rfc/rfc7617
+class SrsHttpAuthMux : public ISrsHttpHandler
+{
+private:
+    bool enabled;
```
xiaozhihong: After version 4.0, let's uniformly name class members using xxx_, with an underscore.

`TRANS_BY_GPT3`

duiniuluantanqin: ok

trunk/src/protocol/srs_protocol_http_stack.cpp
duiniuluantanqin: Yes

`TRANS_BY_GPT3`

trunk/src/kernel/srs_kernel_utility.cpp
duiniuluantanqin: ok

winlinvip:
Please add black box testing and use Go as the client to verify if the auth is correct.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -920,9 +919,87 @@ srs_error_t SrsHttpCorsMux::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessag
...
+        w->header()->set("WWW-Authenticate", SRS_HTTP_AUTH_PREFIX_BASIC);
+        return srs_error_new(SRS_CONSTS_HTTP_Unauthorized, "empty Authorization");
+    }
+
+    if (!srs_string_contains(auth, SRS_HTTP_AUTH_PREFIX_BASIC)) {
+        return srs_error_new(SRS_CONSTS_HTTP_Unauthorized, 
+            "invalid authorization header. Must start with %s", SRS_HTTP_AUTH_PREFIX_BASIC);
+    }
+
+    std::string token = srs_string_trim_start(auth, SRS_HTTP_AUTH_PREFIX_BASIC);
```
xiaozhihong: srs_string_trim_start(auth, SRS_HTTP_AUTH_PREFIX_BASIC)

The meaning of this function is to remove the characters 'B' 'a' 's' 'i' 'c' ' ' from auth, which will also remove some characters from base64.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -24,6 +24,8 @@ using namespace std;
 // @see ISrsHttpMessage._http_ts_send_buffer
 #define SRS_HTTP_TS_SEND_BUFFER_SIZE 4096
 
+#define SRS_HTTP_AUTH_PREFIX_BASIC "Basic "
```
xiaozhihong: Here, Basic has a space, which doesn't seem to fit the usual processing flow. Generally, when defining constant strings, they are without spaces, and the program handles cases with multiple spaces.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.cpp
duiniuluantanqin: Thank you very much for the reminder, it has been resolved.

`TRANS_BY_GPT3`

trunk/src/protocol/srs_protocol_http_stack.cpp
duiniuluantanqin: 规定`Basic`后面是有个空格的，这里应该是没有问题的。暂不处理。' translates to English as:
"The rule is that there should be a space after 'Basic', so there should be no problem here. Not handling it for now.

`TRANS_BY_GPT3`

xiaozhihong (approved):

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -920,9 +920,91 @@ srs_error_t SrsHttpCorsMux::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessag
...
+
+    std::string path = r->path();
+    // only for /api/
+    if (path.find("/api/") == std::string::npos) {
+        return srs_success;
+    }
+
+    std::string auth = r->header()->get("Authorization");
+    if (auth.empty()) {
+        w->header()->set("WWW-Authenticate", SRS_HTTP_AUTH_PREFIX_BASIC);
```
winlinvip:
```
#define SRS_HTTP_AUTH_PREFIX_BASIC "Basic "
```

Need a space?

`TRANS_BY_GPT3`

winlinvip (approved):

winlinvip:
I think it works for me now, please also remember to update the documentation for HTTP basic authentication at [HTTP-API](https://ossrs.net/lts/zh-cn/docs/v5/doc/http-api)

PR #3470, opened 2023-03-16
WebRTC: Support configure CANDIDATE by env

winlinvip:
In dockerfile, we can set the default RTC candidate to env:

```
ENV SRS_RTC_SERVER_CANDIDATE=\$CANDIDATE
CMD ["./objs/srs", "-e"]
```

When starts a docker container, user can setup the candidate by env:

```
docker run --rm -it --env CANDIDATE=1.2.3.4 ossrs/srs:5
```

We should parse the content of SRS_RTC_SERVER_CANDIDATE as env variable name and parse it again.

### DIFF
--- a/trunk/conf/realtime.conf
+++ b/trunk/conf/realtime.conf
@@ -7,7 +7,7 @@ max_connections     1000;
 daemon              off;
 srs_log_tank        console;
 vhost __defaultVhost__ {
-    tcp_nodelay     on
+    tcp_nodelay     on;
     min_latency     on;
     
     play {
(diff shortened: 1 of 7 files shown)


trunk/conf/realtime.conf
```diff
@@ -7,7 +7,7 @@ max_connections     1000;
 daemon              off;
 srs_log_tank        console;
 vhost __defaultVhost__ {
-    tcp_nodelay     on
```
chundonglinlin, line 10: `realtime.flv.conf` also has this problem

`TRANS_BY_GPT3`

chundonglinlin (approved):

mapengfei53 (approved):

Issue #3331, opened 2022-12-23
Remove unnecessary null pointer checks


elfring:
:thought_balloon: An extra null pointer check is not needed in macros like "srs_freep" and "srs_freepa".

`TRANS_BY_GPT3`

winlinvip:
Sounds good! Would you please file a PR?

yashwardhan-jyani:
Hey @winlinvip , i fixed the issue and submitted a PR on this!

winlinvip:
Fixed by https://github.com/ossrs/srs/pull/3477

PR #3477, opened 2023-03-22
Remove unneccessary NULL check in srs_freep. v5.0.150, v6.0.38

yashwardhan-jyani:
Removed the unnecessary NULL pointer checks in trunk/src/core/srs_core.hpp

[https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null](url)

This is my first Pull Request, So if anything's wrong please tell me

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -8,6 +8,7 @@ The changelog for SRS.
 
 ## SRS 6.0 Changelog
 
+* v6.0, 2023-03-25, Merge [#3477](https://github.com/ossrs/srs/pull/3477): Remove unneccessary NULL check in srs_freep. v6.0.37 (#3477)
 * v6.0, 2023-03-22, Merge [#3427](https://github.com/ossrs/srs/pull/3427): WHIP: Support DELETE resource for Larix Broadcaster. v6.0.36 (#3427)
 * v6.0, 2023-03-20, Merge [#3460](https://github.com/ossrs/srs/pull/3460): WebRTC: Support WHIP/WHEP players. v6.0.35 (#3460)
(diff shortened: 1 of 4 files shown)


elfring (changes requested):
The commit message can be improved.
* How do you think about to omit hyphens from the subject?
* Would you like to [refer to a previous suggestion in your change description](https://github.com/ossrs/srs/issues/3331 "Proposal from 2022-12-23")?

winlinvip:
Seems make sense to me, however I wonder if all compiler check the null pointer?

elfring:
>..., however I wonder if all compilers check the null pointer?

:thought_balloon: Would you like to support any compiler versions which would be standard-incompliant in this technical detail?

`TRANS_BY_GPT3`

xiaozhihong:
@yashwardhan-jyani 

Thanks for your PR.

Follow the reference, delete nullptr is safe.
https://cplusplus.com/reference/new/operator%20delete/
image

https://en.cppreference.com/w/cpp/memory/new/operator_delete
image


And I test it with simple code.
```cpp
#include <iostream>

int main()
{
    char* p = NULL;
    delete p;

    return 0;
}
```

It works well. 

Then I gdb the program and find the implement of delete in gcc.
image

image

Obvious, `operator delete` call std::free, std::free call libc free directly, and `man 3 free` tell us
image


But I still have a question, in other platforms which not use gcc, for example, MSVC of Windows, cause the same result? Can you test it? Thanks.

elfring:
> But I still have a question, ..., cause the same result?

How much do you care for standard compliance of applied compiler variants? :thinking:

`TRANS_BY_GPT3`

yashwardhan-jyani:
@xiaozhihong 

I tried this on MSVC compiler and it's working fine.

Screenshot 2023-03-24 145526
 


Ignore the errors, it's the intelliSense, i tried to get rid of them but can't seem to do so.
cx

xiaozhihong (approved):

winlinvip (approved):

Issue #3138, opened 2022-08-08
Cluster: K8s: Proxy to extend origin servers.


winlinvip:
Both cluster and proxy works for system load balancing, in short to serve a large set of connections or clients, but there are some differences.

> 中文: Cluster和Proxy都是为了解决系统的负载均衡问题，简单来说就是支持很多连接或客户端，但是它们从方案上看是有区别的。

Cluster works as an integrity, a set of servers of a cluster works like one server. So it supports a large number of streams, and each stream supports a lots of connections. For example, a cluster supports 100k streams and 100m viewers, or infinety system capacity.

> 中文：Cluster作为一个整体工作，很多服务器工作起来像是单个服务器一样。Cluster支持很多流，每个流支持很多播放连接。比如，Cluster支持10万路流，支持1亿播放观看，或者可以简单认为Cluster的系统容量是无限的。

Proxy works as a agent of SRS Origin server, to proxy streams to a special set of origin servers, and proxy to the same server for a specifical stream. Proxy doesn't extend system capacity for it proxy each UDP or TCP packet, but it helps media server for load balancing.

> 中文：Proxy是作为源站的代理，将流代理到一系列的源站，并且某个特定的流只会代理到同一个源站。Proxy并不会增加系统的容量，因为Proxy转发了每个UDP或TCP包，但是Proxy能帮助实现系统的负载均衡。从系统的整体来看，Proxy也可以实现系统能力的扩容。

Proxy is designed to make origin server stateless, to build a cluster from isolated stateless origin servers to a cluster, and can also be a part of a cluster.

> 中文：Proxy是为了解决源站的状态问题，将源站变成无状态服务。Proxy可以将多个无关的源站，组建成一个源站集群。因此Proxy当然也是集群的一个部分。

<a name="architecture"></a>
## [Architecture](#architecture)

Proxy works with SRS Origin servers, the stream flow works like this:

> 中文：Proxy一般和源站服务器一起工作，它的流图如下所示。

```
Client ---> LB --> Proxy Servers --> Origin Servers

OBS/FFmpeg --RTMP--> K8s(Service) --Proxy--> SRS(pod A)

Browsers --FLV/HLS/SRT--> K8s(Service) --Proxy--> SRS(pod A)

Browsers --+---HTTP-API--> K8s(Service) --Proxy--> SRS(pod A)
           +---WebRTC----> K8s(Service) --Proxy--> SRS(pod A)
```

[...]

PR #3482, opened 2023-03-25
Support: Set up configuration for the proxy cluster.

winlinvip:
This is proxy cluster framework for #3138 

Implements a RTMP stream proxy.


winlinvip:
Not ready.

PR #3509, opened 2023-04-12
BugFix: improve c++ std space import.

suzp1984:
The namespace `std` is imported to `srs_app_utility.cpp`, but the `string` declaration is not inconsistent, some part use `std::string`, while other part use `string`.  

And also the `::getenv()`, makes me confused at the beginning.

The commit deletes the `using namespace std;`,  I remember there is a coding convention that said to avoid importing the whole namespace in C++. What do you guys think about it?


winlinvip:
Thank you, we will take a look.

winlinvip:
In the cpp file, I think it's ok to use `std::string` or `string`.

PR #3534, opened 2023-05-07
Replace sprintf with snprintf to eliminate compile warnings. v6.0.45

chundonglinlin:
- Replaced all occurrences of sprintf with snprintf to address deprecation warnings
- Ensured proper buffer size is passed to snprintf to prevent potential buffer overflows
- Ran tests to confirm that the changes do not introduce any new issues or regressions

Remove compile warnings by using snprintf instead of sprintf, likely

```bash
warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  
Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
sprintf(error, "%d:%d: Unknown value", line_and_col);
```

### DIFF
--- a/trunk/src/app/srs_app_rtc_dtls.cpp
+++ b/trunk/src/app/srs_app_rtc_dtls.cpp
@@ -329,17 +329,20 @@ srs_error_t SrsDtlsCertificate::initialize()
 
     // Show DTLS fingerprint
     if (true) {
-        char fp[100] = {0};
-        char *p = fp;
         unsigned char md[EVP_MAX_MD_SIZE];
         unsigned int n = 0;
 
         // TODO: FIXME: Unused variable.
         /*int r = */X509_digest(dtls_cert, EVP_sha256(), md, &n);
 
+        char* fp = new char[3 * n];
+        SrsAutoFreeA(char, fp);
+        char *p = fp;
+
         for (unsigned int i = 0; i < n; i++, ++p) {
-            snprintf(p, 3, "%02X", md[i]);
-            p += 2;
+            int nb = snprintf(p, 3, "%02X", md[i]);
+            srs_assert(nb > 0 && nb < (3 * n - (p - fp)));
+            p += nb;
 
             if(i < (n-1)) {
                 *p = ':';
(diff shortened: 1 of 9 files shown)


trunk/src/app/srs_app_rtc_dtls.cpp
```diff
@@ -338,7 +338,7 @@ srs_error_t SrsDtlsCertificate::initialize()
         /*int r = */X509_digest(dtls_cert, EVP_sha256(), md, &n);
```
winlinvip, line 336: The value of n may exceed the capacity of the floating-point number, making it larger than its intended size.

trunk/src/app/srs_app_rtc_dtls.cpp
```diff
@@ -338,7 +338,7 @@ srs_error_t SrsDtlsCertificate::initialize()
...
 
         for (unsigned int i = 0; i < n; i++, ++p) {
-            sprintf(p, "%02X", md[i]);
+            snprintf(p, 3, "%02X", md[i]);
```
winlinvip: If the return value of snprintf is an error, then it should be checked and the function should be returned.

trunk/src/app/srs_app_tencentcloud.cpp
```diff
@@ -67,7 +67,7 @@ namespace tencentcloud_api_sign {
         HMAC_CTX_free(ctx);
 #endif
         for (unsigned i = 0; i != digest_len; ++i) {
-            sprintf(&c_hmacsha1[i*2], "%02x", (unsigned int)digest[i]);
+            snprintf(&c_hmacsha1[i*2], 3, "%02x", (unsigned int)digest[i]);
         }
```
chen-guanghua, line 71: 本义是转换成一个字节的十六进制数据，使用2是不是更加合理。
snprintf(&c_hmacsha1[i*2], 2, "%02x", (unsigned int)digest[i]);

chen-guanghua (approved):

PR #3551, opened 2023-05-25
H264: Fix H.264 ISOM reserved bit value. v5.0.161, v6.0.55

KazuoHagihara:
The MPEG-DASH receiver that we're using couldn't receive the stream that SRS ouputs.
The reason was that the value of the reserve bit of AVC sequence header was set to 0.

I fixed the reserved bit value to 1.

image


chundonglinlin:
This is a great fix, and we can use both the documentation standards(ISO_IEC_14496-15-AVC-format-2012, page 16.) and FFmpeg's processing method as reference([ff_isom_write_avcc](https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/avc.c#L225)).

chundonglinlin (approved):

PR #3592, opened 2023-06-20
Fix Permission Issue in depend.sh for OpenSSL Compilation. v5.0.160, v6.0.53

duiniuluantanqin:
When compiling in a WSL environment, if the code is on the C drive and logged in with a non-administrator account, permission issues may cause OpenSSL compilation errors. Add the `chmod` command to fix this.

When compiling in the WSL environment, if the code is on the C drive and logged in with a non-administrator account, there may be permission issues causing errors in the compilation of OpenSSL. Adding the `chmod` command will fix this.

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/auto/depends.sh
+++ b/trunk/auto/depends.sh
@@ -463,7 +463,7 @@ if [[ $SRS_SSL == YES && $SRS_USE_SYS_SSL != YES ]]; then
         ${OPENSSL_UNZIP} &&
         (
             cd ${SRS_OBJS}/${SRS_PLATFORM}/${OPENSSL_CANDIDATE} &&
-            chmod +x ./config && chmod +x ./Configure &&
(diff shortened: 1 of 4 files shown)


trunk/auto/depends.sh
```diff
@@ -463,6 +463,7 @@ if [[ $SRS_SSL == YES && $SRS_USE_SYS_SSL != YES ]]; then
         ${OPENSSL_UNZIP} &&
         (
             cd ${SRS_OBJS}/${SRS_PLATFORM}/${OPENSSL_CANDIDATE} &&
+            chmod +x ./config && chmod +x ./Configure &&
```
winlinvip:
```
chmod +x ./config ./Configure &&
```

xiaozhihong (approved):

PR #3599, opened 2023-06-22
SRT check SRT alive bug.

touch888:
srt serve error code=6007(SrtStreamBusy)(SRT stream alredy exists or busy) : srt stream /k=livegw/d7f58fe0ef busy
image


winlinvip:
Thanks a lot for submitting the patch. Version 5.0 is now stable. Apart from security vulnerabilities, version 4.0 is no longer being updated. SRS supports a smooth upgrade from 4.0 to 5.0. So, we recommend users of 4.0 to upgrade to 5.0.

Thanks again for giving feedback to the community 👍.

`TRANS_BY_GPT4`

PR #3606, opened 2023-06-28
support edge mode downstream video fps stat

b4158813:
- The current version of SRS does not support average frame rate statistics for downstream pulling of edge nodes in origin-edge mode. This PR adds this functionality for (downstream pulling under rtmp and http-flv), allowing it to be obtained in the API interface.
- Support downstream pulling video frame rate (average value if there are multiple clients) statistics in origin-edge mode.

---------

`TRANS_BY_GPT3`


winlinvip:

PR #3607, opened 2023-06-28
Bugfix: Resolved the problem of player not playing FLV in 5.0release.

ycy2019:
srs version:5.0-b1 
srs_player cannot play flv video. I found out the cause of the mistake (show_for_video_ok is not defined).
This piece of code was added in a previous commit
But in 5.0 this function is called "show_for_ok"

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/research/players/srs_player.html
+++ b/trunk/research/players/srs_player.html
@@ -242,7 +242,7 @@
                 return;
             }
 
-            show_for_video_ok();
+            show_for_ok();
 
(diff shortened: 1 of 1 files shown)


chundonglinlin:
Great work! We will review your code and merge it.

winlinvip:
Already fixed. c6eddc72e9a095d7e38a24673b0e023eb3c73ddd

Thanks! 👍

PR #3695, opened 2023-07-20
API: Fix HTTPS callback issue using SNI in TLS client handshake. v4.0.270, v5.0.168, v6.0.61

xiaozhihong:
Fix:Include the SNI in the TLS client handshake.

### DIFF
--- a/trunk/src/protocol/srs_protocol_http_client.cpp
+++ b/trunk/src/protocol/srs_protocol_http_client.cpp
@@ -56,7 +56,7 @@ SrsSslClient::~SrsSslClient()
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-srs_error_t SrsSslClient::handshake()
+srs_error_t SrsSslClient::handshake(const std::string& host)
 {
     srs_error_t err = srs_success;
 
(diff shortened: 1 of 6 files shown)


trunk/src/protocol/srs_protocol_http_client.cpp
```diff
@@ -56,7 +56,7 @@ SrsSslClient::~SrsSslClient()
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-srs_error_t SrsSslClient::handshake()
+srs_error_t SrsSslClient::handshake(const std::string& hostname)
```
chundonglinlin: Should the parameter name use `host` instead of `hostname`?

trunk/src/protocol/srs_protocol_http_client.cpp
```diff
@@ -88,6 +88,11 @@ srs_error_t SrsSslClient::handshake()
     // SSL setup active, as client role.
     SSL_set_connect_state(ssl);
     SSL_set_mode(ssl, SSL_MODE_ENABLE_PARTIAL_WRITE);
+    // If the server address is not in IP address format, set the hostname 
+    // in the Server Name Indication (SNI) field.
+    if (! srs_check_ip_addr_valid(hostname)) {
```
chundonglinlin: Here an extra space, best like `if (!srs_check_ip_addr_valid(host))`

chundonglinlin (approved):

PR #3748, opened 2023-07-30
Rtcp sr and rr for rtc play

runner365:
For WebRTC downlink, add:
* Sending RTCP SR.
* Receiving and processing RTCP RR to calculate RTT.
* Retransmission of NACK based on RTT.

The main functionality is to calculate RTT in the downlink, which prevents unnecessary retransmissions within a time period less than RTT, avoiding unnecessary retransmissions.

---------

`TRANS_BY_GPT3`

### DIFF
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -686,13 +686,13 @@ srs_error_t SrsRtcPlayStream::cycle()
     }
 }
 
-srs_error_t SrsRtcPlayStream::send_rtcp_sr() {
+srs_error_t SrsRtcPlayStream::send_rtcp_sr(int64_t now_ms) {
     srs_error_t err = srs_success;
     for(std::map<uint32_t, SrsRtcVideoSendTrack*>::iterator iter = video_tracks_.begin();
         iter != video_tracks_.end();
         iter++) {
         SrsRtcVideoSendTrack* track = iter->second;
-        if ((err = track->send_rtcp_sr()) != srs_success) {
+        if ((err = track->send_rtcp_sr(now_ms)) != srs_success) {
             return srs_error_wrap(err, "video send rtcp sr error track=%s", track->get_track_id().c_str());
         }
     }
@@ -701,7 +701,7 @@ srs_error_t SrsRtcPlayStream::send_rtcp_sr() {
         iter != audio_tracks_.end();
         iter++) {
         SrsRtcAudioSendTrack* track = iter->second;
-        if ((err = track->send_rtcp_sr()) != srs_success) {
+        if ((err = track->send_rtcp_sr(now_ms)) != srs_success) {
             return srs_error_wrap(err, "audiosend rtcp sr error track=%s", track->get_track_id().c_str());
         }
     }
@@ -842,8 +842,7 @@ srs_error_t SrsRtcPlayStream::on_rtcp_rr(SrsRtcpRR* rtcp, int64_t now_ms)
             audio_iter != audio_tracks_.end();
             audio_iter++) {
             if(ssrc == audio_iter->second->track_desc_->ssrc_) {
-                audio_iter->second->handle_rtcp_rr(rb, now_ms);
-                return err;
+                return audio_iter->second->handle_rtcp_rr(rb, now_ms);
             }
         }
 
@@ -851,8 +850,7 @@ srs_error_t SrsRtcPlayStream::on_rtcp_rr(SrsRtcpRR* rtcp, int64_t now_ms)
             video_iter != video_tracks_.end();
             video_iter++) {
             if(ssrc == video_iter->second->track_desc_->ssrc_) {
-                video_iter->second->handle_rtcp_rr(rb, now_ms);
-                return err;
+                return video_iter->second->handle_rtcp_rr(rb, now_ms);
             }
         }
         srs_warn("rtcp rr find to find track by ssrc:%u", ssrc);
@@ -998,7 +996,8 @@ srs_error_t SrsRtcPlayRtcpTimer::on_timer(srs_utime_t interval)
         return err;
     }
 
-    if ((err = p_->send_rtcp_sr()) != srs_success) {
+    int64_t now_ms = srs_update_system_time();
+    if ((err = p_->send_rtcp_sr(now_ms)) != srs_success) {
         srs_warn("RR err %s", srs_error_desc(err).c_str());
         srs_freep(err);
     }
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -1364,18 +1364,14 @@ srs_error_t SrsRtcPublishStream::send_rtcp_rr()
 {
     srs_error_t err = srs_success;
 
-    for (std::vector<SrsRtcVideoRecvTrack*>::iterator iter = video_tracks_.begin();
-         iter != video_tracks_.end();
-         iter++) {
+    for (std::vector<SrsRtcVideoRecvTrack*>::iterator iter = video_tracks_.begin(); iter != video_tracks_.end(); iter++) {
         SrsRtcVideoRecvTrack* track = *iter;
         if ((err = track->send_rtcp_rr()) != srs_success) {
             return srs_error_wrap(err, "send rtcp rr error, videotrack=%s", track->get_track_id().c_str());
         }
     }
 
-    for (std::vector<SrsRtcAudioRecvTrack*>::iterator iter = audio_tracks_.begin();
-         iter != audio_tracks_.end();
-         iter++) {
+    for (std::vector<SrsRtcAudioRecvTrack*>::iterator iter = audio_tracks_.begin(); iter != audio_tracks_.end(); iter++) {
         SrsRtcAudioRecvTrack* track = *iter;
         if ((err = track->send_rtcp_rr()) != srs_success) {
             return srs_error_wrap(err, "send rtcp rr error: audiotrack=%s", track->get_track_id().c_str());
--- a/trunk/src/app/srs_app_rtc_conn.cpp
+++ b/trunk/src/app/srs_app_rtc_conn.cpp
@@ -438,6 +438,8 @@ SrsRtcPlayStream::SrsRtcPlayStream(SrsRtcConnection* s, const SrsContextId& cid)
 
     cache_ssrc0_ = cache_ssrc1_ = cache_ssrc2_ = 0;
     cache_track0_ = cache_track1_ = cache_track2_ = NULL;
+    
+    timer_rtcp_ = new SrsRtcPlayRtcpTimer(this);
 }
 
 SrsRtcPlayStream::~SrsRtcPlayStream()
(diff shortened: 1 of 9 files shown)


trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -448,6 +450,9 @@ SrsRtcPlayStream::~SrsRtcPlayStream()
 
     _srs_config->unsubscribe(this);
 
+    if (timer_rtcp_) {
+        srs_freep(timer_rtcp_);
```
winlinvip: You can simply release the object directly, srs_freep will check if it is empty.

```cpp
srs_freep(timer_rtcp_);
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -681,6 +686,29 @@ srs_error_t SrsRtcPlayStream::cycle()
     }
 }
 
+srs_error_t SrsRtcPlayStream::send_rtcp_sr(int64_t now_ms) {
```
winlinvip: Please keep the coding style consistent, and the opening brace of member functions should be on a new line.

```cpp
srs_error_t SrsRtcPlayStream::send_rtcp_sr(int64_t now_ms) 
{
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -681,6 +686,29 @@ srs_error_t SrsRtcPlayStream::cycle()
...
+    srs_error_t err = srs_success;
+    for(std::map<uint32_t, SrsRtcVideoSendTrack*>::iterator iter = video_tracks_.begin();
+        iter != video_tracks_.end();
+        iter++) {
```
winlinvip: These three lines can be written in one line, now the screen and editor are both very long, and can tolerate longer lines. Please change it to:

```cpp
    for(std::map<uint32_t, SrsRtcVideoSendTrack*>::iterator iter = video_tracks_.begin(); iter != video_tracks_.end(); iter++) {
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -928,6 +978,33 @@ srs_error_t SrsRtcPlayStream::do_request_keyframe(uint32_t ssrc, SrsContextId ci
...
+srs_error_t SrsRtcPlayRtcpTimer::on_timer(srs_utime_t interval)
+{
+    srs_error_t err = srs_success;
+
+    if (!p_->is_started) {
+        return err;
+    }
+
+    int64_t now_ms = srs_update_system_time();
+    if ((err = p_->send_rtcp_sr(now_ms)) != srs_success) {
```
winlinvip, line 989: For time units, we use `srs_utime_t` uniformly instead of `int64_t`. Actually, here `now_ms` is not in milliseconds but in microseconds, which can be misleading. It needs to be changed to: 

```cpp
srs_utime_t now = srs_update_system_time();
if ((err = p_->send_rtcp_sr(now)) != srs_success) {

srs_error_t send_rtcp_sr(srs_utime_t now);
```

If you need to convert time to milliseconds, you can use the macro definition `srsu2ms(now)`, and if you need to convert it to seconds, use `srsu2s(now)`.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2710,6 +2736,73 @@ void SrsRtcSendTrack::rebuild_packet(SrsRtpPacket* pkt)
...
+    //srs_trace("send rtcp sr ssrc:%u, current ms:%ld, last ms:%ld, diff:%ld", ssrc, now_ms, last_ms_, now_ms - last_ms_);
+    last_sr_ = current_sr;
+    last_ms_ = now_ms;
+    sr->set_ssrc(ssrc);
+    sr->set_ntp(last_sr_ntp_.ntp_);
+    sr->set_rtp_ts(video_rtp_ts);
+    sr->set_rtp_send_packets(send_count_);
+    sr->set_rtp_send_bytes(send_bytes_);
+   
+    char data[1500];
```
winlinvip: There should not be a magic number 1500, but instead use existing constants or macro definitions. You can refer to the serialization of other RTC objects, I remember there will be a buffer cache and a method to specify the corresponding length.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2710,6 +2736,73 @@ void SrsRtcSendTrack::rebuild_packet(SrsRtpPacket* pkt)
...
+    SrsBuffer buffer(data, sr->nb_bytes());
+    sr->encode(&buffer);
+    delete sr;
```
winlinvip: SrsAutoFree should be used instead of naked delete to automatically release memory upon creation, in order to avoid forgetting to delete sr and causing memory leaks.

```cpp
SrsRtcpSR* sr = new SrsRtcpSR();
SrsAutoFree(SrsRtcpSR, sr);
```

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2710,6 +2736,73 @@ void SrsRtcSendTrack::rebuild_packet(SrsRtpPacket* pkt)
...
+    session_->send_rtcp(buffer.data(), buffer.size());
+
+    return err;
+}
+
+void SrsRtcSendTrack::update_rtp_static(int64_t len, uint32_t rtp_ts) {
+    send_count_++;
+    send_bytes_ += len;
+    last_rtp_pkt_ts_ = rtp_ts;
+    last_rtp_ms_     = srs_update_system_time() / 1000;//ms
```
winlinvip, line 2779: Do not directly use int64_t and ms, instead use srs_utime_t and srsu2ms conversion to avoid unit errors and make it easier to understand.

`TRANS_BY_GPT3`

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -799,12 +828,33 @@ srs_error_t SrsRtcPlayStream::on_rtcp(SrsRtcpCommon* rtcp)
...
-    // TODO: FIXME: Implements it.
-
+        for(std::map<uint32_t, SrsRtcVideoSendTrack*>::iterator video_iter = video_tracks_.begin();
+            video_iter != video_tracks_.end();
+            video_iter++) {
+            if(ssrc == video_iter->second->track_desc_->ssrc_) {
+                return video_iter->second->handle_rtcp_rr(rb, now_ms);
+            }
+        }
+        srs_warn("rtcp rr find to find track by ssrc:%u", ssrc);
```
xiaozhihong, line 845: typo

trunk/src/app/srs_app_rtc_source.hpp
```diff
@@ -706,6 +714,19 @@ class SrsRtcSendTrack
     virtual srs_error_t on_rtp(SrsRtpPacket* pkt) = 0;
     virtual srs_error_t on_rtcp(SrsRtpPacket* pkt) = 0;
     virtual srs_error_t on_recv_nack(const std::vector<uint16_t>& lost_seqs);
+public:
+    srs_error_t send_rtcp_sr(int64_t now_ms);
+    void update_rtp_static(int64_t len, uint32_t rtp_ts);
+public:
+    srs_error_t handle_rtcp_rr(const SrsRtcpRB& rb, int64_t now_ms);
+protected:
```
xiaozhihong, line 722: Directly assigning values to member variables in a C++ class seems to be a syntax introduced in C++11.

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2666,6 +2674,24 @@ SrsRtpPacket* SrsRtcSendTrack::fetch_rtp_packet(uint16_t seq)
...
+        if ((pkt->resend_ms_ <= 0) || (pkt->resend_count_ == 0)) {
+            pkt->resend_ms_ = now_ms;
+            pkt->resend_count_++;
+        } else {
+            int64_t diff_t = now_ms - pkt->resend_ms_;
+            int64_t interval = (int64_t)avg_rtt_;
+            interval = (interval > 10) ? (interval - 10) : interval;//for resend interval Residual
+            if (diff_t < (int64_t)interval) {
+                return NULL;
+            }
```
xiaozhihong, line 2695: Should limit the count of retrans packet?

trunk/src/app/srs_app_rtc_conn.cpp
runner365: done

trunk/src/app/srs_app_rtc_conn.cpp
runner365: done

trunk/src/app/srs_app_rtc_source.cpp
runner365: done

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -928,6 +967,33 @@ srs_error_t SrsRtcPlayStream::do_request_keyframe(uint32_t ssrc, SrsContextId ci
...
+
+srs_error_t SrsRtcPlayRtcpTimer::on_timer(srs_utime_t interval)
+{
+    srs_error_t err = srs_success;
+
+    if (!p_->is_started) {
+        return err;
+    }
+
+    srs_utime_t now_ms = srs_update_system_time() / 1000;
```
winlinvip, line 988: You cannot divide by 1000 here, because this function returns `srs_utime_t`, and the variable cannot have a unit, because this type already has a unit.

It should be changed to:

```cpp
srs_utime_t now = srs_update_system_time();

if ((err = p_->send_rtcp_sr(now)) != srs_success) {
```

Your `send_rtcp_sr` should not be in milliseconds or seconds, it should be `srs_utime_t`, because it has a unit, which is a time unit. Don't convert it to milliseconds or seconds, only convert it to milliseconds using `srsu2ms` when it is finally used.

`TRANS_BY_GPT4`

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2710,6 +2743,70 @@ void SrsRtcSendTrack::rebuild_packet(SrsRtpPacket* pkt)
     srs_info("RTC: Correct %s seq=%u/%u, ts=%u/%u", track_desc_->type_.c_str(), seq, pkt->header.get_sequence(), ts, pkt->header.get_timestamp());
 }
 
+srs_error_t SrsRtcSendTrack::send_rtcp_sr(srs_utime_t now_ms) {
```
winlinvip, line 2746: This `srs_utime_t` is `utime`, which is in microseconds (us), so the variable should not carry the unit, or the variable should be `now_us`. However, this type is already in microseconds, so there is no need to carry the us unit. It should be changed to: 

```cpp
srs_error_t SrsRtcSendTrack::send_rtcp_sr(srs_utime_t now) {
    srs_utime_t diff = now_ms - last_rtp_ms_;
    srs_utime_t diff_ts = diff * track_desc_->media_->sample_ / 1000;
    srs_utime_t video_rtp_ts = last_rtp_pkt_ts_ + diff_xxx;
```

You should convert `track_desc_->media_->sample_` to microseconds (us), so it can be directly added or subtracted with `srs_utime_t`.

`TRANS_BY_GPT4`

winlinvip:

PR #3758, opened 2023-08-07
Refine docker detect mechenism. v5.0.190 v6.0.90

winlinvip:
When using Docker, logs are usually printed to console (stdout and stderr). However, since Docker detection occurs late, after log initialization, the default log output may be incorrect. In Docker, logs may still be written to a file instead of the console as expected.

Additionally, the Dockerfile has been improved with a new environment variable `SRS_IN_DOCKER=on` to clearly indicate a Docker environment. If automatic Docker detection fails, the configuration will be read, and this variable will correctly inform SRS that it's in a Docker environment.

Lastly, the default configuration values have been improved for Docker environments. By default, `SRS_LOG_TANK=console` and daemon mode is disabled.

---------

`TRANS_BY_GPT4`

---------


winlinvip:

`TRANS_BY_GPT4`

xiaozhihong (approved):
ok

PR #3759, opened 2023-08-08
Incorrect use of two "int i" instances.

lokenetwork:
Incorrect use of two "int i" instances; change one "int i" to "int j"


chundonglinlin:
The code below should be refactor?
https://github.com/ossrs/srs/blob/73dd8af4c99b528b2d9f9bb59339084efa30c0c3/trunk/src/app/srs_app_ng_exec.cpp#L140-L147

winlinvip (approved):

winlinvip:
image

Duplicate `int i` just works but not good coding practice, need to refactor.

PR #3768, opened 2023-08-15
Support include empty config file. v5.0.173 v6.0.68

winlinvip:
## Include Empty Config File

SRS supports including another configuration in the include package. When generating configurations, we can only generate the changed configurations, while the unchanged configurations are in the fixed files, for example:

```nginx
listen 1935;
include server.conf;
```

In `server.conf`, we can manage the changing configurations with the program:

```nginx
http_api { enabled on; }
```

However, during system initialization, we often create an empty `server.conf`, and the content is generated only after the program starts, so `server.conf` might be an empty file. This also makes it convenient to use a script to confirm the existence of this file:

```bash
touch server.conf
```

Currently, SRS does not support empty configurations and will report an error. This PR is to solve this problem, making it more convenient to use include.

`TRANS_BY_GPT4`

---------

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 [![](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fossrs%2Fsrs.svg?type=small)](https://app.fossa.com/projects/git%2Bgithub.com%2Fossrs%2Fsrs?ref=badge_small)
 [![](https://ossrs.net/wiki/images/srs-faq.svg)](https://ossrs.net/lts/zh-cn/faq)
 [![](https://badgen.net/badge/srs/stackoverflow/orange?icon=terminal)](https://stackoverflow.com/questions/tagged/simple-realtime-server)
-[![](https://opencollective.com/srs-server/tiers/badge.svg)](https://opencollective.com/srs-server/contribute)
+[![](https://opencollective.com/srs-server/tiers/badge.svg)](https://opencollective.com/srs-server)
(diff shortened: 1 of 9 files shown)


winlinvip:
Reload时，会经过几个阶段：
1. 解析新配置：Parse。
2. 转换配置：Transform。
3. 应用配置：Apply

之前是任何阶段出错都会直接退出，这样系统就完全依赖于检查配置：

```bash
./objs/srs -c conf/srs.conf -t
echo $?
#0
```

优化为：如果是在应用配置之前出错，是可以忽略的。如果在应用配置时出错，此时可能有部分配置已经生效，会有未知行为，所以SRS会直接退出。

新增一个HTTP API，查询reload的结果：

```json
curl http://localhost:1985/api/v1/raw?rpc=reload-fetch

{
  "code": 0,
  "data": {
    "err": 0,
    "msg": "Success",
    "state": 0,
    "rid": "0s6y0n9"
  }
}

{
  "code": 0,
  "data": {
    "err": 1023,
    "msg": "code=1023(ConfigInvalid) : parse file : parse buffer containers/conf/srs.release-local.conf : root parse : parse dir : parse include buffer containers/data/config/srs.vhost.conf : read token, line=0, state=0 : line 3: unexpected end of file, expecting ; or \"}\"",
    "state": 1,
    "rid": "0g4z471"
  }
}
```

这样可以知道系统最后一次reload是否成功。

duiniuluantanqin (approved):

PR #3774, opened 2023-08-21
DOC: Include a guide for the Debian and Ubuntu installer.

duiniuluantanqin:
add man page for srs-server


duiniuluantanqin:
move to another pr

PR #3779, opened 2023-08-22
Support HTTP-API for fetching reload result. v5.0.176 v6.0.71

winlinvip:
## Reload Error Ignore

During a Reload, several stages will be passed through:
1. Parsing new configurations: Parse.
2. Transforming configurations: Transform.
3. Applying configurations: Apply.

Previously, any error at any stage would result in a direct exit, making the system completely dependent on configuration checks:

```bash
./objs/srs -c conf/srs.conf -t
echo $?
#0
```

Optimized to: If an error occurs before applying the configuration, it can be ignored. If an error occurs during the application of the configuration, some of the configuration may have already taken effect, leading to unpredictable behavior, so SRS will exit directly.

## Reload Fetch API

Added a new HTTP API to query the result of the reload.

```nginx
http_api {
    enabled         on;
    raw_api {
        enabled on;
        allow_reload on;
    }
}
```

```bash
curl http://localhost:1985/api/v1/raw?rpc=reload-fetch
```

```json
{
  "code": 0,
  "data": {
    "err": 0,
    "msg": "Success",
    "state": 0,
    "rid": "0s6y0n9"
  }
}

{
  "code": 0,
  "data": {
    "err": 1023,
    "msg": "code=1023(ConfigInvalid) : parse file : parse buffer containers/conf/srs.release-local.conf : root parse : parse dir : parse include buffer containers/data/config/srs.vhost.conf : read token, line=0, state=0 : line 3: unexpected end of file, expecting ; or \"}\"",
    "state": 1,
    "rid": "0g4z471"
  }
}
```

This way, you can know if the last reload of the system was successful.

---------

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -1345,7 +1345,6 @@ srs_error_t SrsConfig::reload(SrsReloadState *pstate)
     if ((err = conf.parse_file(config_file.c_str())) != srs_success) {
         return srs_error_wrap(err, "parse file");
     }
-    *pstate = SrsReloadStateParsed;
     srs_info("config reloader parse file success.");
     
     // transform config to compatible with previous style of config.
@@ -1357,13 +1356,11 @@ srs_error_t SrsConfig::reload(SrsReloadState *pstate)
     if ((err = conf.check_config()) != srs_success) {
         return srs_error_wrap(err, "check config");
     }
-    *pstate = SrsReloadStateTransformed;
 
     *pstate = SrsReloadStateApplying;
     if ((err = reload_conf(&conf)) != srs_success) {
         return srs_error_wrap(err, "reload config");
     }
-    *pstate = SrsReloadStateApplied;
 
     *pstate = SrsReloadStateFinished;
     return err;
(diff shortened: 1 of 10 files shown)


trunk/src/app/srs_app_config.cpp
```diff
@@ -1328,30 +1328,39 @@ void SrsConfig::unsubscribe(ISrsReloadHandler* handler)
...
+    *pstate = SrsReloadStateTransforming;
     if ((err = srs_config_transform_vhost(conf.root)) != srs_success) {
         return srs_error_wrap(err, "transform config");
     }
     
     if ((err = conf.check_config()) != srs_success) {
         return srs_error_wrap(err, "check config");
     }
-    
+    *pstate = SrsReloadStateTransformed;
```
duiniuluantanqin: The two states, `SrsReloadStateTransformed` and `SrsReloadStateApplying`, can be combined into a single state?

trunk/src/app/srs_app_config.cpp
```diff
@@ -1328,30 +1328,39 @@ void SrsConfig::unsubscribe(ISrsReloadHandler* handler)
...
+
+    *pstate = SrsReloadStateApplying;
     if ((err = reload_conf(&conf)) != srs_success) {
         return srs_error_wrap(err, "reload config");
     }
-    
+    *pstate = SrsReloadStateApplied;
```
duiniuluantanqin: The same as above

trunk/src/app/srs_app_config.cpp
winlinvip: Fixed.

duiniuluantanqin (approved):

PR #3804, opened 2023-09-18
Support FFmpeg timecode, fix AMF0 parsing failed. v5.0.179 v6.0.77

### LINKED ISSUE #3803: Support FFmpeg timecode, fix AMF0 parsing failed.
See [Error submitting a packet to the muxer: Broken pipe, Error muxing a packet](https://trac.ffmpeg.org/ticket/10565)

Test using the file [timecode.mp4.zip](https://github.com/ossrs/srs/files/12708490/timecode.mp4.zip) or below:

https://github.com/ossrs/srs/assets/2777660/d50947a7-996a-4364-9593-3fb559b8e740

Use this FFmpeg command:

```bash
ffmpeg -re -i timecode.mp4 -map 0 -c copy -f flv rtmp://localhost/live/livestream
```

SRS will fail with this error:

```text
[2023-09-18 11:25:37.576][ERROR][79462][910ul5n0][4] serve error code=2003(Amf0Decode)(Decode AMF0 message failed) : service cycle : rtmp: stream service : rtmp: receive thread : handle publish message : rtmp: consume message : rtmp: decode message : decode message : decode command name : String invalid marker=0
thread [79462][910ul5n0]: do_cycle() [./src/app/srs_app_rtmp_conn.cpp:263][errno=4]
thread [79462][910ul5n0]: service_cycle() [./src/app/srs_app_rtmp_conn.cpp:457][errno=4]
thread [79462][910ul5n0]: do_publishing() [./src/app/srs_app_rtmp_conn.cpp:1030][errno=35]
thread [79462][910ul5n0]: consume() [./src/app/srs_app_recv_thread.cpp:380][errno=35]
thread [79462][910ul5n0]: handle_publish_message() [./src/app/srs_app_rtmp_conn.cpp:1185][errno=35]
thread [79462][910ul5n0]: process_publish_message() [./src/app/srs_app_rtmp_conn.cpp:1230][errno=35]
thread [79462][910ul5n0]: decode_message() [./src/protocol/srs_protocol_rtmp_stack.cpp:384][errno=35]
thread [79462][910ul5n0]: do_decode_message() [./src/protocol/srs_protocol_rtmp_stack.cpp:607][errno=35]
thread [79462][910ul5n0]: srs_amf0_read_string() [./src/protocol/srs_protocol_amf0.cpp:1513][errno=35](Interrupted system call)
```

Here's the pcap file:

[rtmp-amf0-invalid.pcapng.zip](https://github.com/ossrs/srs/files/12644504/rtmp-amf0-invalid.pcapng.zip)

winlinvip:
Please see https://github.com/ossrs/srs/issues/3803 for detail:

1. When using FFmpeg with the `-map 0` option, there may be a 4-byte timecode in the AMF0 Data.
2. SRS should be able to handle this packet without causing a parsing error, as it's generally expected to be an AMF0 string, not a 4-byte timecode.
3. Disregard the timecode since SRS doesn't utilize it.

See [Error submitting a packet to the muxer: Broken pipe, Error muxing a packet](https://trac.ffmpeg.org/ticket/10565)

---------

### DIFF
--- a/trunk/src/protocol/srs_protocol_rtmp_stack.cpp
+++ b/trunk/src/protocol/srs_protocol_rtmp_stack.cpp
@@ -14,7 +14,6 @@
 #include <srs_protocol_stream.hpp>
 #include <srs_protocol_utility.hpp>
 #include <srs_protocol_rtmp_handshake.hpp>
-#include <srs_app_utility.hpp>
 
 // for srs-librtmp, @see https://github.com/ossrs/srs/issues/213
 #ifndef _WIN32
--- a/trunk/src/protocol/srs_protocol_rtmp_stack.cpp
+++ b/trunk/src/protocol/srs_protocol_rtmp_stack.cpp
@@ -14,6 +14,7 @@
 #include <srs_protocol_stream.hpp>
 #include <srs_protocol_utility.hpp>
 #include <srs_protocol_rtmp_handshake.hpp>
+#include <srs_app_utility.hpp>
 
 // for srs-librtmp, @see https://github.com/ossrs/srs/issues/213
 #ifndef _WIN32
@@ -595,6 +596,12 @@ srs_error_t SrsProtocol::do_decode_message(SrsMessageHeader& header, SrsBuffer* 
     
     // decode specified packet type
     if (header.is_amf0_command() || header.is_amf3_command() || header.is_amf0_data() || header.is_amf3_data()) {
+        // Ignore FFmpeg timecode, see https://github.com/ossrs/srs/issues/3803
+        if (stream->left() == 4 && (uint8_t)*stream->head() == 0x00) {
+            srs_warn("Ignore FFmpeg timecode, data=[%s]", srs_string_dumps_hex(stream->head(), 4).c_str());
+            return err;
+        }
+
         // skip 1bytes to decode the amf3 command.
         if (header.is_amf3_command() && stream->require(1)) {
             stream->skip(1);
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -7,6 +7,7 @@ The changelog for SRS.
 <a name="v6-changes"></a>
 
 ## SRS 6.0 Changelog
+* v6.0, 2023-09-18, Merge [#3804](https://github.com/ossrs/srs/pull/3804): Support FFmpeg timecode, fix AMF0 parsing failed. v6.0.77 (#3804)
 * v6.0, 2023-09-18, Merge [#3722](https://github.com/ossrs/srs/pull/3722): Bugfix: HEVC SRT stream supports multiple PPS fields. v6.0.76 (#3722)
 * v6.0, 2023-09-08, Merge [#3597](https://github.com/ossrs/srs/pull/3597): Fix RBSP stream parsing bug, should drop 0x03. v6.0.75 (#3597)
 * v6.0, 2023-09-08, Merge [#3794](https://github.com/ossrs/srs/pull/3794): Support SRS Stack token for authentication. v6.0.74 (#3794)
@@ -88,6 +89,7 @@ The changelog for SRS.
 <a name="v5-changes"></a>
 
 ## SRS 5.0 Changelog
+* v5.0, 2023-09-18, Merge [#3804](https://github.com/ossrs/srs/pull/3804): Support FFmpeg timecode, fix AMF0 parsing failed. v5.0.179 (#3804)
 * v5.0, 2023-09-08, Merge [#3597](https://github.com/ossrs/srs/pull/3597): Fix RBSP stream parsing bug, should drop 0x03. v5.0.178 (#3597)
 * v5.0, 2023-09-07, Merge [#3795](https://github.com/ossrs/srs/pull/3795): Fix dash crash if format not supported. v5.0.177 (#3795)
 * v5.0, 2023-08-30, Merge [#3779](https://github.com/ossrs/srs/pull/3779): Support HTTP-API for fetching reload result. v5.0.176 (#3779)
--- a/trunk/src/app/srs_app_utility.cpp
+++ b/trunk/src/app/srs_app_utility.cpp
@@ -1379,65 +1379,6 @@ void srs_api_dump_summaries(SrsJsonObject* obj)
     sys->set("conn_srs", SrsJsonAny::integer(nrs->nb_conn_srs));
 }
 
-string srs_string_dumps_hex(const std::string& str)
-{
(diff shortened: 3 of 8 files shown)


trunk/src/protocol/srs_protocol_rtmp_stack.cpp
```diff
@@ -595,6 +596,12 @@ srs_error_t SrsProtocol::do_decode_message(SrsMessageHeader& header, SrsBuffer*
     
     // decode specified packet type
     if (header.is_amf0_command() || header.is_amf3_command() || header.is_amf0_data() || header.is_amf3_data()) {
+        // Ignore FFmpeg timecode, see https://github.com/ossrs/srs/issues/3803
+        if (stream->left() == 4 && (uint8_t)*stream->head() == 0x00) {
```
xiaozhihong, line 599: If the first byte is not 0x00, an error will be judged here

`TRANS_BY_GPT4`

xiaozhihong (approved):

trunk/src/protocol/srs_protocol_rtmp_stack.cpp
winlinvip: It should be a 4-byte timestamp, not necessarily starting from 0, but if the meaning of the timecode is startime, there is a high probability that it starts with 0. The high 8 bits should not be used, otherwise, it would be a very large starttime.

The AMF0 string starts with 02, but it should not be 4 bytes, a normal String would definitely not be this short.

So if we only judge by 4 bytes, it's quite lenient, but I think we should make a narrower judgment to avoid problems in some unknown scenarios, so I added the condition of starting with 0.

`TRANS_BY_GPT4`

winlinvip: We've added a new warning log, and we're ignoring messages about `timecode`.

`TRANS_BY_GPT4`

winlinvip: If the first byte isn't 0x00, it will be considered as an error here.

`TRANS_BY_GPT4`

streamingsystems:
https://github.com/ossrs/srs/assets/59889605/e72004cc-7df1-403e-8f83-654d3ee4a269

Hi, 

It seems like maybe this problem is not fixed or there is a related bug. I am running FFMPEG 6 and the latest version of SRS (I just built it from the source). 

When I run this:

ffmpeg -stream_loop -1 -re -i 2.mp4 -stream_loop -1 -re -i 2.mp4 -stream_loop -1 -re -i 2.mp4 -map 0 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/1 -map 1 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/2 -map 2 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/3

(In my real application I have different video as inputs but they were too big to all upload to GitHub but the above command will use the same input video and reproduce the problem)

When the video is done and starts over at the beginning ffmpeg crashes with this: 

av_interleaved_write_frame(): Broken pipesize=    4881kB time=00:00:49.98 bitrate= 800.0kbits/s speed=   1x
[out#1/flv @ 0x7fd52e90cec0] Error muxing a packet
av_interleaved_write_frame(): Broken pipe
[out#2/flv @ 0x7fd52ea112c0] Error muxing a packet

"Broken Pipe" usually means the server hit some type of an error. 

Do you have any developers that also work on FFMPEG we would pay the money to fix this problem in ffmpeg as it seems ffmpeg is not sending the correct bytes in this situation. 

Thanks!

winlinvip:
@streamingsystems In fact, you don't need to modify FFmpeg. I noticed that there is a timeout during the Stream loop. So, you just need to extend the timeout period below to solve the problem:

```nginx
vhost __defaultVhost__ {
    publish { 
        firstpkt_timeout 30000; 
        normal_timeout 30000; 
    }
}
````

`TRANS_BY_GPT4`

streamingsystems:
Good evening, 

It seems that ffmpeg is not sending some data properly that is causing a longer timeout needed? 

If you just use "1" map like this:

ffmpeg -stream_loop -1 -re -i 2.mp4 -map 0 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/1 -map 1

ffmpeg does not exit with a "broken pipe" and runs correctly with no change to the timeout. 

But if I do 2 (or more( maps:

ffmpeg -stream_loop -1 -re -i 2.mp4 -stream_loop -1 -re -i 2.mp4 -map 0 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/1 -map 1 -c copy -f flv rtmp://127.0.0.1:1935/live/ss/2

av_interleaved_write_frame(): Broken pipe  4855kB time=00:00:49.55 bitrate= 802.6kbits/s speed=   1x

To me, this seems to be a bug in the "map" or "rtmp" code in ffmpeg that it's behaving differently by maybe not sending the data related to timecode that is causing the server to hit a timeout and fail?

I use the timouts low for other reasons so I am trying to not have to increase the timeout for what seems to be a bug in ffmpeg. 

As always thanks for your help.

winlinvip:
@streamingsystems In fact, we cannot completely rule out the possibility of a bug in FFmpeg. You could report this to the FFmpeg community. By increasing the timeout configuration in SRS, you should be able to get around this issue without affecting your business. Additionally, if 'map 0' is problematic, is it possible to not use 'map 0' and only use 'map 1', 'map 2', 'map N', etc.?

streamingsystems:
Hi,

I don't believe that map 0 specifically is the problem, it seems when you use more than 1 map (map 0 being the first) it causes the problem. 

What is puzzling and problematic to me is that it seems like ffmpeg with more than 1 map in the command is not sending the proper timing information to the server which is causing the timeout to be hit. 

In other words, with just a single map the server does not need a 30s timeout and ffmpeg works fine. When more than 1 map is used, the server times out which would imply to me that ffmpeg is not sending the proper data to the server which is causing it to timeout. 

I am trying to find a ffmpeg developer that can look at the "map" function. 

If you see this bug report I opened a while ago: 

https://trac.ffmpeg.org/ticket/10565

Steven Luis replies with this comment:

Looks the problem is happened when it publish timecode stream:

  Stream #0:2[0x3](eng): Data: none (tmcd / 0x64636D74)
    Metadata:
      handler_name    : VideoHandler
      timecode        : 01:00:00:00

I am emailing with Steven now to see if I can pay him to look at the code and see if maybe ffmpeg is not sending up the proper timecode information when more than 1 map is used.

winlinvip:
Steven Liu is my good friend and also the maintainer of the FFmpeg RTMP/FLV module. If he is willing to spend time on it, there is hope that he can solve this problem.

`TRANS_BY_GPT4`

PR #3807, opened 2023-09-19
Prevent the output of srt logs in utest. v5.0.181 v6.0.79

xiaozhihong:
Prevent the output of srt logs in utest.

Before 

image

After 

image

---------

### DIFF
--- a/trunk/src/utest/srs_utest.cpp
+++ b/trunk/src/utest/srs_utest.cpp
@@ -21,6 +21,7 @@ using namespace std;
 #include <sys/mman.h>
 
 #ifdef SRS_SRT
+#include <srt/srt.h>
 #include <srs_app_srt_server.hpp>
 #endif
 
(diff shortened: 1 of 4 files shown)


trunk/src/utest/srs_utest.cpp
```diff
@@ -45,6 +46,14 @@ const char* _srs_binary = NULL;
 
 #include <srs_app_st.hpp>
 
+#ifdef SRS_SRT
+static void srs_srt_utest_null_log_handler(void* opaque, int level, const char* file, int line,
```
chundonglinlin, line 50: 1. Can we remove null from the function name?
2. After removing the log, will some errors be ignored? Can we judge the level, for example?
```cpp
  switch (level) {
    case LOG_CRIT:
      break;
    case LOG_ERR:
      break;
    case LOG_WARNING:
      break;
    case LOG_NOTICE:
      break;
    case LOG_DEBUG:
      break;
    default:
      break;
  }
```

`TRANS_BY_GPT4`

xiaozhihong: There's no need, there are error cases in the test cases. As long as the expected result is correct, it's fine. In such cases, libsrt might output a lot of logs.

`TRANS_BY_GPT4`

winlinvip (approved):

PR #3824, opened 2023-10-07
Solve the problem of inaccurate HLS TS duration. v5.0.187 v6.0.87

winlinvip:
1. The comment on the ratio configuration says it can affect the slice duration, but there is no effect after configuring it.
2. The default hls_td_ratio is 1.5, and after setting it to 1, the duration is still slightly more than 10 seconds.
3. Even if the GOP is an integer, like 1 second, the slice is still a non-integer, like 0.998 seconds, which seems a bit unreliable.
4. In the duration of the TS in the m3u8 file, it is one frame less than the duration of the slice.
5. Set hls_dispose to 120s to dispose HLS files when no stream.
6. Use docker.conf for docker.

Before this patch:

```
#EXTINF:10.983, no desc
livestream-0.ts?hls_ctx=3p095hq0
```

After this patch:

```
#EXTINF:10.000, no desc
livestream-0.ts?hls_ctx=3p095hq0
```

Note: If the fragment is set to 10 seconds, but the GOP size cannot be divided by 10, such as not 1, 2, 5, or 10, then the duration of ts will still be more than 10 seconds.

---------

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,5 +57,5 @@ RUN ldd /usr/local/srs/objs/ffmpeg/bin/ffmpeg && \
 # Default workdir and command.
 WORKDIR /usr/local/srs
 ENV SRS_DAEMON=off
-CMD ["./objs/srs", "-c", "conf/srs.conf"]
+CMD ["./objs/srs", "-c", "conf/docker.conf"]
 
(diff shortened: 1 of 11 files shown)


winlinvip:

`TRANS_BY_GPT4`

xiaozhihong (approved):

xiaozhihong (approved):

PR #3828, opened 2023-10-09
Config for production

canhta:


xiaozhihong:
What's the point of this PR? What are the benefits? A bunch of hardcoded IPs, it's just for your own convenience, right?

`TRANS_BY_GPT4`

Issue #2843, opened 2022-01-04
WebRTC: UDP/8000 check tool 检测UDP/8000是否可以访问的工具


winlinvip:
今天发现一个非常奇怪的问题，UDP/8000端口就是不通，安全组和侦听都没有问题，捣腾了半天还是给云服务商提工单排查了。

提工单过程中，发现要排查UDP端口是否通，是个比较麻烦的问题，不像TCP那样可以看22端口是否可以访问，一般没有一个通用的UDP服务器，所以也不好排查。

找了一些简单的UDP的服务器和客户端，可以跑在自己的服务器上，然后在客户端上测试下：

## Go

代码片段：[Gist](https://gist.github.com/winlinvip/e8665ba888e2fd489ccd5a449fadfa73)

启动服务器：

```
go run server.go
```

执行命令看下侦听是否正常：

```
netstat -lnptu | grep  8000
# udp    0    0 0.0.0.0:8000    0.0.0.0:*    1016220/server
```

执行nc命令，给服务器发送消息`Hello`，然后按`Enter`：

```
nc -u 127.0.0.1 8000
# Hello
# Pong: Hello
```

在本机运行客户端：

```
go run client.go
```

结果应该正常：

```bash
Ping: Hello, UDP server
Pong: Hello, UDP server
```

运行客户端，指定远程服务器：

```
go run client.go 101.20.7.11
```

下面是代码，防止大家访问不了gist。其他语言的欢迎补充。

## server.go

```go
/*
Usage:
	go run server.go
See https://gist.github.com/winlinvip/e8665ba888e2fd489ccd5a449fadfa73
 */
package main

import (
	"fmt"
	"net"
	"os"
)

func main() {
	addr := net.UDPAddr{
		Port: 8000,
		IP:   net.ParseIP("0.0.0.0"),
	}
	server, err := net.ListenUDP("udp4", &addr)
	if err != nil {
		fmt.Printf("Listen err %v\n", err)
		os.Exit(-1)
	}
	fmt.Printf("Listen at %v\n", addr.String())

	for {
		p := make([]byte, 1024)
		nn, raddr, err := server.ReadFromUDP(p)
		if err != nil {
			fmt.Printf("Read err  %v", err)
			continue
		}

		msg := p[:nn]
		fmt.Printf("Received %v %s\n", raddr, msg)

		go func(conn *net.UDPConn, raddr *net.UDPAddr, msg []byte) {
[...]

PR #3837, opened 2023-10-15
Support set the ice-ufrag and ice-pwd for connectivity check. v5.0.191 v6.0.91

winlinvip:
Checking the HTTPS API or UDP connectivity for WHIP tests can be difficult. For example, if the UDP port isn't available but the API is fine, OBS only says it can't connect to the server. It's hard to see the HTTPS API response or check if the UDP port is available.

This feature lets you set the ice username and password in SRS. You can then send a STUN request using nc and see the response, making it easier to check UDP port connectivity.

1. Use curl to test the WHIP API, including ice-frag and ice-pwd queries.
2. Use nc to send a STUN binding request to test UDP connectivity.
3. If both the API and UDP are working, you should get a STUN response.

Troubleshooting Connection Failures in OBS WHIP, see https://github.com/ossrs/srs/issues/2843

---------

### DIFF
--- a/trunk/src/app/srs_app_rtc_api.cpp
+++ b/trunk/src/app/srs_app_rtc_api.cpp
@@ -689,9 +689,16 @@ srs_error_t SrsGoApiRtcWhip::do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMe
     ruc->req_->vhost = ruc->req_->host;
     ruc->req_->app = app.empty() ? "live" : app;
     ruc->req_->stream = stream.empty() ? "livestream" : stream;
+    ruc->req_->param = r->query();
+
     ruc->req_->ice_ufrag_ = r->query_get("ice-ufrag");
     ruc->req_->ice_pwd_ = r->query_get("ice-pwd");
-    ruc->req_->param = r->query();
+    if (!ruc->req_->ice_ufrag_.empty() && (ruc->req_->ice_ufrag_.length() < 4 || ruc->req_->ice_ufrag_.length() > 32)) {
+        return srs_error_new(ERROR_RTC_INVALID_ICE, "Invalid ice-ufrag %s", ruc->req_->ice_ufrag_.c_str());
+    }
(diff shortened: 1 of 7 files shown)


winlinvip:
We need to combine 5 and 6.

`TRANS_BY_GPT4`

winlinvip:
There's another related improvement. When accessing the WHIP API using HTTP GET, the error messages are not clear enough. You can refer to this link for more details: https://github.com/ossrs/srs/issues/3838

`TRANS_BY_GPT4`

trunk/src/app/srs_app_rtc_api.cpp
```diff
@@ -691,6 +691,15 @@ srs_error_t SrsGoApiRtcWhip::do_serve_http(ISrsHttpResponseWriter* w, ISrsHttpMe
     ruc->req_->stream = stream.empty() ? "livestream" : stream;
     ruc->req_->param = r->query();
 
+    ruc->req_->ice_ufrag_ = r->query_get("ice-ufrag");
+    ruc->req_->ice_pwd_ = r->query_get("ice-pwd");
+    if (!ruc->req_->ice_ufrag_.empty() && (ruc->req_->ice_ufrag_.length() < 4 || ruc->req_->ice_ufrag_.length() > 32)) {
+        return srs_error_new(ERROR_RTC_INVALID_ICE, "Invalid ice-ufrag %s", ruc->req_->ice_ufrag_.c_str());
+    }
+    if (!ruc->req_->ice_pwd_.empty() && (ruc->req_->ice_pwd_.length() < 22 || ruc->req_->ice_pwd_.length() > 32)) {
```
xiaozhihong: Where are these magic numbers defined? Can they be replaced with const variables?

`TRANS_BY_GPT4`

winlinvip: Replaced with macro definitions and added comments.

`TRANS_BY_GPT4`

xiaozhihong (approved):

PR #3852, opened 2023-10-27
RTC: Refine FFmpeg opus audio noisy issue. v5.0.197 v6.0.97

chundonglinlin:
### Description

When converting between AAC and Opus formats (aac2opus or opus2aac), the `av_frame_get_buffer` API is frequently called.

### Objective

The goal is to optimize the code logic and reduce the frequent allocation and deallocation of memory.

In the case of aac2opus, av_frame_get_buffer is still frequently called.
In the case of opus2aac, the goal is to avoid calling av_frame_get_buffer and reduce memory allocations.

### Additional Note

Before calling the `av_audio_fifo_read` API, use `av_frame_make_writable` to check if the frame is writable. If it is not writable, create a new frame.

---------

`TRANS_BY_GPT4`

---------

### DIFF
--- a/trunk/src/app/srs_app_rtc_codec.cpp
+++ b/trunk/src/app/srs_app_rtc_codec.cpp
@@ -260,6 +260,13 @@ srs_error_t SrsAudioTranscoder::init_enc(SrsAudioCodecId dst_codec, int dst_chan
     if (!enc_frame_) {
         return srs_error_new(ERROR_RTC_RTP_MUXER, "Could not allocate audio encode in frame");
     }
+    enc_frame_->format = enc_->sample_fmt;
+    enc_frame_->nb_samples = enc_->frame_size;
+    enc_frame_->channel_layout = enc_->channel_layout;
+
+    if (av_frame_get_buffer(enc_frame_, 0) < 0) {
+        return srs_error_new(ERROR_RTC_RTP_MUXER, "Could not get audio frame buffer");
+    }
 
     enc_packet_ = av_packet_alloc();
     if (!enc_packet_) {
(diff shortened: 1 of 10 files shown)


winlinvip:
Why should we make this change? What problems might arise? What issues does it solve?

`TRANS_BY_GPT4`

trunk/src/app/srs_app_rtc_codec.cpp
```diff
@@ -381,26 +388,21 @@ srs_error_t SrsAudioTranscoder::encode(std::vector<SrsAudioFrame*> &pkts)
...
-        enc_frame_->format = enc_->sample_fmt;
-        enc_frame_->nb_samples = enc_->frame_size;
-        enc_frame_->channel_layout = enc_->channel_layout;
-
-        if (av_frame_get_buffer(enc_frame_, 0) < 0) {
-            av_frame_free(&enc_frame_);
-            return srs_error_new(ERROR_RTC_RTP_MUXER, "Could not get audio frame buffer");
+        // make sure the frame is writable
+        if (av_frame_make_writable(enc_frame_) < 0) {
+            return srs_error_new(ERROR_RTC_RTP_MUXER, "Could not make writable frame");
```
xiaozhihong, line 393: 这里是不是需要av_alloc_frame, av_frame_get_buffer, 而不是直接返回错误

xiaozhihong: > Should we use `av_alloc_frame`, `av_frame_get_buffer` here, instead of directly returning an error?

> After checking, it's not necessary. `av_frame_make_writable` has already handled it.
image

`TRANS_BY_GPT4`

winlinvip (approved):

PR #3858, opened 2023-10-31
Support build without cache to test if actions fail. v5.0.196 v6.0.96

chundonglinlin:
By default, caching is enabled during compilation, which means that data is cached in Docker. This helps to avoid compiling third-party dependency libraries. However, sometimes when updating third-party libraries, it's necessary to disable caching to temporarily verify if the pipeline can succeed. Therefore, a configure option should be added. When this option is enabled, the compilation cache will not be used, and all third-party libraries will be compiled from scratch.

### DIFF
--- a/trunk/auto/options.sh
+++ b/trunk/auto/options.sh
@@ -37,7 +37,7 @@ SRS_FFMPEG_TOOL=NO
 # FFmpeg fit is the source code for RTC, to transcode audio or video in SRS.
 SRS_FFMPEG_FIT=RESERVED
 # Whether use FFmpeg native opus codec for RTC. If not, use libopus instead.
-SRS_FFMPEG_OPUS=YES
+SRS_FFMPEG_OPUS=NO
 # arguments
 SRS_PREFIX=/usr/local/srs
 SRS_DEFAULT_CONFIG=conf/srs.conf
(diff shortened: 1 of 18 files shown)


trunk/auto/options.sh
```diff
@@ -37,7 +37,7 @@ SRS_FFMPEG_TOOL=NO
 # FFmpeg fit is the source code for RTC, to transcode audio or video in SRS.
 SRS_FFMPEG_FIT=RESERVED
 # Whether use FFmpeg native opus codec for RTC. If not, use libopus instead.
-SRS_FFMPEG_OPUS=YES
+SRS_FFMPEG_OPUS=NO
```
winlinvip: Your modification won't work. This is because `ffmpeg-opus=on` is set in the cache. Even if you change the default to `off`, there will still be issues, specifically the previous crashing problem.

You can add an option, `--build-cache`, which is `on` by default. If it's `off`, the cache won't be used. You can check this switch in the following location:

```bash
if [[ -d /usr/local/srs-cache/srs/trunk/objs && $(pwd) != "/usr/local/srs-cache/srs/trunk" ]]; then
```

This switch doesn't exist yet. You need to add it in your Pull Request (PR).

`TRANS_BY_GPT4`

winlinvip (approved):

PR #3869, opened 2023-11-09
A demo for SRT proxy.

winlinvip:
See https://www.figma.com/file/kItb5HWOI4HimjDp62pas3/SRT-Proxy
And the architecture https://github.com/ossrs/srs/discussions/3634

image


winlinvip:
You only need to merge into 6.0. @xiaozhihong

`TRANS_BY_GPT4`

PR #3883, opened 2023-11-17
Fix opus delay options, use ffmpeg-opus in docker test. v6.0.102

xiaozhihong:
The `ffmpeg-opus` tool allows you to control the delay using the `opus_delay` option. The minimum delay can be set to 2.5ms. However, in practice, you cannot set it this low. You need to set at least 10 frames to allow the audio encoder to lookahead. Otherwise, the sound will be distorted.

`TRANS_BY_GPT4`

---------

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,7 +1,7 @@
 # These are supported funding model platforms
 
 github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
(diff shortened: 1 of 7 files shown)


xiaozhihong:
ffmpeg-opus可以用选项`opus_delay`控制延迟, 最小可以是2.5ms, 但是实际上不能这么设置,需要设置至少10帧, 让音频编码器lookahead, 不然声音听起来会很失真.

chundonglinlin (approved):

PR #3891, opened 2023-11-22
fix 'sed' error in options.sh. v5.0.201 v6.0.103

duiniuluantanqin:
The `-` character, when placed in the middle of a regular expression, is interpreted as a range. It must be placed at the beginning or end to be interpreted as a literal character.

---------

`TRANS_BY_GPT4`

---------


xiaozhihong:
Before 

image



After

image


It works well

xiaozhihong:
The '-' symbol must be placed at the beginning or end. If it's in the middle, it will be interpreted as a range. For example, '0-9' is interpreted as '0123456789'.

`TRANS_BY_GPT4`

xiaozhihong (approved):

PR #3920, opened 2023-12-30
WHIP: Fix bug for converting WHIP to RTMP/HLS. v5.0.208 v6.0.113

winlinvip:
1. When converting RTC to RTMP, it is necessary to synchronize the audio and video timestamps. When the synchronization status changes, whether it is unsynchronized or synchronized, print logs to facilitate troubleshooting of such issues.
2. Chrome uses the STAP-A packet, which means a single RTP packet contains SPS/PPS information. OBS WHIP, on the other hand, sends SPS and PPS in separate RTP packets. Therefore, SPS and PPS are in two independent RTP packets, and SRS needs to cache these two packets.

---------


trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1533,7 +1537,7 @@ srs_error_t SrsRtcFrameBuilder::packet_video_key_frame(SrsRtpPacket* pkt)
     SrsRtpSTAPPayload* stap_payload = dynamic_cast<SrsRtpSTAPPayload*>(pkt->payload());
 
     // Handle SPS/PPS in cache or STAP-A packet.
-    if (stap_payload || raw_payload) {
+    if (stap_payload || (raw_payload && raw_is_sps_pps)) {
```
winlinvip: When raw_is_sps_pps is set to true, the raw_payload must not be null.

The logic you added dictates that the generation of the Sequence header should not be triggered when raw_payload is present but does not contain sps or pps.

`TRANS_BY_GPT4`

winlinvip:
Test the WHIP+WHEP, and WHIP+FLV, all work very well.

image

image

winlinvip:
Need to merge to SRS 5.0 and 6.0 @xiaozhihong

xiaozhihong (approved):

PR #3924, opened 2024-01-06
Upgrade hls.js and set in low latency mode. v6.0.112

winlinvip:
HLS typically has a delay of around 30 seconds, roughly comprising three segments, each lasting 10 seconds. We can reduce the delay to about 5 seconds by lowering the segment duration to 2 seconds and starting playback from the last segment, achieving a stable delay.

Of course, this requires setting the OBS's GOP to 1 second, and the profile to baseline, preset to fast, and tune to zerolatency. Additionally, updating a few configurations in the hls.js player is necessary, such as setting it to start playback from the last segment, setting the maximum buffer, and initiating accelerated playback to reduce latency.

---------


winlinvip:
image

trunk/research/players/srs_player.html
```diff
@@ -264,16 +275,23 @@
...
             return;
         }
 
-        console.error('不支持的URL', r.url, r);
         $('#video_player').hide();
-        $('#audio_player').hide();
+
+        const msg = `Not supported URL ${r.url} ${JSON.stringify(r)}`;
+        console.error(msg);
+        alert(msg);
```
chundonglinlin, line 297: Can the `alert` be removed or can the content of the prompt be modified?

`TRANS_BY_GPT4`

chundonglinlin (approved):

trunk/research/players/srs_player.html
winlinvip: This player is intended for developers, so it would be appropriate to provide more information.

`TRANS_BY_GPT4`

Dockerfile
```diff
@@ -29,7 +29,7 @@ WORKDIR /srs/trunk
 # Build and install SRS.
 # Note that SRT is enabled by default, so we configure without --srt=on.
 # Note that we have copied all files by make install.
-RUN ./configure --sanitizer=off --gb28181=on --h265=on ${CONFARGS} && make ${MAKEARGS} && make install
```
winlinvip, line 32: Removed the `--h265=on` option because it is enabled by default. Moved the `--sanitizer --gb28181` options to the release.yml file.

`TRANS_BY_GPT4`

trunk/src/kernel/srs_kernel_codec.hpp
```diff
@@ -778,7 +778,7 @@ struct SrsHevcRbspVps
     uint32_t vps_max_latency_increase_plus1[8];
     uint8_t vps_max_layer_id;
     uint32_t vps_num_layer_sets_minus1;
-    std::vector<std::vector<uint8_t>> layer_id_included_flag;
+    std::vector< std::vector<uint8_t> > layer_id_included_flag;
```
winlinvip, line 781: Compiling C++ 98 on CentOS 7 will result in failure.

`TRANS_BY_GPT4`

PR #4024, opened 2024-04-14
Debug: reuse endpoint with RTC and server error.

suzp1984:
https://github.com/ossrs/srs/blob/427104f1dab86f5afc7d7b49b02ed27d03ef9346/trunk/src/app/srs_app_server.cpp#L1197-L1201

I believe this is a mistake, if reuse the RTC and http server listener, then Http Server connection will be created with incorrect SSL options.


suzp1984:
conflict with #4028 , so close this one.

PR #4028, opened 2024-04-16
HTTPS: Support config key/cert for HTTPS API. v6.0.137

### LINKED ISSUE #4025: Bug: SrsHttpxConn SSL Key&Cert config problem.
**Describe the bug**
https://github.com/ossrs/srs/blob/427104f1dab86f5afc7d7b49b02ed27d03ef9346/trunk/src/app/srs_app_http_conn.cpp#L382-L387

`SrsHttpxConn` can be used as HTTP API and HTTP Server, for both plain and SSL connection. For the SSL connection, we can config the SSL key & cert in this way.

https://github.com/ossrs/srs/blob/427104f1dab86f5afc7d7b49b02ed27d03ef9346/trunk/conf/https.srs.conf#L7-L28

And `SrsConfig` has apis to get the key & cert.

https://github.com/ossrs/srs/blob/427104f1dab86f5afc7d7b49b02ed27d03ef9346/trunk/src/app/srs_app_config.hpp#L1052-L1053
https://github.com/ossrs/srs/blob/427104f1dab86f5afc7d7b49b02ed27d03ef9346/trunk/src/app/srs_app_config.hpp#L1074-L1075

But `SrsHttpxConn` only calling `get_https_stream_ssl_cert` & `get_https_stream_ssl_key` even for the `Https API` connections.

**Version**
All SRS version.

**To Reproduce**
1. config https api and https stream with different key & cert pair.
2. boot the srs.

**Expected behavior**
http_api.https.key | cert should be loaded correctly. 

**Additional context**
I found this bug when try to do #3701, found this bug and also #4024

suzp1984:


trunk/src/app/srs_app_http_conn.hpp
```diff
@@ -136,8 +136,18 @@ class SrsHttpxConn : public ISrsConnection, public ISrsStartable, public ISrsHtt
     SrsHttpConn* conn;
     // We should never enable the stat, unless HTTP stream connection requires.
     bool enable_stat_;
-public:
-    SrsHttpxConn(bool https, ISrsResourceManager* cm, ISrsProtocolReadWriter* io, ISrsHttpServeMux* m, std::string cip, int port);
+    // ssl key & cert file
+    const std::string ssl_key_file;
```
winlinvip: For member field, should end width `_`, so should rename to `ssl_key_file_` and `ssl_cert_file_`

trunk/src/app/srs_app_server.cpp
```diff
@@ -1207,19 +1210,22 @@ srs_error_t SrsServer::do_on_tcp_client(ISrsListener* listener, srs_netfd_t& stf
             resource = new SrsRtmpConn(this, stfd2, ip, port);
         } else if (listener == api_listener_ || listener == apis_listener_) {
             bool is_https = listener == apis_listener_;
-            resource = new SrsHttpxConn(is_https, this, new SrsTcpConnection(stfd2), http_api_mux, ip, port);
+            resource = new SrsHttpxConn(this, new SrsTcpConnection(stfd2), http_api_mux, ip, port,
```
winlinvip: Change to:

```cpp
string key = listener == apis_listener_ ? _srs_config->get_https_api_ssl_key() : "";
string cert = listener == apis_listener_ ? _srs_config-> get_https_api_ssl_cert() : "";
resource = new SrsHttpxConn(this, new SrsTcpConnection(stfd2), http_api_mux, ip, port, key, cert);
```

trunk/src/app/srs_app_http_conn.hpp
```diff
@@ -136,8 +136,18 @@ class SrsHttpxConn : public ISrsConnection, public ISrsStartable, public ISrsHtt
...
+    const std::string ssl_cert_file;
+    
+public:
+    SrsHttpxConn(ISrsResourceManager* cm,
+                 ISrsProtocolReadWriter* io,
+                 ISrsHttpServeMux* m,
+                 std::string cip,
+                 int port,
+                 const std::string& ssl_key_file = "",
+                 const std::string& ssl_cert_file = "");
```
winlinvip: Multiple lines is very difficult to read, no need to use `const std::string&`, just use `string`, because it's not hot path and not bottleneck. We must keep it easy to read:

```cpp

SrsHttpxConn(ISrsResourceManager* cm, ISrsProtocolReadWriter* io, ISrsHttpServeMux* m, std::string cip, int port, string key, string cert);
```

No default value generally.

trunk/src/app/srs_app_http_conn.hpp
suzp1984: fixed.

trunk/src/app/srs_app_server.cpp
suzp1984: fixed

trunk/src/app/srs_app_http_conn.hpp
suzp1984: fixed.

PR #4030, opened 2024-04-17
fix bootstrap css warning.

suzp1984:
> Screenshot 2024-04-17 at 9 17 07 AM

Focus on the `img` css, the `width: auto\9;` warning. There are 14 '\9' inside `bootstrap.min.css`. And I can't find `v2.3.2` version to compare. So just replace the 14 '\9' with empty space. 

---------

`TRANS_BY_GPT4`


winlinvip:
We should never modify the library, instead we should:

1. Replace with new versions, which may fix that issue.
2. Feed back and file PR to upstream to fix it in bootstrap.

PR #4042, opened 2024-04-24
Refine config directive token parsing. v6.0.135

### LINKED ISSUE #2228: Config: Failed to parse config ending with newline.
**Description**'

Please ensure that the markdown structure is maintained.

> Documentation for realtime config has an error

1. SRS version: `3.0.153`
1. The log of SRS is as follows:'

Please ensure that the markdown structure is maintained.
```
start play smi=0ms, mw_sleep=100, mw_enabled=1, realtime=0, tcp_nodelay=1
```
1. The configuration of SRS is as follows:

Please ensure that the markdown structure is maintained.
```
# the listen ports, split by space.
listen              1935;
vhost __defaultVhost__ {
    tcp_nodelay     on
    min_latency     on;

    play {
        gop_cache       off;
        queue_length    10;
        mw_latency      100;
    }

    publish {
        mr off;
    }
}
```

**Replay**

Please ensure that the markdown structure is maintained.

> Play stream with rtmp client and check log

**Expect**

Please ensure that the markdown structure is maintained.

```
realtime=1
```
If I add the missing semi-colon to the config after `tcp_nodelay     on` then the min_latency setting is applied correctly. The sample configuration is incorrect in both `conf/realtime.conf` and the wiki in both languages.

`TRANS_BY_GPT3`

suzp1984:
make sure one directive token don't span more than two lines.

try to fix #2228


trunk/scripts/verify_confs.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+TRUNK_DIR=$(dirname -- "${BASH_SOURCE[0]}")/../
```
winlinvip: Test file? Should remove it.

trunk/src/utest/srs_utest.cpp
```diff
@@ -223,7 +223,7 @@ VOID TEST(SampleTest, ContextTest)
     cache[0] = cid;
 }
 
-MockProtectedBuffer::MockProtectedBuffer() : size_(0), data_(NULL), raw_memory_(NULL)
+MockProtectedBuffer::MockProtectedBuffer() : raw_memory_(NULL), size_(0), data_(NULL)
```
winlinvip, line 226: Nothing changed?

suzp1984: Yes, just to fix a compiler warning: the init order should be the declare orders in the header file.

https://github.com/ossrs/srs/blob/37f0faae5a36bc9af2b06c91df52aead73e76add/trunk/src/utest/srs_utest.hpp#L112-L117

trunk/scripts/verify_confs.sh
suzp1984: No, the var store the path of `trunk` folder, the almost the source root of srs project, then doing the next steps.
run `srs -t -c [conf file]` for every file inside `${TRUNK_DIR}/conf/*`.

PR #4072, opened 2024-05-29
Feat: HLS support display subtitle WebVTT

fjqgx:


winlinvip:
The issue was identified as being caused by an outdated version of hls.js used by Oryx, which has not been updated, resulting in the inability to play WebVTT. Here is the relevant commit link on GitHub: https://github.com/ossrs/oryx/commit/e29be8c3c88b968578ad48856652484ff14d2479

`TRANS_BY_GPT4`

PR #4073, opened 2024-05-29
improve http response when err happens.

suzp1984:
1. response 200 with code when rtc/v1 remote sdp parse error;
2. response 500 when no handled err.

Try to improve two problem:

1. let `rtc/v1/[whep|whip]` response 200 with body `{code : 400}` when `remote_sdp` parse error happen by follow `rtc/v1/[play | publish]` api's response. The current behavior in this scenario is close the connection without response anything. 

2. Let `SrsHttpServeMux` response 500 when err not handled, the current behavior is also to close the socket connection without response anything.


winlinvip:

PR #4077, opened 2024-06-03
Fix misspelling error in app config. v6.0.133

suzp1984:
1. misspelling fix;
2. remove finished TODO;

---------

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -45,9 +45,9 @@ using namespace srs_internal;
 // @global the version to identify the core.
 const char* _srs_version = "XCORE-" RTMP_SIG_SRS_SERVER;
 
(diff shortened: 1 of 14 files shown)


trunk/src/app/srs_app_config.cpp
```diff
@@ -46,8 +46,8 @@ using namespace srs_internal;
 const char* _srs_version = "XCORE-" RTMP_SIG_SRS_SERVER;
 
 // when user config an invalid value, macros to perfer true or false.
```
duiniuluantanqin: The comment containing `perfer` should also be corrected.

`TRANS_BY_GPT4`

suzp1984: done.
And I search all the `perfer` in `src/` folder, renamed all of them.
`grep -Ri 'perfer' src/`

duiniuluantanqin (approved):

PR #4080, opened 2024-06-07
SmartPtr: Use shared ptr to manage GB objects. v6.0.126

winlinvip:
The object relations: 

gb

Session manages SIP and Media object using shared resource or shared ptr. Note that I actually use SrsExecutorCoroutine to   delete the object when each coroutine is done, because there is always a dedicate coroutine for each object.

For SIP and Media object, they directly use the session by raw pointer, it's safe because session always live longer than session and media object.


trunk/src/app/srs_app_conn.hpp
```diff
@@ -125,98 +126,68 @@ class SrsResourceManager : public ISrsCoroutineHandler, public ISrsResourceManag
...
 public:
-    SrsLazyObjectWrapper() {
-        init(new T(this));
-    }
-    virtual ~SrsLazyObjectWrapper() {
-        resource_->gc_dispose();
-        if (resource_->gc_ref() == 0) {
-            _srs_gc->remove(resource_);
-        }
+    SrsSharedResource(T* ptr) : ptr_(ptr) {
```
suzp1984, line 151: `SrsSharedResource(const SrsSharedPtr<T>& shared_ptr): ptr_(shared_ptr) {}`
I would like to add constructor with `SrsSharedPtr` also.

```
 SrsSharedPtr<T> shared_ptr = new SrsSharedPtr<MyClass>(new MyClass());
 SrsSharedResource<MyClass>* ptr = new SrsSharedResource<MyClass>(shared_ptr);
 (*ptr)->do_something();
```

winlinvip: What is the use scenario?

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -70,11 +70,15 @@ std::string srs_sip_state(SrsGbSipState ostate, SrsGbSipState state)
     return srs_fmt("%s->%s", srs_gb_sip_state(ostate).c_str(), srs_gb_sip_state(state).c_str());
 }
 
-SrsLazyGbSession::SrsLazyGbSession(SrsLazyObjectWrapper<SrsLazyGbSession>* wrapper_root)
+SrsLazyGbSession::SrsLazyGbSession() : sip_(NULL), media_(NULL)
```
suzp1984: I would always prefer initialize the variable in the initialization list, the benefit is to avoid to invoke constructor, assign constructor and destructor again in this constructor body. (line 79 & 80).
```suggestion
SrsLazyGbSession::SrsLazyGbSession() : sip_(new SrsLazyGbSipTcpConn), media_(new SrsLazyGbMediaTcpConn())
```

of course remove line 79 & 80.

trunk/src/app/srs_app_conn.hpp
```diff
@@ -125,98 +126,74 @@ class SrsResourceManager : public ISrsCoroutineHandler, public ISrsResourceManag
...
 //
-// Note that under-layer resource are destroyed by _srs_gc, which is literally equal to srs_freep. However, the root
-// wrapper might be managed by other resource manager, such as _srs_gb_manager for SrsLazyGbSession. Furthermore, other
-// copied out wrappers might be freed by srs_freep. All are ok, because all wrapper and resources are simply normal
-// object, so if you added to manager then you should use manager to remove it, and you can also directly delete it.
+//      ISrsResourceManager* manager = ...;
+//      manager->remove(ptr);
 template<typename T>
-class SrsLazyObjectWrapper : public ISrsResource
+class SrsSharedResource : virtual public ISrsResource
```
suzp1984: About the `virtual base class`, there are no Dreaded Diamond inheritances exist yet, so no need to keep `virtual` keyword here.
ref: 
https://www.geeksforgeeks.org/virtual-base-class-in-c/
https://stackoverflow.com/questions/21558/in-c-what-is-a-virtual-base-class

trunk/src/core/srs_core_autofree.hpp
```diff
@@ -81,4 +81,103 @@ class impl_SrsAutoFree
...
 
+// Shared ptr smart pointer, see https://github.com/ossrs/srs/discussions/3667#discussioncomment-8969107
+// Usage:
+//      SrsSharedPtr<MyClass> ptr(new MyClass());
+//      ptr->do_something();
+//
+//      SrsSharedPtr<MyClass> cp = ptr;
+//      cp->do_something();
+template<class T>
+class SrsSharedPtr
```
suzp1984, line 92: It seems `SrsSharedPtr` shared similar api with `std::shared_ptr`, that's good, so why not add a compiling option to control whether use a template alias from `std::shared_ptr` or self defined `shared ptr`.

```suggestion
#ifdef _USE_STD_SMART_PTR
template <class T>
using SrsSharedPtr = std::shared_ptr<T>;
#else
template <class T>
class SrsSharedPtr { 
// current definition 
};
#endif
```
The benefit is to quickly workaround the shared ptr bugs in the future, e.g. the `ref_count_` is not atomic, it's ok for current single-thread coroutine solution, but what if multi-thread coroutine are implemented in the future, then it's disaster. For the basic level infrastructure codes, it's a smart idea to always has a replacement solution.

trunk/src/core/srs_core_autofree.hpp
```diff
@@ -81,4 +81,103 @@ class impl_SrsAutoFree
     }
 };
 
+// Shared ptr smart pointer, see https://github.com/ossrs/srs/discussions/3667#discussioncomment-8969107
```
suzp1984, line 84: Not a problem of this header file `srs_core_autofree.hpp`, but the empty `srs_core_autofree.cpp`, there is no reason to keep a empty `cpp` file except to waste compiler power.

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -899,22 +908,20 @@ srs_error_t SrsLazyGbSipTcpConn::start()
         return srs_error_wrap(err, "sender");
     }
 
-    return err;
-}
+    // Wait for the SIP connection to be terminated.
+    while (true) {
```
suzp1984: This `while` loop never break except `return`, is it a problem?

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -463,27 +458,33 @@ srs_error_t SrsGbListener::on_tcp_client(ISrsListener* listener, srs_netfd_t stf
...
-        resource->setup(conf_, sip_listener_, media_listener_, stfd);
+        SrsLazyGbSipTcpConn* raw_conn = new SrsLazyGbSipTcpConn();
+        raw_conn->setup(conf_, sip_listener_, media_listener_, stfd);
+
+        SrsSharedResource<SrsLazyGbSipTcpConn>* conn = new SrsSharedResource<SrsLazyGbSipTcpConn>(raw_conn);
+        _srs_gb_manager->add(conn, NULL);
 
-        if ((err = resource->start()) != srs_success) {
-            srs_freep(conn);
+        SrsExecutorCoroutine* executor = new SrsExecutorCoroutine(_srs_gb_manager, conn, raw_conn, raw_conn);
```
suzp1984, line 464: I think `executor` instance here is a memory leak. Here is the reason:

`SrsExecutorCorountine` is an `ISrsResource` which means it can use an `ISrsResourceManager` to manager its lifecycle, which is `_srs_gb_manager` in this case, check codes in `SrsExecutorCoroutine::cycle()`:

```
srs_error_t SrsExecutorCoroutine::cycle()
{
    srs_error_t err = handler_->cycle();
    if (callback_) callback_->on_executor_done(this);
    manager_->remove(this);
    return err;
}

```


We have `manager_->remove(this);`, but where is the `_srs_gb_manager->add(executor);`?

And Check anther two `SrsExecutorCoroutine` instances, for `SrsLazyGbMediaTcpConn` and `SrsLazyGbSession`,  they have same problem.

I didn't running any memory leak detector to confirm yet, but I think `valgrind` can detect it.

trunk/src/app/srs_app_conn.hpp
suzp1984: No such scenario yet, but missing this constructor make me felt unsafe in this case:

```
SrsSharedPtr<MyClass> shared_ptr(new MyClass());
SrsSharedResource<MyClass>* ptr = new SrsSharedResource<MyClass>(shared_ptr.get());
(*ptr)->do_something();
```

My top comment has errors, here is the fixed one.
```
 SrsSharedPtr<MyClass> shared_ptr(new MyClass());
 SrsSharedResource<MyClass>* ptr = new SrsSharedResource<MyClass>(shared_ptr);
 (*ptr)->do_something();
```

winlinvip: Won't fix. Should never over design. Should never fix non-exists problem.

trunk/src/app/srs_app_gb28181.cpp
winlinvip: Fixed.

trunk/src/app/srs_app_conn.hpp
winlinvip: Fixed.

trunk/src/core/srs_core_autofree.hpp
winlinvip: Won't fix. Using std::shared_ptr does not provide any benefits because if SrsSharedPtr has a bug, we would need to fix it.

`TRANS_BY_GPT4`

trunk/src/app/srs_app_gb28181.cpp
winlinvip: Seems no problem.

trunk/src/app/srs_app_gb28181.cpp
winlinvip: Seems no problem. Won't fix.

trunk/src/core/srs_core_autofree.hpp
winlinvip: To save on compilation time, please provide the specific amount of time saved by reducing one cpp file.
This is unrelated to the current PR; please submit a separate PR for this matter.

`TRANS_BY_GPT4`

winlinvip:
See https://github.com/ossrs/srs/discussions/3667#discussioncomment-8969107

Issue #3784, opened 2023-08-26
WebRTC: SRS 5 docker crash when RTMP2RTC over TCP.


namnm:
> Note: Please read FAQ before file an issue, see #2716

## Description

While I was testing with rtmp to webrtc, the webrtc stream worked well. Then I refreshed the browser, and srs crashed

1. SRS Version: 5

2. SRS Log:

```
[2023-08-26 02:12:23.990][INFO][1][85uf3e42] HTTP #0 172.19.0.1:43648 GET http://localhost:8080/players/rtc_player.html?autostart=true, content-length=-1
[2023-08-26 02:12:23.991][INFO][1][85uf3e42] http match file=./objs/nginx/html/players/rtc_player.html, pattern=/, upath=/players/rtc_player.html
[2023-08-26 02:12:23.993][INFO][1][85uf3e42] TCP: before dispose resource(HttpConn)(0x60700005e7b0), conns=3, zombies=0, ign=0, inz=0, ind=0
[2023-08-26 02:12:23.993][WARN][1][85uf3e42][104] client disconnect peer. ret=1007
[2023-08-26 02:12:23.993][INFO][1][95fw2y11] TCP: clear zombies=1 resources, conns=3, removing=0, unsubs=0
[2023-08-26 02:12:23.993][INFO][1][85uf3e42] TCP: disposing #0 resource(HttpConn)(0x60700005e7b0), conns=3, disposing=1, zombies=0
[2023-08-26 02:12:24.004][INFO][1][5114w0fh] DTLS: After done, got 39 bytes
[2023-08-26 02:12:24.004][INFO][1][5114w0fh] DTLS: State Passive RECV, done=1, arq=0, r0=39, len=39, cnt=21, size=26, hs=0
[2023-08-26 02:12:24.005][WARN][1][5114w0fh][0] DTLS: SSL3 alert method=read type=warning, desc=CN(close notify), where=16388, ret=256, r1=0
[2023-08-26 02:12:24.005][INFO][1][5114w0fh] RTC: session destroy by DTLS alert(warning CN), username=p81x7808:tctP
[2023-08-26 02:12:24.005][INFO][1][5114w0fh] RTC: before dispose resource(RtcConn)(0x61d000194680), conns=1, zombies=0, ign=0, inz=0, ind=0
[2023-08-26 02:12:24.005][INFO][1][5114w0fh] RTC: session detach from [5114w0fh](RtcConn), disposing=1
[...]

PR #4083, opened 2024-06-12
SmartPtr: Use shared ptr in RTC TCP connection. v6.0.127

winlinvip:
Fix issue https://github.com/ossrs/srs/issues/3784

### DIFF
--- a/trunk/src/app/srs_app_rtc_network.cpp
+++ b/trunk/src/app/srs_app_rtc_network.cpp
@@ -436,7 +436,7 @@ srs_error_t SrsRtcUdpNetwork::write(void* buf, size_t size, ssize_t* nwrite)
     return sendonly_skt_->sendto(buf, size, SRS_UTIME_NO_TIMEOUT);
 }
 
-SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* delta) : owner_(NULL)
+SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* delta) : owner_(new SrsRtcTcpConn())
 {
     conn_ = conn;
     delta_ = delta;
@@ -448,6 +448,7 @@ SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* de
 
 SrsRtcTcpNetwork::~SrsRtcTcpNetwork()
 {
+    owner_->interrupt();
     srs_freep(transport_);
 }
 
@@ -693,13 +694,20 @@ void SrsRtcTcpNetwork::dispose()
 
 #define SRS_RTC_TCP_PACKET_MAX 1500
 
-SrsRtcTcpConn::SrsRtcTcpConn(ISrsProtocolReadWriter* skt, std::string cip, int port)
+SrsRtcTcpConn::SrsRtcTcpConn()
 {
     wrapper_ = NULL;
     owner_coroutine_ = NULL;
     owner_cid_ = NULL;
     cid_ = _srs_context->get_id();
 
+    pkt_ = NULL;
+    delta_ = NULL;
+    skt_ = NULL;
+}
+
+SrsRtcTcpConn::SrsRtcTcpConn(ISrsProtocolReadWriter* skt, std::string cip, int port) : SrsRtcTcpConn()
+{
     ip_ = cip;
     port_ = port;
     skt_ = skt;
@@ -726,6 +734,12 @@ void SrsRtcTcpConn::setup_owner(SrsSharedResource<SrsRtcTcpConn>* wrapper, ISrsI
 ISrsKbpsDelta* SrsRtcTcpConn::delta()
 {
     return delta_;
+}
+
+void SrsRtcTcpConn::interrupt()
+{
+    session_ = NULL;
+    if (owner_coroutine_) owner_coroutine_->interrupt();
 }
 
 std::string SrsRtcTcpConn::desc()
@@ -863,7 +877,7 @@ srs_error_t SrsRtcTcpConn::handshake()
 
     // Should support only one TCP candidate.
     SrsRtcTcpNetwork* network = dynamic_cast<SrsRtcTcpNetwork*>(session->tcp());
-    if (!network->owner().get()) {
+    if (network->owner().get() != this) {
         network->set_owner(*wrapper_);
         session_ = session;
     }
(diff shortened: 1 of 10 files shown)


trunk/src/app/srs_app_server.cpp
```diff
@@ -1227,11 +1226,26 @@ srs_error_t SrsServer::do_on_tcp_client(ISrsListener* listener, srs_netfd_t& stf
         }
     }
 
+    // For RTC TCP connection, use resource executor to manage the resource.
```
suzp1984, line 1230: It's better to add conditional compile by following above: `#ifdef SRS_RTC`.

trunk/src/app/srs_app_server.cpp
```diff
@@ -43,6 +43,7 @@ using namespace std;
 #ifdef SRS_GB28181
 #include <srs_app_gb28181.hpp>
 #endif
+#include <srs_app_rtc_server.hpp>
```
suzp1984: Maybe add conditional compile here: `#ifdef SRS_RTC`, or consider move the `extern SrsResourceManager _srs_rtc_manager;` declaration and initialization to `srs_app_rtc_network.hpp|cpp`.

trunk/src/app/srs_app_server.cpp
suzp1984: I notice this problem, when I check the github workflow test error on `build-centos7`.
To reproduce: `./configure -rtc=off; make`

To fix this compiling error, there are still some places to fix.
1. `src/app/srs_app_stream_bridge.cpp`, delete line 11, `// #include <srs_app_rtc_codec.hpp>`, because there is no ref of `SrsAudioTranscoder` in `srs_app_stream_bridge.cpp`;
https://github.com/ossrs/srs/blob/6834ec208d67fa47c21536d1f1041bb6d60c1834/trunk/src/app/srs_app_stream_bridge.cpp#L11
2. `src/app/srs_app_stream_bridge.hpp`, there is a `SrsAudioTranscoder` forward declaration, it seems useless, there is no warning, but useless. 
https://github.com/ossrs/srs/blob/6834ec208d67fa47c21536d1f1041bb6d60c1834/trunk/src/app/srs_app_stream_bridge.hpp#L22
3. fix warning:
https://github.com/ossrs/srs/blob/6834ec208d67fa47c21536d1f1041bb6d60c1834/trunk/src/app/srs_app_threads.cpp#L271-L281
   move `buf` declaration into the `SRS_RTC` conditional compiling block.
4. fix warning:
https://github.com/ossrs/srs/blob/6834ec208d67fa47c21536d1f1041bb6d60c1834/trunk/src/app/srs_app_stream_bridge.hpp#L62-L63
    according to the `srs_app_stream_bridge.cpp`, the warning can be fixed by following conditional compiling.
    ```
    #ifdef SRS_RTC
    SrsRtcSource* source_;
    #endif
    #ifdef SRS_FFMPEG_FIT
    SrsRtcRtpBuilder* rtp_builder_;
    #endif
    ```
    But I would rather suggest reconsider the conditional compiling in `srs_app_stream_bridge.cpp`. e.g. 
    ```
    #if defined(SRS_RTC) && defined(SRS_FFMPEG_FIT)
    srs_freep(rtp_builder_);
    #endif
    ```

suzp1984 (commented):
fix compiling error, when rtc is off. 
`./configure -rtc=off; make`

trunk/src/app/srs_app_server.cpp
```diff
@@ -1227,11 +1226,26 @@ srs_error_t SrsServer::do_on_tcp_client(ISrsListener* listener, srs_netfd_t& stf
         }
     }
 
+    // For RTC TCP connection, use resource executor to manage the resource.
+    SrsRtcTcpConn* raw_conn = dynamic_cast<SrsRtcTcpConn*>(resource);
+    if (raw_conn) {
```
suzp1984, line 1232: Be careful the `raw_conn (SrsRtcTcpConn resource)` and its `conn (SrsSharedResource wrapper)`, the `raw_conn` will be add to `conn_manager`, the tcp connection resource manager of `SrsServer`, in line 1244, and its wrapper, a SrsSharedResource, will created below in line 1233, this wrapper didn't added to any resource manager, but it will be released by `_srs_rtc_manager` in `SrsExecutorCoroutine`'s destructor. Then the `raw_conn` stored in `conn_manager` will became a wild pointer, the good news is that the `conn_manager` will never remove it, so the process will never crash, but it's still a problem, is it?

So, there are two problem: 
1. the `raw_conn` will be added to `conn_manager`, which will never remove it;
2. the `raw_conn` inside `conn_manager` will became wild pointer once its wapper, the `SrsSharedResource` released by `_srs_rtc_manager` inside `SrsExecutorCoroutine`'s  desctructor.

suzp1984 (commented):
wild pointer problem of `SrsRtcTcpConn` instance in `conn_manager`, the tcp connection resource manager.

trunk/src/app/srs_app_server.cpp
winlinvip: Fixed

trunk/src/app/srs_app_server.cpp
winlinvip: Fixed

trunk/src/app/srs_app_server.cpp
winlinvip: Nop, manager never adds raw_conn:

```cpp

    // For RTC TCP connection, use resource executor to manage the resource.
    SrsRtcTcpConn* raw_conn = dynamic_cast<SrsRtcTcpConn*>(resource);
    if (raw_conn) {
        // ......
        return err;
    }
```

Won't fix this.

suzp1984: get it.

trunk/src/app/srs_app_rtc_network.cpp
```diff
@@ -436,19 +436,19 @@ srs_error_t SrsRtcUdpNetwork::write(void* buf, size_t size, ssize_t* nwrite)
     return sendonly_skt_->sendto(buf, size, SRS_UTIME_NO_TIMEOUT);
 }
 
-SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* delta)
+SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* delta) : owner_(new SrsRtcTcpConn())
```
suzp1984, line 439: Current branch seems rebased on the #4084 now, which already supported constructor the `SrsSharedResource` instance from a NULL pointer, so I guess this initialization can be improved by:
```suggestion
SrsRtcTcpNetwork::SrsRtcTcpNetwork(SrsRtcConnection* conn, SrsEphemeralDelta* delta) : owner_(NULL)
```

suzp1984: No, reconsider this suggestion and let `SrsSharedResource` accept NULL pointer constructor, because 
https://github.com/ossrs/srs/blob/7b9c52b28357b49028eca82127d9d9c1855d5dd9/trunk/src/app/srs_app_conn.hpp#L163-L165

in this case, it's `owner_->interrupt()`.

PR #4087, opened 2024-06-13
Fix hevc test failures

suzp1984:
Try to fix two blackbox test:
1. TestSlow_SrtPublish_HttpTsPlay_HEVC_Basic: fixed by enlarge the wait time from 3 seconds to 4 before run ffprobe task, which will record the stream by ffmpeg first.
2 TestSlow_SrtPublish_HlsPlay_HEVC_Basic: fixed by wait 16 seconds before run ffprobe task.
About the 2 case: it seems ridiculous to add 16 seconds delay before run ffprobe task. 

> So, I start #4088 to check the github action workflow process, I start this branch from a very earlier version `6.0.113 (srs/core/srs_core_version6.hpp)`, what I found it that, inside `SRS blackbox-test`, the srs version `6.0.128`, the latest version, was printed. That's really wired.

I confirmed this is not the SRS source code's problem, check https://github.com/suzp1984/srs/actions/runs/9511600525/job/26218025559
the srs code 6.0.113 was checkout and running actions based on them, still met same problem.

### DIFF
--- a/trunk/3rdparty/srs-bench/blackbox/hevc_test.go
+++ b/trunk/3rdparty/srs-bench/blackbox/hevc_test.go
@@ -1007,7 +1007,7 @@ func TestSlow_SrtPublish_HlsPlay_HEVC_Basic(t *testing.T) {
 	case <-ctx.Done():
 		r2 = fmt.Errorf("timeout")
 		return
-	case <-time.After(15 * time.Second):
+	case <-time.After(20 * time.Second):
 	}
 
 	// Start FFprobe to detect and verify stream.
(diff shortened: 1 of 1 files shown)


trunk/3rdparty/srs-bench/blackbox/hevc_test.go
```diff
@@ -995,11 +994,14 @@ func TestSlow_SrtPublish_HlsPlay_HEVC_Basic(t *testing.T) {
 		defer wg.Done()
 		<-svr.ReadyCtx().Done()
 
+		// wait for ffmpeg
+		time.Sleep(3 * time.Second)
+
```
suzp1984: No need to sleep here, `ffmpeg` can publish stream immediately after `srs` ready.

PR #4089, opened 2024-06-14
SmartPtr: Support shared ptr for live source. v6.0.129

winlinvip:
Detail change log:

1. [Simple,Refactor] Remove member fields of http entry, etc. https://github.com/ossrs/srs/pull/4089/commits/e34b3d3aa44f56b49c5c0b8e58c0285c4d9094d4
2. [Ignore] Rename source to live_source. https://github.com/ossrs/srs/pull/4089/commits/846f95ec96f183909070c4b3f5ef0e9dd7d5448d
3. [Ignore] Use directly ptr in consumer. https://github.com/ossrs/srs/pull/4089/commits/d38af021ad532ed9b4043c529dd5e54b9cbf5c01
4. [Complex, Important] Use shared ptr for live source. https://github.com/ossrs/srs/pull/4089/commits/88f922413a1fb5cb920ab64f8b3805420932602a 

The object relationship:

live-source

### DIFF
--- a/trunk/src/app/srs_app_rtmp_conn.hpp
+++ b/trunk/src/app/srs_app_rtmp_conn.hpp
@@ -16,6 +16,7 @@
 #include <srs_app_reload.hpp>
 #include <srs_protocol_rtmp_stack.hpp>
 #include <srs_protocol_rtmp_conn.hpp>
+#include <srs_core_autofree.hpp>
 
 class SrsServer;
 class SrsRtmpServer;
@@ -145,14 +146,14 @@ private:
     // The stream(play/publish) service cycle, identify client first.
     virtual srs_error_t stream_service_cycle();
     virtual srs_error_t check_vhost(bool try_default_vhost);
-    virtual srs_error_t playing(SrsLiveSource* source);
-    virtual srs_error_t do_playing(SrsLiveSource* source, SrsLiveConsumer* consumer, SrsQueueRecvThread* trd);
-    virtual srs_error_t publishing(SrsLiveSource* source);
-    virtual srs_error_t do_publishing(SrsLiveSource* source, SrsPublishRecvThread* trd);
-    virtual srs_error_t acquire_publish(SrsLiveSource* source);
-    virtual void release_publish(SrsLiveSource* source);
-    virtual srs_error_t handle_publish_message(SrsLiveSource* source, SrsCommonMessage* msg);
-    virtual srs_error_t process_publish_message(SrsLiveSource* source, SrsCommonMessage* msg);
+    virtual srs_error_t playing(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t do_playing(SrsSharedPtr<SrsLiveSource> source, SrsLiveConsumer* consumer, SrsQueueRecvThread* trd);
+    virtual srs_error_t publishing(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t do_publishing(SrsSharedPtr<SrsLiveSource> source, SrsPublishRecvThread* trd);
+    virtual srs_error_t acquire_publish(SrsSharedPtr<SrsLiveSource> source);
+    virtual void release_publish(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t handle_publish_message(SrsSharedPtr<SrsLiveSource>& source, SrsCommonMessage* msg);
+    virtual srs_error_t process_publish_message(SrsSharedPtr<SrsLiveSource>& source, SrsCommonMessage* msg);
     virtual srs_error_t process_play_control_msg(SrsLiveConsumer* consumer, SrsCommonMessage* msg);
     virtual void set_sock_options();
 private:
--- a/trunk/src/app/srs_app_source.cpp
+++ b/trunk/src/app/srs_app_source.cpp
@@ -1765,11 +1765,15 @@ srs_error_t SrsLiveSourceManager::fetch_or_create(SrsRequest* r, ISrsLiveSourceH
 
     // Use lock to protect coroutine switch.
     // @bug https://github.com/ossrs/srs/issues/1230
-    // TODO: FIXME: Use smaller lock.
+    // TODO: FIXME: Use smaller scope lock.
     SrsLocker(lock);
-    
-    SrsLiveSource* source = NULL;
-    if ((source = fetch(r)) != NULL) {
+
+    string stream_url = r->get_stream_url();
+    std::map<std::string, SrsLiveSource*>::iterator it = pool.find(stream_url);
+
+    if (it != pool.end()) {
+        SrsLiveSource* source = it->second;
+
         // we always update the request of resource,
         // for origin auth is on, the token in request maybe invalid,
         // and we only need to update the token of request, it's simple.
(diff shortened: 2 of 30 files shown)


trunk/src/app/srs_app_source.cpp
```diff
@@ -1778,40 +1782,34 @@ srs_error_t SrsLiveSourceManager::fetch_or_create(SrsRequest* r, ISrsLiveSourceH
...
+    std::map<std::string, SrsLiveSource*>::iterator it = pool.find(stream_url);
+
+    if (it == pool.end()) {
         return NULL;
     }
-    
-    source = pool[stream_url];
-    
+
+    SrsLiveSource* source = it->second;
```
suzp1984: just `return it->second;` seems good.

suzp1984 (approved):

trunk/src/app/srs_app_source.cpp
winlinvip: Seems no big difference. Won't fix.

winlinvip:
See https://github.com/ossrs/srs/discussions/3667#discussioncomment-8969107

trunk/src/app/srs_app_rtmp_conn.hpp
```diff
@@ -145,14 +146,14 @@ class SrsRtmpConn : public ISrsConnection, public ISrsStartable, public ISrsRelo
...
-    virtual void release_publish(SrsLiveSource* source);
-    virtual srs_error_t handle_publish_message(SrsLiveSource* source, SrsCommonMessage* msg);
-    virtual srs_error_t process_publish_message(SrsLiveSource* source, SrsCommonMessage* msg);
+    virtual srs_error_t playing(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t do_playing(SrsSharedPtr<SrsLiveSource> source, SrsLiveConsumer* consumer, SrsQueueRecvThread* trd);
+    virtual srs_error_t publishing(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t do_publishing(SrsSharedPtr<SrsLiveSource> source, SrsPublishRecvThread* trd);
+    virtual srs_error_t acquire_publish(SrsSharedPtr<SrsLiveSource> source);
+    virtual void release_publish(SrsSharedPtr<SrsLiveSource> source);
+    virtual srs_error_t handle_publish_message(SrsSharedPtr<SrsLiveSource>& source, SrsCommonMessage* msg);
```
suzp1984, line 155: why above `api` can not accept a `SrsSharedPtr` reference as input just like this one, `handle_publish_message`? 
use ref as long as possible, seems better.

trunk/src/app/srs_app_source.cpp
```diff
@@ -1769,55 +1770,54 @@ srs_error_t SrsLiveSourceManager::fetch_or_create(SrsRequest* r, ISrsLiveSourceH
...
     // TODO: FIXME: Use smaller scope lock.
     SrsLocker(lock);
     
     string stream_url = r->get_stream_url();
-    std::map<std::string, SrsLiveSource*>::iterator it = pool.find(stream_url);
+    std::map< std::string, SrsSharedPtr<SrsLiveSource> >::iterator it = pool.find(stream_url);
 
     if (it == pool.end()) {
-        return NULL;
+        return SrsSharedPtr<SrsLiveSource>(NULL);
```
suzp1984, line 1809: How about design the fetch api in this way?
 `SrsSharedPtr<SrsLiveSource>& SrsLiveSourceManager::fetch(SrsRequest* r)`, return a `SrsSharedPtr` reference, and Introduce a globally `SrsSharedPtr<SrsLiveSource> _SHARED_NULL_LIVE_SOURCE(NULL)`, return it if the null shared ptr is needed.
 
And for `fetch_or_create`, return the `SrsSharedPtr` reference instead of though the reference parameter: `SrsSharedPtr<SrsLiveSource>& SrsLiveSourceManager::fetch_or_create(SrsRequest* r, ISrsLiveSourceHandler* h, srs_error_t& err)`

trunk/src/app/srs_app_source.cpp
```diff
@@ -1769,55 +1770,54 @@ srs_error_t SrsLiveSourceManager::fetch_or_create(SrsRequest* r, ISrsLiveSourceH
...
+SrsSharedPtr<SrsLiveSource> SrsLiveSourceManager::fetch(SrsRequest* r)
 {
     // Use lock to protect coroutine switch.
     // @bug https://github.com/ossrs/srs/issues/1230
     // TODO: FIXME: Use smaller scope lock.
     SrsLocker(lock);
     
     string stream_url = r->get_stream_url();
-    std::map<std::string, SrsLiveSource*>::iterator it = pool.find(stream_url);
+    std::map< std::string, SrsSharedPtr<SrsLiveSource> >::iterator it = pool.find(stream_url);
```
suzp1984, line 1806: How many `SrsLiveSource`s do SRS supported in a single node? there are no limitations, just restricted by memory and network bandwidth. So why not use `std:unordered_map` to replace this `std::map`, it could be faster in this find operation.

trunk/src/app/srs_app_source.cpp
```diff
@@ -1993,7 +1980,7 @@ bool SrsLiveSource::publisher_is_idle_for(srs_utime_t timeout)
     return false;
 }
 
-srs_error_t SrsLiveSource::initialize(SrsRequest* r, ISrsLiveSourceHandler* h)
+srs_error_t SrsLiveSource::initialize(SrsSharedPtr<SrsLiveSource> wrapper, SrsRequest* r, ISrsLiveSourceHandler* h)
```
suzp1984, line 1983: this method and every initialize embedded this is method can be declared to just accept a `SrsSharedPtr` reference. It will save a lot of `SrsSharedPtr` construction and destruction operations.

trunk/src/app/srs_app_rtmp_conn.hpp
winlinvip: We should use instance like `SrsSharedPtr<SrsLiveSource> source` for almost all use scenarios, but for some special and frequently used api like handle_publish_message we should not use instance, instead we should use pointer like `SrsSharedPtr<SrsLiveSource>* source` or reference like `SrsSharedPtr<SrsLiveSource>& source`.

Won't fix.

trunk/src/app/srs_app_source.cpp
winlinvip: For `fetch_or_create` it may return error, so we should use `**pps` or `& pps` as output parameter.

For `fetch`, a global NULL instance or temporary instance is ok, because it's not a hot path.

Won't fix.

trunk/src/app/srs_app_source.cpp
winlinvip: For performance improvement, please file another PR and please provide benchmark data.

Won't fix by this PR.

trunk/src/app/srs_app_source.cpp
winlinvip: I have claimed for many times: If you want to improve performance, please provide benchmark data. Most of performance improvement does not work. Never fix problem that does not exsit.

For this case, `initialize` is not hot path, so there is no performance bottleneck. We should never use reference unless we really need it.

Won't fix.

Issue #1222, opened 2018-09-12
Does it possible use env variable on conf 


jhaoheng:
as title

Thanks

winlinvip:
Why use env varables?

matclayton:
This would be very useful for us as well. By using Env var in the conf files we can change the configuration from outside the pod. For example we would like to be able to do, which right now we dynamically change in a script before starting srs 

`on_publish ${CALLBACK_URL}/_internal/live/on_publish/?internal_key=${INTERNAL_KEY};`

daxaxelrod:
+1

winlinvip:
Make sense.

winlinvip:
SRS 5.0 supports config with ENV, see #2277

irg1008:
Is this anywhere in the docs? It's so difficult to find something out with this lib... hm

irg1008:
As fare as that PR goes we can't still set http_hooks env vars

suzp1984:
@irg1008 which http_hooks env vars didn't works as you expected? Could you gives the cmd args or steps to reproduce?
As I know all the configs are documented in `conf/full.conf`, which can be the config guideline, for `http_hooks` section, `on_connect` and `on_close` are missing.

irg1008:
> @irg1008 which http_hooks env vars didn't works as you expected? Could you gives the cmd args or steps to reproduce?
> 
> As I know all the configs are documented in `conf/full.conf`, which can be the config guideline, for `http_hooks` section, `on_connect` and `on_close` are missing.

Yeah I looked at conf file but didn't find any env vars for http hooks

suzp1984:
> > @irg1008 which http_hooks env vars didn't works as you expected? Could you gives the cmd args or steps to reproduce?
> > As I know all the configs are documented in `conf/full.conf`, which can be the config guideline, for `http_hooks` section, `on_connect` and `on_close` are missing.
> 
> Yeah I looked at conf file but didn't find any env vars for http hooks

https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/conf/full.conf#L1587-L1588

env **_SRS_VHOST_HTTP_HOOKS_ON_PUBLISH_**

PR #4092, opened 2024-06-16
Config: Improve env config to support multi values. v7.0.2

suzp1984:
1. add on_connect & on_close directives to conf/full.conf;
2. let http_hooks env overwrite support multi values; e.g. SRS_VHOST_HTTP_HOOKS_ON_CONNECT="http://127.0.0.1/api/connect http://localhost/api/connect"

related to https://github.com/ossrs/srs/issues/1222#issuecomment-2170424703
Above comments said `http_hook` env may not works as expected, as I found there are still has some issue in `http_hooks` env configuration, but this PR may not target above problem.

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -7,6 +7,7 @@ The changelog for SRS.
 <a name="v7-changes"></a>
 
 ## SRS 7.0 Changelog
+* v7.0, 2024-08-13, Merge [#4092](https://github.com/ossrs/srs/pull/4092): Config: Improve env config to support multi values. v7.0.2 (#4092)
 * v7.0, 2024-08-12, Merge [#4131](https://github.com/ossrs/srs/pull/4131): Transcode: More generic h264/h265 codec support. v7.0.1 (#4131)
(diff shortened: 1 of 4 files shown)


trunk/conf/full.conf
```diff
@@ -1568,6 +1568,44 @@ vhost hooks.callback.srs.com {
         # Overwrite by env SRS_VHOST_HTTP_HOOKS_ENABLED for all vhosts.
         # default off.
         enabled on;
+        # when client(encoder) connect to vhost/app/stream, call the hook,
```
winlinvip: We should hide this config, because `connect` and `close` is only supported by RTMP protocol, no such callback for RTC/SRT/HLS etc.

So it's not recommended to use `connect` and `close`, instead user should use `publish` and `play` instead`.

To be compatible with previous systems, we should support `connect` and `close` in code, but only remove it from wiki and docs. We may remove the code in 5 years later, when no user use this feature.

BTW: Please never attach some other code that is not related to this PR. This will make the code review very difficult and hard.

suzp1984: removed this deprecated directives.

winlinvip (approved):

PR #4097, opened 2024-06-19
SmartPtr: Support load test for source by srs-bench. v6.0.130

### LINKED ISSUE #4093: SrsLiveSource cleanup is different to SrsSrtSource
Release v6.0-d5

The original code:
```c++
void SrsLiveSource::on_consumer_destroy(SrsLiveConsumer* consumer)
{
    ...
    if (consumers.empty()) {
        play_edge->on_all_client_stop();

        // For edge server, the stream die when the last player quit, because the edge stream is created by player
        // activities, so it should die when all players quit.
        if (_srs_config->get_vhost_is_edge(req->vhost)) {
            stream_die_at_ = srs_get_system_time();
        }

        // When no players, the publisher is idle now.
        publisher_idle_at_ = srs_get_system_time();
    }
}
```

Code that should be changed,  Similar to SrsSrtSource::on_consumer_destroy():
```c++
void SrsLiveSource::on_consumer_destroy(SrsLiveConsumer* consumer)
{
    ...
    if (consumers.empty()) {
        play_edge->on_all_client_stop();

        // For edge server, the stream die when the last player quit, because the edge stream is created by player
        // activities, so it should die when all players quit.
        if (_srs_config->get_vhost_is_edge(req->vhost)) {
            stream_die_at_ = srs_get_system_time();
        } else if (_can_publish) {  // should die when no publishers and consumers.
            stream_die_at_ = srs_get_system_time();
        }

        // When no players, the publisher is idle now.
        publisher_idle_at_ = srs_get_system_time();
    }
}
```

### LINKED ISSUE #4096: Source: The memory leak seems still exist
The memory issue with SRS still appears to persist. I use RTMP to push streams to SRS, and then play them using WebRTC. After playing for a while, I close the stream, then push and play again, only to close it once more.
Version: 6.0 develop branch    2024.6.18
this is my test result:
unit is KB
Startup memory: SRS 19032.
datetime    play 16 streams    close all streams 
13:58         77624                  53152
14:03         79229                  79220
14:06         95260                  95260
14:13         96260                  96260
14:18         97256                  97256
14:28         98428                  98428
14:35        99500                   99500
14:42        100768                 100768 
14:53        101780                 101900
14:59        104476   104476
15:33        107744     107744

`TRANS_BY_GPT4`

winlinvip:
1. Add live benchmark support in srs-bench, which only connects and disconnects without any media transport, to test source creation and disposal and verify source memory leaks.
2. SmartPtr: Support cleanup of HTTP-FLV stream. Unregister the HTTP-FLV handler for the pattern and clean up the objects and resources.
3. Support benchmarking RTMP/SRT with srs-bench by integrating the gosrt and oryx RTMP libraries.
4. Refine SRT and RTC sources by using a timer to clean up the sources, following the same strategy as the Live source.

---------


trunk/src/app/srs_app_dash.cpp
```diff
@@ -734,6 +738,14 @@ srs_error_t SrsDash::cycle()
     return err;
 }
 
+srs_utime_t SrsDash::cleanup_delay()
+{
+    if (!enabled) {
+        return 0;
+    }
+    return _srs_config->get_dash_dispose(req->vhost) * 1.1;
```
suzp1984, line 748: `dash_dispose` is not illustrated in `full.conf`, which can be used as the doc. It's better to add it to `full.conf`.

Another problem, the default `dash_dispose` config is 0, or user can config the `[hls | dash]_dispose` to 0 second, then this `cleanup_delay` will be 0, is the cleanup_delay has to be keep > 0 ? In this case:
`_srs_config->get_dash_dispose(req->vhost) > 0 ? _srs_config->get_dash_dispose(req->vhost) * 1.1 : 1`
Let the `cleanup_delay` at least 1 second.

trunk/src/app/srs_app_source.cpp
```diff
@@ -910,6 +910,15 @@ bool SrsOriginHub::active()
     return is_active;
 }
 
+srs_utime_t SrsOriginHub::cleanup_delay()
+{
+    srs_utime_t delay = hls->cleanup_delay();
+    if (delay > 0) {
+        return delay;
+    }
+    return dash->cleanup_delay();
```
suzp1984: why the `cleanup_delay` of `SrsOriginHub` is the max of `hls` and `dash`?
```suggestion
    return std::max(dash->cleanup_delay(), hls->cleanup_delay);
```

Another puzzle, the default cleanup_delay of `hls` is 120 * 1.1 seconds = 132 seconds ~ 2 mins, while the `dash`'s cleanup_delay is 0 seconds, is that normal?

trunk/src/app/srs_app_dash.cpp
winlinvip: Add dash_dispose to full.conf.
If user config dispose to 0, this cleanup_delay will return 0, which will be ignored.
It works as expectation. Won't fix.

trunk/src/app/srs_app_source.cpp
winlinvip: Fixed. Note that we use `srs_max` instead.

trunk/conf/full.conf
```diff
@@ -1758,6 +1758,21 @@ vhost dash.srs.com {
...
+        # Overwrite by env SRS_VHOST_DASH_DASH_CLEANUP for all vhosts.
+        # default: on
+        dash_cleanup on;
+        # If there is no incoming packets, dispose DASH in this timeout in seconds,
+        # which removes all DASH files including m3u8 and ts files.
+        # @remark 0 to disable dispose for publisher.
+        # @remark apply for publisher timeout only, while "etc/init.d/srs stop" always dispose DASH.
+        # Overwrite by env SRS_VHOST_DASH_DASH_DISPOSE for all vhosts.
+        # default: 120
+        dash_dispose 120;
```
suzp1984, line 1775: https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/src/app/srs_app_config.cpp#L6862-L6866

the default value in the code need to match the value here, be careful, change the default value in the `srs_app_config` may has impact.

winlinvip: The previous default value is 0, but hls dispose defaults to 120. I think it's ok to change to the same.

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -754,6 +752,32 @@ srs_error_t SrsHttpServeMux::handle(std::string pattern, ISrsHttpHandler* handle
...
+
+            // We don't free the handler, because user should free it.
+            if (entry->handler != handler) {
+                srs_freep(entry);
+            }
+        }
+    }
+
+    std::string vhost = pattern;
+    if (pattern.at(0) != '/') {
```
suzp1984, line 771: `vhosts` store the `vhost` -> `ISrsHttpHandler` map, but the value, `ISrsHttpHandler` never used, so is is ok to downgrade the map -> `set`?
About the code to extract the `vhost`, they are repeated in:
https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/src/protocol/srs_protocol_http_stack.cpp#L706-L712

Maybe a another method to get the vhost from pattern.

trunk/src/protocol/srs_protocol_http_stack.cpp
```diff
@@ -691,14 +691,12 @@ srs_error_t SrsHttpServeMux::handle(std::string pattern, ISrsHttpHandler* handle
     srs_assert(handler);
     
     if (pattern.empty()) {
-        srs_freep(handler);
```
suzp1984, line 694: The `ISrsHttpHander` memory release still not perfect.
For a normal process, the `ISrsHttpHandler` is created in:
https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/src/app/srs_app_http_stream.cpp#L991
and released in 
https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/src/protocol/srs_protocol_http_stack.cpp#L624-L627

I would think if it's possible to manager it in same place, `SrsLiveEntry` in this case. I see the auto release in `SrsHttpStreamServer::http_unmount(SrsRequest* r)`, the `ISrsHttpHander`, or `SrsLiveStream`, seems freed twice in `SrsHttpMuxEntry` destructor and `
SrsHttpStreamServer::http_unmount`

And For exception cases, like here, `handler` not freed here, but this exception not handled well, the `handler` not released below at least.
https://github.com/ossrs/srs/blob/e3d74fb04504cce64f78119627efe3aae825dda3/trunk/src/app/srs_app_http_stream.cpp#L1008-L1010

trunk/src/app/srs_app_dash.cpp
suzp1984: I didn't notice the `dash_dispose` default value in `srs_app_config.cpp` already changed to 120 to match the `full.conf`. No problem here.

trunk/src/app/srs_app_http_stream.cpp
```diff
@@ -1032,13 +1056,40 @@ srs_error_t SrsHttpStreamServer::http_mount(SrsRequest* r)
 void SrsHttpStreamServer::http_unmount(SrsRequest* r)
 {
     std::string sid = r->get_stream_url();
-    
-    if (sflvs.find(sid) == sflvs.end()) {
+
+    std::map<std::string, SrsLiveEntry*>::iterator it = sflvs.find(sid);
```
suzp1984: About the name of `sflvs` and `tflvs`, is there any short-terms that I didn't know?, it store `FLV/TS/MP3/AAC` streams, but name it as `http-live-stream`, short as `hls`, seems a conflict with apple `HLS`, is there any better ideas? or just a mis-concept in my mind.

trunk/src/app/srs_app_http_stream.cpp
```diff
@@ -1032,13 +1056,40 @@ srs_error_t SrsHttpStreamServer::http_mount(SrsRequest* r)
...
-    SrsLiveEntry* entry = sflvs[sid];
-    entry->stream->entry->enabled = false;
+
+    // Free all HTTP resources.
+    SrsLiveEntry* entry = it->second;
+    SrsAutoFree(SrsLiveEntry, entry);
+    sflvs.erase(it);
+
+    SrsLiveStream* stream = entry->stream;
+    SrsAutoFree(SrsLiveStream, stream);
```
suzp1984, line 1071: the `SrsAutoFree` not good enough here, when `stream` being released, `stream = NULL`, but the entry->stream still hold an not NULL pointer, but it's memory already released.

So is it a good idea to let the `SrsLiveEntry` destructor to free `stream`, `cache`?

trunk/conf/full.conf
suzp1984: I didn't notice the dash_dispose default value in srs_app_config.cpp already changed to 120 to match the full.conf. No problem here.

trunk/src/app/srs_app_hls.cpp
```diff
@@ -1272,6 +1276,14 @@ srs_error_t SrsHls::cycle()
     return err;
 }
 
+srs_utime_t SrsHls::cleanup_delay()
+{
+    if (!enabled) {
+        return 0;
+    }
+    return _srs_config->get_hls_dispose(req->vhost) * 1.1;
```
duiniuluantanqin, line 1286: What does the magic number 1.1 mean?

trunk/src/app/srs_app_http_stream.cpp
duiniuluantanqin:
```
    // The http live streaming template, to create streams.
    std::map<std::string, SrsLiveEntry*> tflvs;
    // The http live streaming streams, crote by template.
    std::map<std::string, SrsLiveEntry*> sflvs;
```
Perhaps this comment can answer your question?

trunk/src/protocol/srs_protocol_http_stack.cpp
winlinvip: Need another PR to refine the code, because it not about the smart ptr.

Won't fix in this PR.

trunk/src/protocol/srs_protocol_http_stack.cpp
winlinvip: First, in http_unmount, won't free it twice, please check it carefully.

Second, if mux.handle failed, it won't free the stream, but the http_unmount will do.

So I think it's ok. Won't fix.

trunk/src/app/srs_app_http_stream.cpp
winlinvip: Accept. Rename tflvs to templateHandlers, sflvs to streamHandlers.

trunk/src/app/srs_app_http_stream.cpp
winlinvip: I think it's ok, because the entry is removed and freed:

```cpp
    SrsLiveEntry* entry = it->second;
    SrsAutoFree(SrsLiveEntry, entry);
    streamHandlers.erase(it);

    SrsLiveStream* stream = entry->stream;
    SrsAutoFree(SrsLiveStream, stream);

    SrsBufferCache* cache = entry->cache;
    SrsAutoFree(SrsBufferCache, cache);
```

Won't fix.

trunk/src/app/srs_app_hls.cpp
winlinvip: Add comments:

```cpp
    // We use larger timeout to cleanup the HLS, after disposed it if required.
    return _srs_config->get_hls_dispose(req->vhost) * 1.1;
```

suzp1984 (approved):

Issue #3369, opened 2023-01-09
GB28181 push stream failed, SRS 5008 error


Doscene:
National standard flow push failed, log prompt:
```err: code=5008(RtcStun)(RTC do STUN or ICE failed) ```

The INVITE request message is as follows:
```INVITE sip:51190200041320000008@192.168.0.101:5860;transport=udp SIP/2.0
Call-ID: e7634e25054f61ebbe6adf528baf7c16@192.168.0.104
CSeq: 1527 INVITE
From: <sip:51190200042000000001@5119020004:5060>;tag=1673234007732
To: <sip:51190200041320000008@192.168.0.101:5860>
Via: SIP/2.0/UDP 192.168.0.104:5060;branch=z9hG4bK1673234007732
Max-Forwards: 70
User-Agent: ldscloud V1.0.1
Content-Encoding: GB2312
Subject: 51190200041320000008:0902000001,192.168.0.104:0
Content-Type: Application/SDP
Content-Length: 206

v=0
o=51190200041320000008 0 0 IN 192.168.0.104
s=Play
c=IN IP4 192.168.0.106
t=0 0
m=video 8000 RTP/AVP 96 98 97
a=recvonly
a=rtpmap:96 PS/90000
a=rtpmap:98 H264/90000
a=rtpmap:97 MPEG4/90000
y=0902000001```
```
1. SRS Version:  5.0.128
```
```
2. SRS Log:

```
[2023-01-09 03:09:01.284][INFO][1][963c6429] Process: cpu=1.00%,49MB, threads=2
[2023-01-09 03:09:02.205][WARN][1][m173ps86][11] handle udp pkt, count=1/1, err: code=5008(RtcStun)(RTC do STUN or ICE failed) : size=1172, data=[80 60 00 01 00 20 ba bd] : no session, peer_id=192.168.0.101:12002, fast=-2148779970514605888
thread [1][m173ps86]: cycle() [./src/app/srs_app_listener.cpp:743][errno=11]
thread [1][m173ps86]: on_udp_packet() [./src/app/srs_app_rtc_server.cpp:430][errno=11]
[2023-01-09 03:09:02.205][INFO][1][m173ps86] <- RTC RECV #15, udp 1, pps 0/0, schedule 1
[2023-01-09 03:09:03.553][INFO][1][84yn0d8f] Hybrid cpu=2.99%,50MB, cid=72,1, timer=58,0,0, clock=0,14,31,0,0,0,0,0,0
[...]

PR #4101, opened 2024-06-21
GB28181: Support external SIP server. v6.0.144

duiniuluantanqin:
For #3369 to support an external powerful SIP server, do not use the embedded SIP server of SRS.
For more information, detailed steps, system architecture, and background explanation, please see https://ossrs.net/lts/zh-cn/docs/v6/doc/gb28181#external-sip

---------

### DIFF
--- a/trunk/src/app/srs_app_gb28181.cpp
+++ b/trunk/src/app/srs_app_gb28181.cpp
@@ -22,6 +22,10 @@
 #include <srs_app_pithy_print.hpp>
 #include <srs_app_rtmp_conn.hpp>
 #include <srs_protocol_raw_avc.hpp>
+#include <srs_app_server.hpp>
+#include <srs_protocol_json.hpp>
+#include <srs_app_http_api.hpp>
+#include <srs_app_statistic.hpp>
 
 #include <sstream>
 using namespace std;
@@ -421,11 +425,11 @@ srs_error_t SrsGbListener::initialize(SrsConfDirective* conf)
 
     bool sip_enabled = _srs_config->get_stream_caster_sip_enable(conf);
     if (!sip_enabled) {
-        return srs_error_new(ERROR_GB_CONFIG, "GB SIP is required");
+        srs_warn("GB SIP is disabled.");
+    } else {
+        int port = _srs_config->get_stream_caster_sip_listen(conf);
+        sip_listener_->set_endpoint(ip, port)->set_label("SIP-TCP");
     }
-
-    int port = _srs_config->get_stream_caster_sip_listen(conf);
-    sip_listener_->set_endpoint(ip, port)->set_label("SIP-TCP");
 
     return err;
 }
@@ -440,6 +444,24 @@ srs_error_t SrsGbListener::listen()
 
     if ((err = sip_listener_->listen()) != srs_success) {
         return srs_error_wrap(err, "listen");
+    }
+
+    if ((err = listen_api()) != srs_success) {
+        return srs_error_wrap(err, "listen api");
+    }
+
+    return err;
+}
+
+srs_error_t SrsGbListener::listen_api()
+{
+    srs_error_t err = srs_success;
+
+    // TODO: FIXME: Fetch api from hybrid manager, not from SRS.
+    ISrsHttpServeMux* http_api_mux = _srs_hybrid->srs()->instance()->api_server();
+
+    if ((err = http_api_mux->handle("/gb/v1/publish/", new SrsGoApiGbPublish(conf_))) != srs_success) {
+        return srs_error_wrap(err, "handle publish");
     }
 
     return err;
@@ -550,11 +572,16 @@ std::string SrsGbSipTcpConn::device_id()
     return register_->device_id();
 }
 
+void SrsGbSipTcpConn::set_device_id(const std::string &id)
+{
+    register_->from_address_user_ = id;
+}
+
 void SrsGbSipTcpConn::set_cid(const SrsContextId& cid)
 {
     if (owner_cid_) owner_cid_->set_cid(cid);
-    receiver_->set_cid(cid);
-    sender_->set_cid(cid);
+    if (receiver_) receiver_->set_cid(cid);
+    if (sender_) sender_->set_cid(cid);
     cid_ = cid;
 }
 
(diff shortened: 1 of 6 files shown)


trunk/src/app/srs_app_gb28181.hpp
```diff
@@ -234,6 +263,10 @@ class SrsGbSipTcpConn : public ISrsResource, public ISrsCoroutineHandler, public
 public:
     // Get the SIP device id.
     std::string device_id();
+    // For use with external SIP signaling server ONLY
+    // When using an external SIP signaling server, device id are not available, so manual configuration is required
+    // This id will be used as the stream name in the RTMP protocol
+    void set_device_id(const std::string & id);
```
suzp1984: `const std::string & id` -> `const std::string& id`, no more whitespace.

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2684,5 +2711,121 @@ void srs_sip_parse_address(const std::string& address, std::string& user, std::s
...
+        res->set("code", SrsJsonAny::integer(srs_error_code(err)));
+        res->set("desc", SrsJsonAny::str(srs_error_code_str(err).c_str()));
+        srs_freep(err);
+        return srs_api_response(w, r, res->dumps());
+    }
+
+    return srs_api_response(w, r, res->dumps());
+}
+
+srs_error_t SrsGoApiGbPublish::do_serve_http(ISrsHttpResponseWriter *w, ISrsHttpMessage *r, SrsJsonObject *res)
```
suzp1984: consistency pointer format: `ISrsHttpResponseWriter *w` -> `ISrsHttpResponseWriter* w`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2684,5 +2711,121 @@ void srs_sip_parse_address(const std::string& address, std::string& user, std::s
...
+        return srs_error_new(ERROR_SYSTEM_STREAM_BUSY, "stream already exists");
+    }
+
+    session = dynamic_cast<SrsSharedResource<SrsGbSession>*>(_srs_gb_manager->find_by_fast_id(ssrc));
+    if (session) {
+        return srs_error_new(ERROR_SYSTEM_STREAM_BUSY, "ssrc already exists");
+    }
+
+    SrsGbSession* raw_session = NULL;
+    if (!session) {
```
suzp1984: no need to check `session` is empty, because if it's not empty, error already returned, check above.

trunk/src/app/srs_app_gb28181.hpp
```diff
@@ -187,6 +215,7 @@ class SrsGbListener : public ISrsListener, public ISrsTcpHandler
 public:
     srs_error_t initialize(SrsConfDirective* conf);
     srs_error_t listen();
+    srs_error_t listen_api();
```
suzp1984: is the `private` access scope much suitable for `listen_api`?

suzp1984:
https://ossrs.net/lts/zh-cn/docs/v5/doc/gb28181#usage
the doc need to updated, as I found, gb28181-2022 also supported by srs. GB28181-2022 is also the only option in my camera.

trunk/src/app/srs_app_gb28181.hpp
duiniuluantanqin: ok

trunk/src/app/srs_app_gb28181.cpp
duiniuluantanqin: ok

trunk/src/app/srs_app_gb28181.cpp
duiniuluantanqin: ok

trunk/src/app/srs_app_gb28181.hpp
duiniuluantanqin: Very attentive, thank you

duiniuluantanqin:
> https://ossrs.net/lts/zh-cn/docs/v5/doc/gb28181#usage the doc need to updated, as I found, gb28181-2022 also supported by srs. GB28181-2022 is also the only option in my camera.

Yeah, this is the work I am currently doing, and I will submit the changes as soon as possible.

Update: https://github.com/ossrs/srs-docs/pull/67

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -22,6 +22,10 @@
 #include <srs_app_pithy_print.hpp>
 #include <srs_app_rtmp_conn.hpp>
 #include <srs_protocol_raw_avc.hpp>
+#include <srs_app_server.hpp>
+#include <srs_protocol_json.hpp>
+#include <srs_app_http_api.hpp>
+#include <srs_app_statistic.hpp>
```
suzp1984, line 28: Is this include, `srs_app_statistice.hpp`, a useless one?

duiniuluantanqin: It's not being used here at the moment, but it will be needed for adding statistical functionality later on. Let's keep it for now.

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -2684,5 +2711,118 @@ void srs_sip_parse_address(const std::string& address, std::string& user, std::s
...
+    int port = _srs_config->get_stream_caster_listen(conf_);
+    res->set("port", SrsJsonAny::integer(port));
+    res->set("is_tcp", SrsJsonAny::boolean(true)); // only tcp supported
+
+    srs_trace("GB publish id: %s, ssrc=%lu", id.c_str(), ssrc);
+
+    return err;
+}
+
+srs_error_t SrsGoApiGbPublish::bind_session(std::string id, uint64_t ssrc)
```
suzp1984, line 2788: This is a suggestion, is it possible to refactor this, `SrsGoApiGbPublish::bind_session` and `SrsGbMediaTcpConn::bind_session`, to one service object, whose duty is to `bind_session`? While the benefit is that just one piece of code to manager the `session`. This is maybe a big refactor, let gb28181 support both the internal sip and external sip(http api) by a more explicit way, let sip do just the message transport duty, the session is managed by another object.

duiniuluantanqin: Maybe we can add function `fetch_or_create` for `_srs_gb_manager`. But it`s a big refator as you say.
@winlinvip Please help analyze whether it is feasible?

duiniuluantanqin: Let's do this: I'll raise a separate PR to address this issue.

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -421,12 +425,12 @@ srs_error_t SrsGbListener::initialize(SrsConfDirective* conf)
 
     bool sip_enabled = _srs_config->get_stream_caster_sip_enable(conf);
     if (!sip_enabled) {
-        return srs_error_new(ERROR_GB_CONFIG, "GB SIP is required");
+        srs_warn("GB SIP is disabled.");
```
suzp1984, line 427: one more thing to review. About the conf directives of `sip.enable`:
https://github.com/ossrs/srs/blob/ea7e2c28491ae61cc1b7aaa207ce38a74fa36487/trunk/conf/full.conf#L681-L683

The code here means: `GB internal SIP is disabled`, so don't break the GB process, SRS can still support the external SIP. 

So, there are a few thing to consider:
1. improve the log: `GB [internal|embedded] SIP is disabled`;
2. Internal SIP is no longer a necessary one, so use `srs_trace` to replace `srs_warn`;
3. Maybe rename `sip_enable` to `internal_sip_enabled`;
4. External sip is the future, so no need to enable or disable it?

duiniuluantanqin: In essence, the external SIP server is not part of SRS. You can look at my implementation, even use it directly, but it's not mandatory. So, I don't want to modify it for now.

winlinvip (approved):

PR #4109, opened 2024-07-08
UniquePtr: Support SrsUniquePtr to replace SrsAutoFree. v6.0.136

winlinvip:
To manage an object:

```cpp
// Before
MyClass* ptr = new MyClass();
SrsAutoFree(MyClass, ptr);
ptr->do_something();

// Now
SrsUniquePtr<MyClass> ptr(new MyClass());
ptr->do_something();
```

To manage an array of objects:

```cpp
// Before
char* ptr = new char[10];
SrsAutoFreeA(char, ptr);
ptr[0] = 0xf;

// Now
SrsUniquePtr<char[]> ptr(new char[10]);
ptr[0] = 0xf;
```

In fact, SrsUniquePtr is a limited subset of SrsAutoFree, mainly managing pointers and arrays. SrsUniquePtr is better than SrsAutoFree because it has the same API to standard unique ptr.

```cpp
SrsUniquePtr<MyClass> ptr(new MyClass());
ptr->do_something();
MyClass* p = ptr.get();
```

SrsAutoFree actually uses a pointer to a pointer, so it can be set to NULL, allowing the pointer's value to be changed later (this usage is different from SrsUniquePtr).

```cpp
// OK to free ptr correctly.
MyClass* ptr;
SrsAutoFree(MyClass, ptr);
ptr = new MyClass();

// Crash because ptr is an invalid pointer.
MyClass* ptr;
SrsUniquePtr<MyClass> ptr(ptr);
ptr = new MyClass();
```

Additionally, SrsAutoFreeH can use specific release functions, which SrsUniquePtr does not support.

---------


trunk/src/utest/srs_utest_protocol.cpp
```diff
@@ -3355,14 +3355,14 @@ VOID TEST(ProtocolStackTest, ProtocolAckSizeFlow)
     }
     // recv auto send acked size. #1
     if (true) {
-        SrsCommonMessage* msg = NULL; SrsAutoFree(SrsCommonMessage, msg);
+        SrsCommonMessage* msg = NULL; SrsUniquePtr<SrsCommonMessage> msg_uptr(msg);
```
suzp1984: memory leak warning in `srs utest`, check the `utest-regression-blackbox-test`.

```suggestion
        SrsCommonMessage* msg = NULL;
        HELPER_ASSERT_SUCCESS(proto.recv_message(&msg));
        SrsUniquePtr<SrsCommonMessage> msg_uptr(msg)
        
```

winlinvip: Fixed.

trunk/src/core/srs_core_deprecated.hpp
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright (c) 2013-2024 The SRS Authors
+//
+// SPDX-License-Identifier: MIT
+//
+
+#ifndef SRS_CORE_DEPRECATED_HPP
+#define SRS_CORE_DEPRECATED_HPP
+
+#include <srs_core.hpp>
```
suzp1984, line 10: this include header,`srs_core.hpp`, is useless, can be removed.

winlinvip: I want to make sure the first included file is this one.

Won't fix.

trunk/src/core/srs_core_deprecated.cpp
```diff
@@ -0,0 +1,8 @@
+//
+// Copyright (c) 2013-2024 The SRS Authors
+//
+// SPDX-License-Identifier: MIT
+//
+
+#include <srs_core_deprecated.hpp>
```
suzp1984, line 7: I don't know whether it's a code conventions or not, but an empty cpp, seems useless.

winlinvip: 我花时间确认了下：这个空文件是有作用的，configure脚本依赖它来生成编译对象，如果去掉空文件，那么修改这个头文件后，也不会导致make重新编译。

此外，对于这种不是特别重要的一些东西，希望能不要追求一种“绝对”和“纯粹”的意图，如果希望这么做，也需要自己花时间尝试和验证。

winlinvip: 这个影响不大，不过删除也可以，我删除了。

winlinvip: It will cause the utest fail. Please file another PR if you want to remove it.

Won't fix.

trunk/src/app/srs_app_http_conn.cpp
```diff
@@ -155,14 +155,13 @@ srs_error_t SrsHttpConn::do_cycle()
         return srs_error_wrap(err, "start");
     }
 
-    SrsRequest* last_req = NULL;
-    SrsAutoFree(SrsRequest, last_req);
-
     // process all http messages.
-    err = process_requests(&last_req);
+    SrsRequest* last_req_raw = NULL;
+    err = process_requests(&last_req_raw);
```
suzp1984, line 160: `err` need to freed and processed if it's not `srs_success`, otherwise `err != srs_success` would be a memory leak.

trunk/src/app/srs_app_recv_thread.cpp
```diff
@@ -577,12 +577,12 @@ srs_error_t SrsHttpRecvThread::cycle()
     srs_error_t err = srs_success;
     
     while ((err = trd->pull()) == srs_success) {
-        ISrsHttpMessage* req = NULL;
-        SrsAutoFree(ISrsHttpMessage, req);
-        
-        if ((err = conn->pop_message(&req)) != srs_success) {
+        ISrsHttpMessage* req_raw = NULL;
+        if ((err = conn->pop_message(&req_raw)) != srs_success) {
```
suzp1984: https://github.com/ossrs/srs/blob/baf22d01c1a0f22ae5739f94664763d7b92b173f/trunk/src/app/srs_app_http_conn.cpp#L338

This method will always sent NULL message out.
Suggest refactor `pop_message` or use `NULL to replace `req_raw`.

```suggestion
        if ((err = conn->pop_message(NULL)) != srs_success) {
```

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -644,17 +644,18 @@ srs_error_t SrsRtcPlayStream::cycle()
...
-    SrsRtcConsumer* consumer = NULL;
-    SrsAutoFree(SrsRtcConsumer, consumer);
-    if ((err = source->create_consumer(consumer)) != srs_success) {
+    SrsRtcConsumer* consumer_raw = NULL;
+    if ((err = source->create_consumer(consumer_raw)) != srs_success) {
         return srs_error_wrap(err, "create consumer, source=%s", req_->get_stream_url().c_str());
     }
 
-    srs_assert(consumer);
+    srs_assert(consumer_raw);
```
suzp1984, line 652: `consumer_raw` will alway has value, no need to check with assert?

https://github.com/ossrs/srs/blob/baf22d01c1a0f22ae5739f94664763d7b92b173f/trunk/src/app/srs_app_rtc_source.cpp#L550-L562

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1266,12 +1263,12 @@ srs_error_t SrsRtcRtpBuilder::package_nalus(SrsSharedPtrMessage* msg, const vect
...
         pkt->header.set_sequence(video_sequence++);
         pkt->header.set_timestamp(msg->timestamp * 90);
-        pkt->set_payload(raw, SrsRtspPacketPayloadTypeNALU);
+        pkt->set_payload(raw_raw, SrsRtspPacketPayloadTypeNALU);
         pkt->wrap(msg);
     } else {
         // We must free it, should never use RTP packets to free it,
         // because more than one RTP packet will refer to it.
-        SrsAutoFree(SrsRtpRawNALUs, raw);
+        SrsUniquePtr<SrsRtpRawNALUs> raw(raw_raw);
```
suzp1984, line 1271: I would rather use a `srs_freep(raw_raw)` manually in this `else` scope to simple the code.

trunk/src/app/srs_app_rtmp_conn.cpp
```diff
@@ -755,25 +755,26 @@ srs_error_t SrsRtmpConn::playing(SrsSharedPtr<SrsLiveSource> source)
     set_sock_options();
     
     // Create a consumer of source.
-    SrsLiveConsumer* consumer = NULL;
-    SrsAutoFree(SrsLiveConsumer, consumer);
-    if ((err = source->create_consumer(consumer)) != srs_success) {
+    SrsLiveConsumer* consumer_raw = NULL;
+    if ((err = source->create_consumer(consumer_raw)) != srs_success) {
```
suzp1984, line 759: https://github.com/ossrs/srs/blob/baf22d01c1a0f22ae5739f94664763d7b92b173f/trunk/src/app/srs_app_source.cpp#L2662-L2682

This is typical case of a potential memory leak maybe introduced by `SrsUniquePtr` compared with `SrsAutoFree`.

`source->create_consumer` can return not null `consumer_raw` value with a not srs_success err.
In this case, the `consumer_raw` will be a memory leak. 

Two ways to overcome this problem:
```suggestion
    if ((err = source->create_consumer(consumer_raw)) != srs_success) {
        srs_freep(consumer_raw); // free consumer_raw anyway;
```
Or use `SrsAutoFree` in this case.

I would rather always prefer `SrsAutoFree` in any cases.

trunk/src/app/srs_app_srt_conn.cpp
```diff
@@ -487,20 +486,20 @@ srs_error_t SrsMpegtsSrtConn::do_playing()
...
-    SrsSrtConsumer* consumer = NULL;
-    SrsAutoFree(SrsSrtConsumer, consumer);
-    if ((err = srt_source_->create_consumer(consumer)) != srs_success) {
+    SrsSrtConsumer* consumer_raw = NULL;
+    if ((err = srt_source_->create_consumer(consumer_raw)) != srs_success) {
         return srs_error_wrap(err, "create consumer, ts source=%s", req_->get_stream_url().c_str());
     }
-    srs_assert(consumer);
+
+    srs_assert(consumer_raw);
```
suzp1984, line 494: `consumer_raw` always `true` here

trunk/src/kernel/srs_kernel_mp4.cpp
```diff
@@ -5431,15 +5430,14 @@ srs_error_t SrsMp4Decoder::initialize(ISrsReadSeeker* rs)
     off_t offset = -1;
     
     while (true) {
-        SrsMp4Box* box = NULL;
-        SrsAutoFree(SrsMp4Box, box);
-        
-        if ((err = load_next_box(&box, 0)) != srs_success) {
+        SrsMp4Box* box_raw = NULL;
+        if ((err = load_next_box(&box_raw, 0)) != srs_success) {
```
suzp1984, line 5434: This is bug here, `box_raw` is managed by `SrsUniquePtr` here and `SrsAutoFree` inside `load_next_box`, line L5655. so this pointer will be released twice. It's ok for both of them are managed by `SrsAutoFree`, but not this way.

trunk/src/app/srs_app_http_conn.cpp
winlinvip: The err is returned:

```cpp
    err = process_requests(&last_req_raw);
    if ((r0 = on_disconnect(last_req.get())) != srs_success) {
        err = srs_error_wrap(err, "on disconnect %s", srs_error_desc(r0).c_str());
        srs_freep(r0);
    }
    
    return err;
```

Won't fix.

trunk/src/app/srs_app_recv_thread.cpp
winlinvip: Fixed.

trunk/src/app/srs_app_rtc_conn.cpp
winlinvip: Not a bug or logic error.

Won't fix.

trunk/src/app/srs_app_rtc_source.cpp
winlinvip: Nop, it's used later:

```cpp
        SrsUniquePtr<SrsRtpRawNALUs> raw(raw_raw);
        uint8_t header = raw->skip_first_byte();
```

Won't fix.

trunk/src/app/srs_app_rtmp_conn.cpp
winlinvip: If `create_consumer` return an error, then it should make sure the `**consumer` should not be set to any object.

trunk/src/app/srs_app_srt_conn.cpp
winlinvip: Not a bug or an error.

Won't fix.

trunk/src/app/srs_app_rtmp_conn.cpp
suzp1984: the problem is `create_consumer` need to refined, don't create `consumer` before return err.

```suggestion
    srs_error_t SrsLiveSource::create_consumer(SrsLiveConsumer*& consumer) 
 { 
     srs_error_t err = srs_success; 
  
     // for edge, when play edge stream, check the state 
     if (_srs_config->get_vhost_is_edge(req->vhost)) { 
         // notice edge to start for the first client. 
         if ((err = play_edge->on_client_play()) != srs_success) { 
             return srs_error_wrap(err, "play edge"); 
         } 
     } 
     
     // There should be one consumer, so reset the timeout. 
     stream_die_at_ = 0; 
     publisher_idle_at_ = 0; 
     
     consumer = new SrsLiveConsumer(this); 
     consumers.push_back(consumer); 
      
     return err; 
 } 
```

suzp1984 (approved):
all done

trunk/src/app/srs_app_rtmp_conn.cpp
winlinvip: Fixed.

trunk/scripts/copy_to_gits.sh
```diff
@@ -29,3 +34,6 @@ cp -R 3rdparty/srs-bench/* ~/git/srs-bench/ &&
 cp -R 3rdparty/srs-bench/.gitignore ~/git/srs-bench/ &&
 (cd ~/git/srs-bench && git st)
 
+echo "Copy state-threads"
+cp -R 3rdparty/st-srs/* ~/git/state-threads/ &&
+(cd ~/git/state-threads && git st)
```
suzp1984, line 39: what's `git st`? Is it a shortcut?

suzp1984: `git st` == `git status`

winlinvip: Fixed.

PR #4111, opened 2024-07-10
DASH: Fix time unit error for disposing. v6.0.138

suzp1984:
## Cause
dash auto dispose is configured by seconds, but the code compare by usecond, 1 second = 1,000,000 useconds.

releated to #4097
Bug introduced after #4097 supported Dash auto dispose after a timeout without media data.

## How to reproduce

1. `./objs/srs -c conf/dash.conf`
2. publish a rtmp stream.
3. play dash stream. -> no dash stream, always 404 error.

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -7,6 +7,7 @@ The changelog for SRS.
 <a name="v6-changes"></a>
 
 ## SRS 6.0 Changelog
+* v6.0, 2024-07-13, Merge [#4111](https://github.com/ossrs/srs/pull/4111): DASH: Fix time unit error for disposing. v6.0.138 (#4111)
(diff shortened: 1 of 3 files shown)


trunk/src/app/srs_app_dash.cpp
```diff
@@ -720,7 +720,7 @@ srs_error_t SrsDash::cycle()
     if (dash_dispose <= 0) {
         return err;
     }
-    if (srs_get_system_time() - last_update_time_ <= dash_dispose) {
+    if (srs_get_system_time() - last_update_time_ <= dash_dispose * 1000000) {
```
winlinvip: Should convert dash_dispose to us, by multiple SRS_UTIME_SECONDS. Should never use magic number for time object, instead, all time objects should use the same unit such as us.

For example:

```cpp
srs_utime_t SrsConfig::get_dash_dispose(std::string vhost)
{
    static srs_utime_t DEFAULT = 120 * SRS_UTIME_SECONDS;
```

This means `srs_utime_t` is a time object with unit as us.

suzp1984: get it, fixed.

PR #4115, opened 2024-07-11
HLS: Add missing newline to end of session manifest. v6.0.143

Zlo:
The session HLS manifest file lacks a terminating newline in the final line.
This may cause strict players to reject it.


winlinvip:
Thank you for this PR! Is there any specification about the newline? As you know, some player might also reject it when got empty line.

Zlo:
Well, there is the common POSIX rule that a text file should contain lines which should end with a terminating newline (some tools will ignore the final line if it is not terminated) and there is the RFC stating HLS files are UTF-8 text files:
* [definition of a text file](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_403)
* [definition of a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)
* [HLS files are text files](https://datatracker.ietf.org/doc/html/rfc8216#section-2)

We do have embedded players which cannot play it without the newline and I have not come across any player yet that cannot play it when the newline is added, but that obviously isn't any proper proof.

Edit: Oh, actually the RFC mentions this in itself already:
[Definition of a playlist, second alinea](https://datatracker.ietf.org/doc/html/rfc8216#section-4.1)

winlinvip:
Make sense, thank you.

winlinvip (approved):

PR #4120, opened 2024-07-16
Let cluster config support overwritten by ENV

suzp1984:
The env conf overwrite is convenient and necessary for the container (docker & k8s) deployment. So this commit can let SRS support deployed as edge cluster by containers.


winlinvip:

Issue #1784, opened 2020-05-28
SrsHttpApi对象析构时程序出现coredump，srs_close_stfd引起


leiyoacang:
**描述(Description)**
SrsHttpApi对象在析构时会时程序coredump，具体是在调用srs_close_stfd时，触发了assert

> 描述你遇到了什么问题(Please description your issue here)

1. SRS版本(Version): ``

SRS版本：2.0.243
1. SRS的日志如下(Log):
```
core文件信息如下：
#0  0x00007f472098499f in raise () from /lib64/libc.so.6
#1  0x00007f472096ecf5 in abort () from /lib64/libc.so.6
#2  0x00007f472096ebc9 in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#3  0x00007f472097ce56 in __assert_fail () from /lib64/libc.so.6
#4  0x00000000004acbcc in srs_close_stfd (stfd=@0xac0d10: 0xb030a0) at src/app/srs_app_st.cpp:240
#5  0x0000000000474e80 in SrsConnection::dispose (this=0xac0cf0) at src/app/srs_app_conn.cpp:72
#6  0x0000000000474ca3 in SrsConnection::~SrsConnection (this=0xac0cf0, 
    __vtt_parm=0x947b08 <VTT for SrsHttpApi+8>, __in_chrg=<optimized out>) at src/app/srs_app_conn.cpp:55
#7  0x00000000004e52e9 in SrsHttpApi::~SrsHttpApi (this=0xac0cf0, __in_chrg=<optimized out>, 
    __vtt_parm=<optimized out>) at src/app/srs_app_http_api.cpp:2007
#8  0x00000000004e5381 in SrsHttpApi::~SrsHttpApi (this=0xac0cf0, __in_chrg=<optimized out>, 
    __vtt_parm=<optimized out>) at src/app/srs_app_http_api.cpp:2010
#9  0x000000000046ee15 in SrsServer::remove (this=0xabd820, conn=0xac0cf0) at src/app/srs_app_server.cpp:1167
#10 0x00000000004750ba in SrsConnection::on_thread_stop (this=0xac0cf0) at src/app/srs_app_conn.cpp:112
#11 0x00000000004a7ba3 in SrsOneCycleThread::on_thread_stop (this=0xabdc70) at src/app/srs_app_thread.cpp:398
#12 0x00000000004a73d7 in internal::SrsThread::thread_cycle (this=0xb05610) at src/app/srs_app_thread.cpp:240
[...]

Issue #511, opened 2015-10-26
边缘拉流，系统经常出现ret=1018(Device or resource busy) 和 ret=1018(No such file or directory) 错误


gqf2008:
日志
```
[2015-10-26 13:36:08.153][error][13948][3998][16] http post on_play uri failed. client_id=3998, url=http://127.0.0.1/lcs/api/rtmp/on_play/shnh-edge2-live1.evideocloud.net, request={"action":"on_play","client_id":3998,"ip":"222.88.95.177","vhost":"live1.evideocloud.net","app":"live","stream":"dxhdbh__gBML4E6B40Lv","pageUrl":""}, response=, code=0, ret=1018(Device or resource busy)
[2015-10-26 13:36:08.153][error][13948][3998][16] hook client on_play failed. url=http://127.0.0.1/lcs/api/rtmp/on_play/shnh-edge2-live1.evideocloud.net, ret=1018(Device or resource busy)
[2015-10-26 13:36:08.153][error][13948][3998][16] http hook on_play failed. ret=1018(Device or resource busy)
[2015-10-26 13:36:08.153][error][13948][3998][16] stream service cycle failed. ret=1018(Device or resource busy)
[2015-10-26 13:36:08.154][error][13948][3998][2] connect to server error. ip=127.0.0.1, port=80, ret=1018(No such file or directory)
[2015-10-26 13:36:08.342][error][13948][4000][104] rtmp handshake failed. ret=1008(Connection reset by peer)
[2015-10-26 13:36:09.032][error][13948][3990][2] connect to server error. ip=127.0.0.1, port=80, ret=1018(No such file or directory)
[2015-10-26 13:36:09.032][error][13948][3990][2] connect to server error. ip=127.0.0.1, port=80, ret=1018(No such file or directory)
[2015-10-26 13:36:09.251][error][13948][4002][2] connect to server error. ip=127.0.0.1, port=80, ret=1018(No such file or directory)
[2015-10-26 13:36:09.251][error][13948][4002][16] http post on_connect uri failed. client_id=4002, url=http://127.0.0.1/lcs/api/rtmp/on_connect/shnh-edge2-live1.evideocloud.net, request={"action":"on_connect","client_id":4002,"ip":"222.186.130.3","vhost":"live1.evideocloud.net","app":"live","tcUrl":"rtmp://live1.evideocloud.net:1935/live","pageUrl":""}, response=, code=0, ret=1018(Device or resource busy)
[...]

PR #4126, opened 2024-07-23
Edge: Improve stability for state and fd closing. v5.0.214 v6.0.139

winlinvip:
1. Should always stop coroutine before close fd, see #511, #1784
2. When edge forwarder coroutine quit, always set the error code.
3. Do not unpublish if invalid state.

---------

### DIFF
--- a/trunk/src/app/srs_app_edge.cpp
+++ b/trunk/src/app/srs_app_edge.cpp
@@ -782,6 +782,10 @@ srs_error_t SrsEdgeForwarder::start()
         
         url = srs_generate_rtmp_url(server, port, req->host, vhost, req->app, req->stream, req->param);
     }
+
+    // We must stop the coroutine before disposing the sdk.
+    srs_freep(trd);
+    trd = new SrsSTCoroutine("edge-fwr", this, _srs_context->get_id());
     
     // open socket.
     srs_freep(sdk);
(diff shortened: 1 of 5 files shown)


trunk/src/app/srs_app_edge.cpp
```diff
@@ -840,7 +845,13 @@ srs_error_t SrsEdgeForwarder::cycle()
             return srs_error_wrap(err, "thread pull");
         }
 
-        if ((err = do_cycle()) != srs_success) {
+        // If coroutine stopping, we should always set the quit error code.
+        err = do_cycle();
+        if (send_error_code == 0) {
```
suzp1984:
```suggestion
      if ((err = do_cycle()) != srs_success) {
            if (send_error_code == ERROR_SUCCESS) {
                // If coroutine stopping, we should always set the quit error code.
                send_error_code = srs_error_code(err);
            }
            return srs_error_wrap(err, "do cycle");
      }
```

suggest another way.

About `send_error_code`, I believe there is a dead loop:
https://github.com/ossrs/srs/blob/8c034e5a13abf209c7cddfb32ae1d32ce6627f09/trunk/src/app/srs_app_edge.cpp#L864-L884

there are no place to set `send_error_code` = ERROR_SUCCESS in the loop, only set the value when `sdk->recv_message(&msg)`.

suzp1984 (commented):
review `send_error_code` problem.

trunk/src/app/srs_app_edge.cpp
winlinvip: not dead loop, the publish owner will stop it when error.

suzp1984 (approved):

PR #4127, opened 2024-07-24
Transcode: Support video codec such as h264_qsv and libx265. v6.0.145

jb-alvarado:
Currently only libx264 ffmpeg encoder is supported. This pull request add also h264_qsv. But maybe a more generic solution with oder encoders would be useful to.

### DIFF
--- a/trunk/doc/CHANGELOG.md
+++ b/trunk/doc/CHANGELOG.md
@@ -7,6 +7,7 @@ The changelog for SRS.
 <a name="v6-changes"></a>
 
 ## SRS 6.0 Changelog
+* v6.0, 2024-07-27, Merge [#4127](https://github.com/ossrs/srs/pull/4127): Transcode: Support video codec such as h264_qsv and libx265. v6.0.145 (#4127)
 * v6.0, 2024-07-27, Merge [#4101](https://github.com/ossrs/srs/pull/4101): GB28181: Support external SIP server. v6.0.144 (#4101)
 * v6.0, 2024-07-24, Merge [#4115](https://github.com/ossrs/srs/pull/4115): HLS: Add missing newline to end of session manifest. v6.0.143 (#4115)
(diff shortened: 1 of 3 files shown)


winlinvip:
I think it's also reasonable to support libx265?

jb-alvarado:
Sure! And there is also VP8/9 and AV1. But I ask my self is it really necessary to limit the codec? What if someone only wants to record the stream to disk with a different codec, or stream something exotic over UDP?

The other question to is, is the check of `vprofile.empty()` and `vpreset.empty()` needed? When a codec supports this, and the setting is not present then it uses a default.

But I'm following your lead here.

winlinvip:
The transcoding feature is a subset of FFmpeg, limited to protocols supported by RTMP, so there's no need to support VP8/VP9/AV1 at the moment.

Supporting all codecs and parameters that FFmpeg supports would make it very complex and difficult to use, so it would be better to use FFmpeg directly.

As for profile and preset, I recommend making them mandatory to specify. If not specified, there will still be a default value, which might result in low video quality, and users might not notice this parameter. By making it mandatory, it will be easier for users to know how to adjust it.

jb-alvarado:
> The transcoding feature is a subset of FFmpeg, limited to protocols supported by RTMP, so there's no need to support VP8/VP9/AV1 at the moment.

There is [work](https://github.com/veovera/enhanced-rtmp) on supporting this other codec too, But yeah maybe is not widely supported.

Is fine, then I just also add x265.

winlinvip (approved):

winlinvip:
Yep, I know enhanced RTMP has AV1 and Opus support, but SRS doesn't support them right now. We will add other codecs in future.

Thank you for your work. :)

pannal:
Any plans on when you'll add other codecs to RTMP/MPEG-FLV/TS? It'd be awesome if SRS could support those. Especially with many newer GPUs and iGPUs supporting encoding to AV1 natively.

Thanks!

PR #4142, opened 2024-08-13
Config: Add more utest for env config. v6.0.147 v7.0.4

suzp1984:
1. don't use static variable to store the result;
2. add more UT to handle the multi value and values with whitespaces;

related to #4092 

https://github.com/ossrs/srs/blob/16e569d82357757ddac6ef91d7a5fe7837319909/trunk/src/app/srs_app_config.cpp#L71-L82

`static SrsConfDirective* dir` removed, this static var here is to avoid the memory leak, I add the `SrsConfDirective` instance to the `env_dirs` directive container, which will destroy itself inside `SrsConfig` destructor.

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -69,17 +69,18 @@ const char* _srs_version = "XCORE-" RTMP_SIG_SRS_SERVER;
 #define SRS_OVERWRITE_BY_ENV_FLOAT_SECONDS(key) if (!srs_getenv(key).empty()) return srs_utime_t(::atof(srs_getenv(key).c_str()) * SRS_UTIME_SECONDS)
 #define SRS_OVERWRITE_BY_ENV_FLOAT_MILLISECONDS(key) if (!srs_getenv(key).empty()) return srs_utime_t(::atof(srs_getenv(key).c_str()) * SRS_UTIME_MILLISECONDS)
 #define SRS_OVERWRITE_BY_ENV_DIRECTIVE(key) {                                 \
-        SrsConfDirective* dir = env_dirs->get(key);                           \
+        SrsConfDirective* dir = env_cache_->get(key);                         \
         if (!dir && !srs_getenv(key).empty()) {                               \
             std::vector<string> vec = srs_string_split(srs_getenv(key), " "); \
             dir = new SrsConfDirective();                                     \
             dir->name = key;                                                  \
             for (size_t i = 0; i < vec.size(); ++i) {                         \
-                if (!vec[i].empty()) {                                        \
-                    dir->args.push_back(vec[i]);                              \
+                std::string value = vec[i];                                   \
+                if (!value.empty()) {                                        \
+                    dir->args.push_back(value);                              \
                 }                                                             \
             }                                                                 \
-            env_dirs->directives.push_back(dir);                              \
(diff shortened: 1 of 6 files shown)


trunk/src/app/srs_app_config.cpp
```diff
@@ -69,14 +69,17 @@ const char* _srs_version = "XCORE-" RTMP_SIG_SRS_SERVER;
 #define SRS_OVERWRITE_BY_ENV_FLOAT_SECONDS(key) if (!srs_getenv(key).empty()) return srs_utime_t(::atof(srs_getenv(key).c_str()) * SRS_UTIME_SECONDS)
 #define SRS_OVERWRITE_BY_ENV_FLOAT_MILLISECONDS(key) if (!srs_getenv(key).empty()) return srs_utime_t(::atof(srs_getenv(key).c_str()) * SRS_UTIME_MILLISECONDS)
 #define SRS_OVERWRITE_BY_ENV_DIRECTIVE(key) {                                 \
-        static SrsConfDirective* dir = NULL;                                  \
+        SrsConfDirective* dir = env_dirs->get(key);                           \
```
winlinvip: Why change local variable to member field?

suzp1984: the `static` var is to workaround the `memory leak`.
the problem of `static` is that the env cannot set twice in UT. once the env is parsed, the `static` dir set value, this macro will not parse any new values again, else a new process started, that's a problem for the UT.

If a better UT is our target, the solution is to add another root `directive`, env_dirs, to store the env parsed directives. Then this macro can parse new env value for each `SrsConfig` instance once.

winlinvip: There is a logical error here: using static creates a variable in each function, while you are sharing a variable in the config. This will cause incompatibility with the previous logic.

For example, if you call get_vhost_on_connect first and then call get_vhost_on_close, the latter will retrieve the data from the former.

suzp1984: No, I don't think so. 

`SRS_OVERWRITE_BY_ENV_DIRECTIVE` is a macro, which will expand before compiling, while the `env_dirs` is the `dir`'s container, which is managed by `SrsConfig`.

The `dir` variable is shared among function, that's true, because its container `env_dirs` is the member variable of `SrsConfig`, whose member function can access it. But `get_vhost_on_connect` and `get_vhost_on_close` only create or retrieve the `dir` by its key, in this case, they are `SRS_VHOST_HTTP_HOOKS_ON_CONNECT` and `SRS_VHOST_HTTP_HOOKS_ON_CLOSE`.

If ` env_dirs->get("SRS_VHOST_HTTP_HOOKS_ON_CONNECT"); ` is none, then create a new one and parse the env.

```
    dir = new SrsConfDirective();                                     
    dir->name = key;
    dir->args.push_back(args...);
```
store the `dir` to `env_dirs`: 
`env_dirs->directives.push_back(dir);`

Almost same logic compared with the `static` solution, but the `get_vhost_on_connect` will parse the env just  once per each `SrsConfig` instance. It is convenience for the UT.
While the `static` solution also create the `dir` and parse env once in `SrsConfig` Class scope.

winlinvip: Yep, you are correct, no problem.

winlinvip (approved):

winlinvip (approved):

PR #4149, opened 2024-08-20
ST: Replace macros with explicit code for better understanding. v7.0.7

winlinvip:
Improvements for ST(State Threads):

1. ST: Use g++ for CXX compiler.
2. ST: Remove macros for clist.
3. ST: Remove macros for global thread and vp.
4. ST: Remove macros for vp queue operations.
5. ST: Remove macros for context switch.
6. ST: Remove macros for setjmp/longjmp.
7. ST: Remove macro for stack pad.
8. ST: Refine macro for valgrind.

---------

### DIFF
--- a/trunk/3rdparty/st-srs/Makefile
+++ b/trunk/3rdparty/st-srs/Makefile
@@ -46,6 +46,7 @@ VERSION     = 1.9
 ##########################
 
 CC          = cc
+CXX         = g++
 AR          = ar
 LD          = ld
 RANLIB      = ranlib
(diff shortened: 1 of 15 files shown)


suzp1984:
Is a `inline` function is much better?

And for the Macros keep them, use a compiling option to control the Macro definition.
The benefit is that most of the code will not changed. The Macro can be keep as deprecated or the functions can be an experimental.

```
#ifdef USE_MACRO
#define _ST_SWITCH_CONTEXT(_thread) ....
#else
#define _ST_SWITCH_CONTEXT(_thread)  _st_switch_context(_thread)
#endif 
```

winlinvip:
> Is a `inline` function is much better?

Winlin: Change to inline.

> And for the Macros keep them, use a compiling option to control the Macro definition. The benefit is that most of the code will not changed. The Macro can be keep as deprecated or the functions can be an experimental.

Winlin: I don't believe that duplicated code is a good practice.

suzp1984 (approved):

winlinvip:
See also https://github.com/ossrs/srs/commit/55610cf68904e5b90edb7dd810c5d56c63aff915

PR #4152, opened 2024-08-22
Build: Support static linked executable

### LINKED ISSUE #4148: Regarding the issue of static compilation not taking effect
I specified the -- static option when configuring, but the compiled SRS program is still dynamically linked. There is no such problem in version 3.0, but upgrading to version 4.0 will cause this issue, which has been bothering me for a long time. I hope the experts can give me some advice!

=======================================================================================

我在configure时指定了--static选项，但是编译得到的srs程序仍然是dynamically linked，在3.0版本没有这个问题，但是升级4.0版本就会这样，困扰了很长时间，望大佬们赐教！

suzp1984:
related to #4148 : support static linked srs executable.

Only verified Linux and MacOS platform, Cygwin wait for verify.

## How to use static linking option and verify the static linked executable?

> docker run -it --rm -v \`pwd\`:/srs -w /srs ossrs/srs:ubuntu20 bash -c "./configure --static=on && make"
> docker run -it --rm -v \`pwd\`:/srs -w /srs ossrs/srs:ubuntu20 bash -c "ldd ./objs/srs"

if you just want the static linking the c++ std library, use option `--static-stdcpp=on`: 

> docker run -it --rm -v \`pwd\`:/srs -w /srs ossrs/srs:ubuntu20 bash -c "./configure --static-stdcpp=on && make"

# MacOS platform

Mac OS X does not support statically linked executable binaries. The compiler exit with error for `-static`.
@see https://developer.apple.com/library/archive/qa/qa1118/_index.html

for `-static-libstdc++` compiler option: MacOS didn't implement it, the compiler just give a unused warning.

# Linux platform

`-static`, `-static-libstdc++` and `-static-libgcc` support them well.
But this PR not include the `-static-libgcc`.

# Cygwin

Not yet verified.

### DIFF
--- a/trunk/configure
+++ b/trunk/configure
@@ -224,6 +224,10 @@ fi
 # so we need to link the c++ libraries staticly but not all.
 # @see https://stackoverflow.com/a/26107550
 if [[ $SRS_STATIC == YES ]]; then
+    SrsLinkOptions="${SrsLinkOptions} -static";
+fi
+
+if [[ $SRS_STATIC_STD_CPP == YES ]]; then
     SrsLinkOptions="${SrsLinkOptions} -static-libstdc++";
 fi
 
--- a/trunk/auto/options.sh
+++ b/trunk/auto/options.sh
@@ -46,6 +46,7 @@ SRS_JOBS=1
 # If enabled, force to use SRS_JOBS for make on linux, however you're able to overwrite by -jN on macOS.
 SRS_FORCE_MAKE_JOBS=YES
 SRS_STATIC=NO
+SRS_STATIC_STD_CPP=NO
 # If enabled, link shared libraries for libst.so which uses MPL license.
 # See https://ossrs.net/lts/zh-cn/license#state-threads
 SRS_SHARED_ST=NO
@@ -205,6 +206,7 @@ Performance:                @see https://ossrs.net/lts/zh-cn/docs/v5/doc/perform
 
 Toolchain options:
   --static=on|off           Whether add '-static' to link options. Default: $(value2switch $SRS_STATIC)
+  --static-stdcpp=on|off    Whether add '-static-libstdc++' to link options. Default: $(value2switch $SRS_STATIC_STD_CPP)
   --cc=<CC>                 Toolchain: Use c compiler CC. Default: $SRS_TOOL_CC
   --cxx=<CXX>               Toolchain: Use c++ compiler CXX. Default: $SRS_TOOL_CXX
   --ar=<AR>                 Toolchain: Use archive tool AR. Default: $SRS_TOOL_CXX
@@ -306,6 +308,7 @@ function parse_user_option() {
         --config)                       SRS_DEFAULT_CONFIG=${value} ;;
 
         --static)                       SRS_STATIC=$(switch2value $value) ;;
+        --static-stdcpp)                SRS_STATIC_STD_CPP=$(switch2value $value) ;;
         --cpu)                          SRS_CROSS_BUILD_CPU=${value} ;;
         --arch)                         SRS_CROSS_BUILD_ARCH=${value} ;;
         --host)                         SRS_CROSS_BUILD_HOST=${value} ;;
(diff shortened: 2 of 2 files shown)


trunk/configure
```diff
@@ -224,6 +224,10 @@ fi
 # so we need to link the c++ libraries staticly but not all.
 # @see https://stackoverflow.com/a/26107550
 if [[ $SRS_STATIC == YES ]]; then
```
winlinvip, line 226: Why need `SRS_STATIC_STD_CPP`? Maybe we should only support `SRS_STATIC` which leads to ` -static-libstdc++ -static`?

suzp1984: `-static` will generate a static linked executer, 
`-static-libstdc++` only link the static libstdc++ library.
So `-static` will cover `-static-libstdc++` scope, I think it's no need to combine `-static -static-libstdc++` together.

https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html

winlinvip: What is the output if use `-static` without `-static-libstdc++`? User only want a static binary without depends, majority of users can't understand the nuance of these two options.

suzp1984: ## On Linux:
`-static` output:

> docker run -it --rm -v \`pwd\`:/srs -w /srs ossrs/srs:ubuntu20 bash -c "./configure --static=on && make"


```
/usr/bin/ld: ./objs/src/kernel/srs_kernel_error.o: in function `parse_symbol_offset(char*)':
/srs/./src/kernel/srs_kernel_error.cpp:73: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: ./objs/src/kernel/srs_kernel_utility.o: in function `srs_dns_resolve(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int&)':
/srs/./src/kernel/srs_kernel_utility.cpp:161: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: ./objs/openssl/lib/libcrypto.a(b_sock.o): in function `BIO_gethostbyname':
b_sock.c:(.text+0x78): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
make[1]: Leaving directory '/srs'
```

**The warning said: `dlopen`, `getaddrinfo` and `gethostbyname` requires libc similar runtime shared library to support.** 

> docker run -it --rm -v \`pwd\`:/srs -w /srs ossrs/srs:ubuntu20 bash -c "ldd ./objs/srs"

> not a dynamic executable


## with both `-static -static-ibstdc++`

same warning like `-static`.
`ldd ./objs/srs`
same output: `not a dynamic executable`.

## git show 6314c273ff19e870c7b74f0b57e50aaf904a2649

on above commit, `SRS_STATIC` use `-static-libstdc++` to replace `-static`.
But I think keep fully static linked exec can be an option, but not recommended.

winlinvip: Upon investigation, when we utilize certain functions such as dlopen and gethostbyname, we cannot achieve complete static linking, even though we use the -static flag; full static linking is not possible.

Only by removing the dependencies on these functions can we fully utilize the -static flag to create a version that truly does not depend on glibc.

Therefore, it is necessary to remove the dependencies on these functions. The dependency on dlopen comes from third-party libraries, which might be removed through compilation options. Functions like gethostbyname may require the use of different libraries or an implementation by the ST team itself.

https://stackoverflow.com/questions/15165306/compile-a-static-binary-which-code-there-a-function-gethostbyname

https://github.com/xmrig/xmrig/issues/2902

https://www.reddit.com/r/haskell/comments/vqqq7x/comment/iez099a/

`TRANS_BY_GPT4`

winlinvip: To facilitate full static linking, we can only support the `-static` option, rather than also having a `-static-libstdc++` option, which users might not understand how to use.

Essentially, when we support `-static`, a multitude of warnings are generated. These warnings actually indicate potential issues, hence our support for `-static` is not robust and leaves underlying problems unresolved.

We absolutely cannot offer an option that carries potential risks, as it would make any arising issues even more difficult to address.

`TRANS_BY_GPT4`

PR #4161, opened 2024-08-29
Fix: Remove redundant enabled checks in HLS/DASH cleanup to prevent resource leaks. v6.0.173 v7.0.54

suzp1984:
## Problem
HLS and DASH components had redundant `enabled` flag checks in their `cycle()` and `cleanup_delay()` methods that prevented proper cleanup of files when components were disabled. This created a race condition where:

1. Stream stops publishing and HLS/DASH components get disabled
2. `cycle()` returns early without performing disposal operations  
3. `cleanup_delay()` returns 0 instead of configured disposal timeout
4. Source cleanup doesn't wait long enough for file disposal
5. HLS/DASH files remain on disk without proper cleanup

## Root Cause
The `enabled` flag should control processing of **new incoming streams**, but should NOT prevent **cleanup of existing files** from previously enabled streams.

## Solution
Remove redundant `enabled` checks from:
- `SrsHls::cycle()` and `SrsDash::cycle()` - Allow disposal logic to run even when disabled
- `SrsHls::cleanup_delay()` and `SrsDash::cleanup_delay()` - Always return proper disposal timeout


winlinvip:
In the source cycle, it will call hls dispose:

```cpp
srs_error_t SrsLiveSourceManager::notify(int event, srs_utime_t interval, srs_utime_t tick) {
        // Do cycle source to cleanup components, such as hls dispose.
        if ((err = source->cycle()) != srs_success) {

srs_error_t SrsOriginHub::cycle() {
    if ((err = hls->cycle()) != srs_success) {
        return srs_error_wrap(err, "hls cycle");
    }

srs_error_t SrsHls::cycle() {
    if (srs_get_system_time() - last_update_time <= hls_dispose) {
        return err;
    }

    dispose();
```

The stream dead detection will check the delay cleanup:

```cpp
srs_error_t SrsLiveSourceManager::notify(int event, srs_utime_t interval, srs_utime_t tick)
        if (source->stream_is_dead()) {

bool SrsLiveSource::stream_is_dead() {
    if (now < stream_die_at_ + hub->cleanup_delay()) {
        return false;
    }

srs_utime_t SrsOriginHub::cleanup_delay() {
    srs_utime_t hls_delay = hls->cleanup_delay();

srs_utime_t SrsHls::cleanup_delay()
{
    if (!enabled) {
        return 0;
    }

    // We use larger timeout to cleanup the HLS, after disposed it if required.
    return _srs_config->get_hls_dispose(req->vhost) * 1.1;
}
```

After FFmepg quit, the HLS is not enabled, so it also disable the dispose. I think we should fix this issue by modify `SrsHls::cleanup_delay` rather than call hls dispose again.

In general, we should think about how to dispose a `!enabled` stream source in cycle, not dispose it again before destroy it.

cdgraff:
+1 I reach here for the same issue, some timeline to fix this? or merge this changes?

pisken:
hi, how long to release this fix? I'm having this problem and it's causing disk space overflow.

winlinvip (approved):

PR #4158, opened 2024-08-26
Proxy: Support proxy server for SRS. v7.0.16

winlinvip:
Please note that the proxy server is a new architecture or the next version of the Origin Cluster, which allows the publication of multiple streams. The SRS origin cluster consists of a group of origin servers designed to handle a large number of streams.

```text
                         +-----------------------+
                     +---+ SRS Proxy(Deployment) +------+---------------------+
+-----------------+  |   +-----------+-----------+      +                     +
| LB(K8s Service) +--+               +(Redis/MESH)      + SRS Origin Servers  +
+-----------------+  |   +-----------+-----------+      +    (Deployment)     +
                     +---+ SRS Proxy(Deployment) +------+---------------------+
                         +-----------------------+
```

The new origin cluster is designed as a collection of proxy servers. For more information, see [Discussion #3634](https://github.com/ossrs/srs/discussions/3634). If you prefer to use the old origin cluster, please switch to a version before SRS 6.0.

A proxy server can be used for a set of origin servers, which are isolated and dedicated origin servers. The main improvement in the new architecture is to store the state for origin servers in the proxy server, rather than using MESH to communicate between origin servers. With a proxy server, you can deploy origin servers as stateless servers, such as in a Kubernetes (K8s) deployment.

Now that the proxy server is a stateful server, it uses Redis to store the states. For faster development, we use Go to develop the proxy server, instead of C/C++. Therefore, the proxy server itself is also stateless, with all states stored in the Redis server or cluster. This makes the new origin cluster architecture very powerful and robust.

The proxy server is also an architecture designed to solve multiple process bottlenecks. You can run hundreds of SRS origin servers with one proxy server on the same machine. This solution can utilize multi-core machines, such as servers with 128 CPUs. Thus, we can keep SRS single-threaded and very simple. See https://github.com/ossrs/srs/discussions/3665#discussioncomment-6474441 for details.

```text
                                       +--------------------+
                               +-------+ SRS Origin Server  +
                               +       +--------------------+
                               +
+-----------------------+      +       +--------------------+
+ SRS Proxy(Deployment) +------+-------+ SRS Origin Server  +
+-----------------------+      +       +--------------------+
                               +
                               +       +--------------------+
                               +-------+ SRS Origin Server  +
                                       +--------------------+
```

Keep in mind that the proxy server for the Origin Cluster is designed to handle many streams. To address the issue of many viewers, we will enhance the Edge Cluster to support more protocols.

```text
+------------------+                                               +--------------------+
[...]

PR #4171, opened 2024-09-05
Heartbeat: Report ports for proxy server. v5.0.215 v6.0.156 v7.0.15

winlinvip:
The heartbeat of SRS is a timer that requests an HTTP URL. We can use this heartbeat to report the necessary information for registering the backend server with the proxy server.

```text
SRS(backend) --heartbeat---> Proxy server
```

A proxy server is a specialized load balancer for media servers. It operates at the application level rather than the TCP level. For more information about the proxy server, see issue #4158.

Note that we will merge this PR into SRS 5.0+, allowing the use of SRS 5.0+ as the backend server, not limited to SRS 7.0. However, the proxy server is introduced in SRS 7.0.

It's also possible to implement a registration service, allowing you to use other media servers as backend servers. For example, if you gather information about an nginx-rtmp server and register it with the proxy server, the proxy will forward RTMP streams to nginx-rtmp. The backend server is not limited to SRS.

---------

### DIFF
--- a/trunk/src/app/srs_app_config.hpp
+++ b/trunk/src/app/srs_app_config.hpp
@@ -1119,6 +1119,7 @@ public:
     virtual std::string get_heartbeat_device_id();
     // Whether report with summaries of http api: /api/v1/summaries.
     virtual bool get_heartbeat_summaries();
+    bool get_heartbeat_ports();
 // stats section
 private:
     // Get the stats directive.
--- a/trunk/src/app/srs_app_heartbeat.cpp
+++ b/trunk/src/app/srs_app_heartbeat.cpp
@@ -50,18 +50,23 @@ srs_error_t SrsHttpHeartbeat::do_heartbeat()
         return srs_error_wrap(err, "http uri parse hartbeart url failed. url=%s", url.c_str());
     }
     
-    SrsIPAddress* ip = NULL;
+    string ip;
     std::string device_id = _srs_config->get_heartbeat_device_id();
-    
-    vector<SrsIPAddress*>& ips = srs_get_local_ips();
-    if (!ips.empty()) {
-        ip = ips[_srs_config->get_stats_network() % (int)ips.size()];
+
+    // Try to load the ip from the environment variable.
+    ip = srs_getenv("srs.device.ip"); // SRS_DEVICE_IP
+    if (ip.empty()) {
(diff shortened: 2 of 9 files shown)


trunk/src/app/srs_app_config.hpp
```diff
@@ -1119,6 +1119,7 @@ class SrsConfig
     virtual std::string get_heartbeat_device_id();
     // Whether report with summaries of http api: /api/v1/summaries.
     virtual bool get_heartbeat_summaries();
+    bool get_heartbeat_ports();
```
suzp1984, line 1122: let this method `virtual`?

trunk/src/app/srs_app_heartbeat.cpp
```diff
@@ -48,25 +50,84 @@ srs_error_t SrsHttpHeartbeat::do_heartbeat()
...
+            uint16_t endpoint = _srs_config->get_srt_listen_port();
+            o->append(SrsJsonAny::str(srs_int2str(endpoint).c_str()));
+        }
+
+        // For WebRTC listen endpoints.
+        if (_srs_config->get_rtc_server_enabled()) {
+            SrsJsonArray* o = SrsJsonAny::array();
+            obj->set("rtc", o);
+
+            int endpoint = _srs_config->get_rtc_server_listen();
```
suzp1984, line 127: the `proxy` seems only support RTC over UDP right now, I guess `RTC over TCP` can be supported soon, so why not add tcp endpoint, `SrsConfig::get_rtc_server_tcp_listen`, before the new feature appear?

suzp1984 (approved):

trunk/src/app/srs_app_config.hpp
winlinvip: Unnecessary.

trunk/src/app/srs_app_heartbeat.cpp
winlinvip: Make sense. The UDP and TCP listen port is the same, so the rtc endpoint should be format as `protocol://ip:port` like `udp://:8000` or `tcp://:8000` as such.

jurry:
Is there any plan tu publish this version in hub.docker.com? I'd like to use the new proxy and register servers via this improved heartbeat (and I use helm charts for the deployment)

Also, when will it be available in Oryx?

PR #4199, opened 2024-10-09
fix param's repeat in http callback

### LINKED ISSUE #3930: The "param" parameter is repeated in the http callback
**Describe the bug**
The parameters carried by the "on_publish" and "on_unpublish" events of the HTTP callback will be repeated when the callback is triggered.

**Version**
docker  ossrs/srs:5
SRS/5.0.205

**To Reproduce**
1.docker run srs
image

2. Adjust the SRS configuration settings (using the rtmp2rtc.conf file as a reference).
image

3.obs starts broadcasting
image

4.Request parameter content is duplicated
image

**Expected behavior**
Content appears only once

`TRANS_BY_GPT4`

accestfromim:
This PR try to fix #3930 .

Line 59 in trunk/src/protocol/srs_protocol_utility.cpp,  causes the repeat of the param
because in some case, tcUrl has already contained param, but argument param repeat it.
And after this line, it repeats
```
fullUrl += param.empty() ? "" : (param.at(0) == '?' ? param : "?" + param);
```

line 67 to 69 below are able to get the param, if append at line 59, it will repeat
```
fullUrl = fullUrl.substr(0, pos_query) // rtmp://ip/app/app2
                  + fullUrl.substr(pos_rslash) // /stream
                  + fullUrl.substr(pos_query, pos_rslash - pos_query); // ?k=v
```

successed!

---------

`TRANS_BY_GPT4`

### DIFF
--- a/trunk/src/protocol/srs_protocol_utility.cpp
+++ b/trunk/src/protocol/srs_protocol_utility.cpp
@@ -52,11 +52,6 @@ void srs_discovery_tc_url(string tcUrl, string& schema, string& host, string& vh
     //      rtmp://ip/app?vhost=VHOST/stream
     string fullUrl = srs_string_replace(tcUrl, "...vhost...", "?vhost=");
 
-    // Standard URL is:
-    //      rtmp://ip/app/app2/stream?k=v
-    // Where after last slash is stream.
-    fullUrl += stream.empty() ? "/" : (stream.at(0) == '/' ? stream : "/" + stream);
-
     // First, we covert the FMLE URL to standard URL:
     //      rtmp://ip/app/app2?k=v/stream , or:
     //      rtmp://ip/app/app2#k=v/stream
--- a/trunk/src/protocol/srs_protocol_utility.cpp
+++ b/trunk/src/protocol/srs_protocol_utility.cpp
@@ -56,7 +56,6 @@ void srs_discovery_tc_url(string tcUrl, string& schema, string& host, string& vh
     //      rtmp://ip/app/app2/stream?k=v
     // Where after last slash is stream.
     fullUrl += stream.empty() ? "/" : (stream.at(0) == '/' ? stream : "/" + stream);
-    fullUrl += param.empty() ? "" : (param.at(0) == '?' ? param : "?" + param);
 
     // First, we covert the FMLE URL to standard URL:
     //      rtmp://ip/app/app2?k=v/stream , or:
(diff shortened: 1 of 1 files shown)


accestfromim:
ok, now all checks have passed.

The problem is, when calling this function,
```
void srs_discovery_tc_url(string tcUrl, string &schema, string &host, string &vhost, string &app, string &stream, int &port, string &param)
```
 if tcUrl contains param, the argrument param should be empty, or it will cause repeat in param
 by adding a check in this function, I solved this problem, but I think the proper solution should be calling this function properly. And doing this will lead to a large amount of modifies to current code, some will bring new problem, so I didn't choose to do so.

accestfromim:
And the same question also exist with arguement 'stream',
according to the test of this function, tcUrl shouldn't contain stream name, or streamUrl will repeat too! as shown in original issue #3930 
but this problem can't be solved in the same way as the repeat of param

winlinvip:
The function `srs_discovery_tc_url` is a utility function, please update the utests to cover your modification.

accestfromim:
Thank you for replying!

Next, I would like to confirm with you what you want me to do.

My modification has passed the CI, indicating that it has passed all previous tests and has not introduced any new bugs.

My modification addresses this issue: when tcUrl contains param, passing a non-empty param does not cause param to be duplicated after calling this function.

So, when you ask me to add a test, are you referring to adding a test for the scenario where when tcUrl contains param, passing a non-empty param does not cause param to be duplicated after calling this function?

Additionally, there are tests for this utility function in the three test files trunk\src\utest\srs_utest_rtmp.cpp, trunk\src\utest\srs_utest_protocol2.cpp, and trunk\src\utest\srs_utest_rtmp.cpp. Should I add this test to one of these files, or should I add it to all of them, or should I create a separate test file?

This is my first pull request, and I really hope to get your kindly guidance. Thank you very much.

accestfromim:
> The function `srs_discovery_tc_url` is a utility function, please update the utests to cover your modification.

Thank you for replying!

Next, I would like to confirm with you what you want me to do.

My modification has passed the CI, indicating that it has passed all previous tests and has not introduced any new bugs.

My modification addresses this issue: when tcUrl contains param, passing a non-empty param does not cause param to be duplicated after calling this function.

So, when you ask me to add a test, are you referring to adding a test for the scenario where when tcUrl contains param, passing a non-empty param does not cause param to be duplicated after calling this function?

Additionally, there are tests for this utility function in the three test files trunk\src\utest\srs_utest_protocol.cpp, trunk\src\utest\srs_utest_protocol2.cpp, and trunk\src\utest\srs_utest_rtmp.cpp. Should I add this test to one of these files, or should I add it to all of them, or should I create a separate test file?

This is my first pull request, and I really hope to get your kindly guidance. Thank you very much.

Issue #4198, opened 2024-10-09
Inconsistent info displayed on Stream and Client pages


laurfb:
Hi.
I use SRS to distribute some srt streams.
Everything is functional, but there is an error in the monitoring part.
For example, in the Streams page SRS server display all 5 active publishing sources while in the Clients page SRS report only 3 of them as active (3 from 5).
Furthermore, stream parameters (video, audio) are not displayed in the appropriate fields (Video, Audio) in Streams page. Only Inbound and Outbound data.
Bug or I'm doing something wrong?
regards,
Laur

**Version**
SRS v6.0.155, v5.0.x

**To Reproduce**
1. Publish few srt stream to SRS (ex. h264, 1080p50, 2-8Mbps, AAC audio, 2ch, 256Kbps)
2. Play those streams.
3. Check and compare info from Streams and Clients page.

**Expected behavior**
View all streams active in Clients page.
View streams info (video and audio parameters) in Clients page.

**Screenshots**
Please find them attached.
![streams](https://github.com/user-attachments/assets/27680a2c-5057-4c88-bfc4-89da6b148bff)
![clients](https://github.com/user-attachments/assets/71b55bf5-fb0f-45fe-b98b-105dae4b5813)

TagorePde:
Also, if you have more of than 10 streamings, like 12 for example you only see 10 always.

laurfb:
**Later edit**, regarding (**NO**) Audio/Video info on the console Stream page:

If "**srt_to_rtmp**" is set to "**on**" (in conf file) the info regarding the streams (codec, level, frame size, bitrate) is displayed but with some errors (ex. Audio bitrate is always **44100** regardless of the real value of the srt input stream - it seems that the parsing routine extract, incorrect, the value from the resulting rtmp (transcoded) stream, not the audio bitrate from the input srt stream, as expected.
![sshot-350](https://github.com/user-attachments/assets/65321ecb-4dac-4b40-bce8-cd8b6d12419c)

Also, for specific streams, the corresponding reported frame size for srt video is inaccurate (ex. for 1920x1080 frame size srs routine report "128x2144"). Or video compression level reported as "other" instead of "4.1".
Correct values, from the input srt streams for that example are: h264, profile High, level 4.1, **1920x1080** frame size, AAC (LC), **48000**, stereo.
If "**srt_to_rtmp**" option is off **NO** info will be displayed for any srt stream.
[...]

PR #4211, opened 2024-10-24
rise default displayed streams/clients from 10 to 100

laurfb:
Relative to  #4198 issue (limited clients/streams displayed on srs console).
More than enaugh for most users.
Maybe it would be more useful to define a variable in the conf file...


winlinvip:

PR #4216, opened 2024-10-29
HLS:  Fix error when stream has extension. #4215 v5.0.217 v6.0.158 v7.0.19

### LINKED ISSUE #4215: flv-play plays the stream address that has not started the push stream, causing the name after the push stream to be abnormally concatenated with.flv suffix, causing the hls address to be abnormally
**Describe the bug**
flv-play plays the stream address that has not started the push stream, causing the name after the push stream to be abnormally concatenated with.flv suffix, causing the hls address to be abnormally
If the client pulls the stream from SRS before the stream is initiated, and streaming starts subsequently, it will result in the stream name having a .flv suffix, which causes anomalies in the m3u8 URL.

Step1:
flv-play plays the stream address that has not started the push stream
image

Step2:
Push the flow to the same path
image

Step3:
stream name with.flv suffix
image

Step4:
HLS path error
image

**Version**
Docker image: ossrs/srs:v6
Version number: SRS/6.0.155

`TRANS_BY_GPT4`

duiniuluantanqin:
for [issues/4215](https://github.com/ossrs/srs/issues/4215)

---------

### DIFF
--- a/trunk/src/app/srs_app_http_stream.cpp
+++ b/trunk/src/app/srs_app_http_stream.cpp
@@ -630,6 +630,8 @@ srs_error_t SrsLiveStream::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessage
     // Correct the app and stream by path, which is created from template.
     // @remark Be careful that the stream has extension now, might cause identify fail.
     req->stream = srs_path_basename(r->path());
+    // remove the extension of stream if have. for instance, test.flv -> test
+    req->stream = srs_path_filename(req->stream);
 
     // update client ip
     req->ip = hc->remote_ip();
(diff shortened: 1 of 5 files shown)


suzp1984 (approved):

trunk/src/app/srs_app_http_stream.cpp
```diff
@@ -630,6 +630,8 @@ srs_error_t SrsLiveStream::serve_http(ISrsHttpResponseWriter* w, ISrsHttpMessage
     // Correct the app and stream by path, which is created from template.
     // @remark Be careful that the stream has extension now, might cause identify fail.
     req->stream = srs_path_basename(r->path());
+    // remove the extension of stream if have. for instance, test.flv -> test
+    req->stream = srs_path_filename(req->stream);
```
winlinvip, line 634: If the extension removed, the stat API can't know which type of stream, RTMP or FLV or TS.

winlinvip (approved):

PR #4232, opened 2024-11-13
improve timestamp type: int -> uint.

suzp1984:
when I check the timestamp type from RTC -> RTMP.

RTC(uint32_t) -> RTMP (uint64_t).

Some timestamp type can be improved.

e.g. `int64_t` convert to `int32_t`.


suzp1984:
close this PR, because introduced timestamp bug.

Issue #4139, opened 2024-08-09
WebRTC: h264 FU-A parse for both S and E flag


chai51:
**Describe the bug**  
In RTC3984, the **FU-A** format only explains the timing when the `S` flag and the `E` flag appear, so when there are both the S flag and the E flag in a **FU-A** packet, this is also allowed.

**Version**  
Any version

**To Reproduce**  
- Send a **FU-A** packet to SRS via WebRTC, with both the `S` and `E` flags set to 1.  
- In the process of webrtc2rtmp: when `SrsRtcFrameBuilder::packet_video_rtmp` parses the RTP packet, the length of the NALU is not written; the NALUnitLength read by `SrsFormat::do_avc_demux_ibmf_format` is an invalid value.

**Screenshots**
![image](https://github.com/user-attachments/assets/c35c836e-56bc-4553-9b79-067d6a245843)

winlinvip:
Nice work! Would you please file an PullRequest to fix this issue?

PR #4243, opened 2024-11-25
WebRTC: fix h264 FU-A only one package

chai51:
Regarding issue #4139, the web can parse it normally. It also conforms to the RFC documentation standards.

---------

`TRANS_BY_GPT4`

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -1834,7 +1834,12 @@ srs_error_t SrsRtcFrameBuilder::packet_video_rtmp(const uint16_t start, const ui
             if (fua_payload->start) {
                 nalu_len = fua_payload->size + 1;
                 //skip 4 bytes to write nalu_len future
-                payload.skip(4);
+                if (fua_payload->end) {
+                    //only one FU-A
+                    payload.write_4bytes(nalu_len);
+                } else {
+                    payload.skip(4);
+                }
                 payload.write_1bytes(fua_payload->nri | fua_payload->nalu_type);
                 payload.write_bytes(fua_payload->payload, fua_payload->size);
             } else {


suzp1984:
> A fragmented NAL unit MUST NOT be transmitted in one FU; that is, the
> Start bit and End bit MUST NOT both be set to one in the same FU header.

This comments is come from the [rfc6184#section-5.8](https://datatracker.ietf.org/doc/html/rfc6184#section-5.8), appeared in the middle of page 32.
Same comments also appears in the [rfc3984#section-5.8](https://datatracker.ietf.org/doc/html/rfc3984#section-5.8), in the middle of page 30.

So, both [rfc6184#section-5.8](https://datatracker.ietf.org/doc/html/rfc6184#section-5.8) and [rfc3984#section-5.8](https://datatracker.ietf.org/doc/html/rfc3984#section-5.8) said the `S` and `E` bit indicators can not both to true in FU header, then I would like to question #4139 , is that true? This PR is based on and to resolve the #4139. 

The top comments are reasonable, there are multi NAL unit types, check [https://datatracker.ietf.org/doc/html/rfc6184#section-5.2](https://datatracker.ietf.org/doc/html/rfc6184#section-5.2), if `S` and `E` bits are both true, which means this NAL unit need not to be fragmented, the single NAL unit type or aggregation packet types are more suitable here.

Then Let's return to #4139 , Let assume it really happens in some RTC implementation, I would suggest provide more details Step to Step reproduction guides to reproduce this case.

chai51:
Thank you very much for your patient explanation. At that time, I didn't find any document indicating that the S and E flags cannot be 1 simultaneously. When I was implementing it, I found that h5 could unpack it normally, while srs would report an error and discard it. That's why I finally discovered this phenomenon. If the documents all describe it in this way, then indeed my implementation does not conform to the specification. Thank you again.

PR #4252, opened 2024-12-08
WebRTC: Fix missing msid in play answer

parallelcc:
`a=ssrc:<ssrc> msid:<stream-id> <track-id>` is missing in play answer due to not setting the msid for track, so audio and video tracks are not explicitly associated to the same stream, which may cause player avsync and compatibility issues


duiniuluantanqin:
> player avsync and compatibility issues

Could you please provide a detailed explanation regarding these two points?

`TRANS_BY_GPT4`

winlinvip:

Issue #3701, opened 2023-07-22
HTTP streaming support listening at both IPv4 and IPv6


ChowDPa02k:
> Note: Please read FAQ before file an issue, see #2716

## Description

Please description your issue here

1. SRS Version: 6.0.59(Bee) Windows Build

2. SRS Log:

```
[2023-07-22 20:11:38.163][WARN][2025][al8g3399][0] stats network use index=0, ip=192.168.1.10, ifname={404977E6-A52A-4A91-8D9A-956076CEE4F4}
[2023-07-22 20:11:38.163][WARN][2025][al8g3399][0] stats disk not configed, disk iops disabled.
[2023-07-22 20:11:38.163][INFO][2025][al8g3399] write log to console
[2023-07-22 20:11:38.163][INFO][2025][al8g3399] features, rch:on, dash:on, hls:on, hds:off, srt:off, hc:on, ha:on, hs:on, hp:on, dvr:on, trans:on, inge:on, stat:on, sc:on
[2023-07-22 20:11:38.164][INFO][2025][al8g3399] SRS on amd64 x86_64, conf:conf\live.conf, limit:1000, writev:1024, encoding:little-endian, HZ:1000
[2023-07-22 20:11:38.164][INFO][2025][al8g3399] mw sleep:350ms. mr enabled:on, default:0, sleep:350ms
[2023-07-22 20:11:38.164][INFO][2025][al8g3399] gc:on, pq:30000ms, cscc:[0,16), csa:on, tn:on(may hurts performance), ss:auto(guess by merged write)
[2023-07-22 20:11:38.164][INFO][2025][al8g3399] system default latency(ms): mw(0-350) + mr(0-350) + play-queue(0-30000)
[2023-07-22 20:11:38.164][WARN][2025][al8g3399][0] SRS/6.0.59 is not stable
[2023-07-22 20:11:38.164][INFO][2025][al8g3399] Run in single thread mode
[2023-07-22 20:11:38.168][INFO][2025][al8g3399] fingerprint=47:9E:E9:E2:6E:70:99:BF:27:F9:65:9F:4F:99:C4:14:69:34:DE:67:58:B7:04:BD:5D:71:32:01:E2:1A:5E:81
[2023-07-22 20:11:38.168][INFO][2025][al8g3399] CircuitBreaker: enabled=1, high=2x90, critical=1x95, dying=5x99
[...]

PR #4265, opened 2024-12-23
fix Issue 3701  : HTTP streaming support listening at both IPv4 and IPv6

sangbumlikeagod:
this pr is for #3701 

**Main Changes**

In srs_app_server.cpp, I have updated the following components:
api_listener_
apis_listener_
http_listener_
https_listener_
These components have been replaced with SrsMultipleTcpListeners to enable support for multiple port inputs. The new SrsMultipleTcpListeners operate similarly to the original SrsTcpListener but provide enhanced functionality for handling multiple ports.

**Considerations**

There is a functional relationship between api_listener_ and http_listener_ (as well as between apis_listener_ and https_listener_). Specifically, when the port of api_listener_ matches that of http_listener_, they share the same HTTP multiplexer (http_mux) for efficiency.
With the introduction of multiple port handling via SrsMultipleTcpListeners, careful consideration is required to ensure proper reuse of HTTP components without conflicts.
If the configuration option reuse_api_over_server_ is enabled, the following behavior occurs:
The APIs / and /api/v1/versions are no longer handled by SrsServer::http_handle. Instead, these requests are processed by the HTTP server components.
However, under the new design with SrsMultipleTcpListeners, if there is any overlap between the ports of api(s)_listener_ and http(s)_listener_, the following issues may arise:
Regardless of the value of reuse_api_over_server_:
An error will occur if the two listeners share any ports.
If reuse_api_over_server_ == true:
Some ports in the range of api(s)_listener_ will be unable to handle requests for / and /api/v1/versions, as these are already managed by the HTTP server components.
If reuse_api_over_server_ == false:
The overlapping ports between the two listeners will result in duplicate definitions for / and /api/v1/versions, causing conflicts.
To address these potential issues, I have implemented logic to enforce one of the following conditions:
There must be no overlap between the ports used by api(s)_listener_ and those used by http(s)_listener_.
Alternatively, the range of ports in http(s)_listener_ must fully encompass those in api(s)_listener_.
If neither condition is satisfied, an error will be raised to prevent misconfiguration and ensure consistent behavior.

### DIFF
--- a/trunk/src/app/srs_app_config.cpp
+++ b/trunk/src/app/srs_app_config.cpp
@@ -7741,6 +7741,10 @@ std::vector<std::string> SrsConfig::get_http_apis_listens()
         return ports;
     }
     conf = conf->get("listen");
+    if (!conf) {
+        ports.push_back(DEFAULT);
+        return ports;
+    }
     for (int i = 0; i < (int)conf->args.size(); i++) {
         ports.push_back(conf->args.at(i));
     }
@@ -7969,6 +7973,10 @@ std::vector<std::string> SrsConfig::get_https_apis_listens()
         return ports;
     }
     conf = conf->get("listen");
+    if (!conf) {
+        ports.push_back(DEFAULT);
+        return ports;
+    }
     for (int i = 0; i < (int)conf->args.size(); i++) {
         ports.push_back(conf->args.at(i));
     }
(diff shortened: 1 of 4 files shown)


winlinvip:
Thanks for your PR. Need to postpone this feature.

PR #4295, opened 2025-02-22
RTC: audio packet jitter buffer. v7.0.48

chundonglinlin:
Rtp packets may be retransmitted, disordered, jittery, delayed, etc.There may be abnormalities when converting to rtmp.

To reproduce this problem, you need to set the network reordering by [tc-ui](https://github.com/ossrs/tc-ui). Note that you need a linux server, and start it by docker:

```bash
docker run --network=host --privileged -it --restart always -d \
    --name tc -v /lib/modules:/lib/modules:ro ossrs/tc-ui:1
```

Set up 5% packet reordering and a 1ms delay; then you will notice that the audio is stuttering, somewhat noisy, and lacks fluency.

```bash
curl http://localhost:2023/tc/api/v1/config/raw -X POST \
  -d 'tcset ens5 --direction incoming --delay 40ms --reordering 5% --port 8000'
```

> Note: Even without network conditions, the natural state can also cause packet reordering, especially in public cloud platforms such as AWS EC2.

> Note: You can use command `curl http://localhost:2023/tc/api/v1/config/raw -X POST -d 'tcdel --all ens5'` to reset the network condition settings.

Check the web console, you will see the reordering setup:

TC Settings

Then, publish stream via WHIP: http://localhost:8080/players/whip.html

And, play via HTTP-FLV: http://localhost:8080/players/srs_player.html

Finished by AI:

* [AI: Extract audio jitter buffer to class AudioPacketCache](https://github.com/ossrs/srs/pull/4295/commits/a4097d9374f40a63488c9af3117a22e888961842)
* [AI: Add utest and fix bug.](https://github.com/ossrs/srs/pull/4295/commits/c919227af5cda2666cd2ac5772a6ae8a459745f8)

---------


trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1476,14 +1476,93 @@ srs_error_t SrsRtcFrameBuilder::on_rtp(SrsRtpPacket *pkt)
...
+        if (!force_process) {
+            // If packet is before window start (shouldn't happen normally)
+            if (srs_rtp_seq_distance(last_audio_seq_num_, next_seq) < 0) {
+                // Process it anyway as it's already late
+                srs_warn("Late audio packet, seq=%u, expected>=%u", next_seq, last_audio_seq_num_);
+            } else if (srs_rtp_seq_distance(next_seq, window_end) < 0) {
+                // If packet is beyond window end, stop processing
+                srs_warn("Audio packet beyond window end, seq=%u, window_end=%u", next_seq, window_end);
+                break;
+            } else if (srs_rtp_seq_distance(last_audio_seq_num_, next_seq) > 1) {// If there's a gap and we haven't exceeded wait time, wait for missing packets
```
duiniuluantanqin: `// If there's a gap and we haven't exceeded wait time, wait for missing packets`
This comment has been duplicated.

`TRANS_BY_GPT4`

duiniuluantanqin (approved):

trunk/src/app/srs_app_rtc_source.hpp
```diff
@@ -335,6 +342,13 @@ class SrsRtcFrameBuilder
     uint16_t header_sn_;
     uint16_t lost_sn_;
     int64_t rtp_key_frame_ts_;
+
+    // Audio jitter buffer, map sequence number to packet
+    std::map<uint16_t, SrsRtpPacket*> audio_buffer_;
```
winlinvip, line 393: Obviously, the audio jitter buffer should be a class with its own fields and behaviors, rather than being directly exposed and mixed with the SrsRtcFrameBuilder code.

```
    // Audio jitter buffer, map sequence number to packet
    std::map<uint16_t, SrsRtpPacket*> audio_buffer_;
    // Last processed sequence number
    uint16_t last_audio_seq_num_;
    // Last time we processed the jitter buffer
    int64_t last_audio_process_time_ms_;
```

winlinvip, line 393: It seems that the audio buffer is similar to SrsRtpRingBuffer; please consider whether you can use it directly.

winlinvip (approved):

PR #4300, opened 2025-03-13
fixed:Log level does not work on V2 logs

xengine-qyt:


trunk/src/app/srs_app_utility.cpp
```diff
@@ -55,15 +55,17 @@ SrsLogLevel srs_get_log_level(string level)
...
         return SrsLogLevelTrace;
-    } else if ("warn" == level) {
+    } else if ("DEBUG" == level) {
+        return SrsLogLevelDebug;
+    } else if ("INFO" == level) {
+        return SrsLogLevelInfo;
+    } else if ("WARN" == level) {
         return SrsLogLevelWarn;
-    } else if ("error" == level) {
+    } else if ("ERROR" == level) {
```
suzp1984: I think this PR don't make sense.
The SRS_LOG_LEVEL_V2 is enabled by default.
https://github.com/ossrs/srs/blob/93cba246bc60e88a449ca62fa9652622e67c279a/trunk/auto/options.sh#L71

The `SRS_LOG_LEVEL_V2`, which means the log level names defined in https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/Level.java
or 
https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels/2031209#2031209

This func is just a bridge between log level enum, defined by SRS, and `srs.conf`. So it's not necessary to redefine another `SrsLogLevelXXX` for log level 2. 

The only fault of this func is log level `fatal` is missing, which can be mapping to `SrsLogLevelError` anyway.

If the `level` string are upper cases, the `srs.conf` and Unit test cases need to change to match this change, so I don't think its a good idea also.

trunk/src/kernel/srs_kernel_log.cpp
```diff
@@ -14,7 +14,7 @@
 const char* srs_log_level_strings[] = {
 #ifdef SRS_LOG_LEVEL_V2
         // The v2 log level specs by log4j.
-        "FORB",     "TRACE",     "DEBUG",    NULL,   "INFO",    NULL, NULL, NULL,
+        "FORB",     "VERB",    "TRACE",     "DEBUG",   "INFO",    NULL, NULL, NULL,
```
suzp1984: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels/2031209#2031209
https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/Level.java

`VERB` is not the term defined by log level v2.
And the log level sequence is incorrect.
srs_log_level_strings[1] = TRACE;
srs_log_level_strings[2] = DEBUG;
srs_log_level_strings[4] = INFO;
srs_log_level_strings[8] = WARN;
srs_log_level_strings[16] = ERROR;
srs_log_level_strings[32] = FATAL; 
`FATAL` is not there, instead a [32] = OFF, maybe it is the only arguable point.

trunk/src/kernel/srs_kernel_log.hpp
```diff
@@ -20,20 +20,36 @@
 
 // The log level, see https://github.com/apache/logging-log4j2/blob/release-2.x/log4j-api/src/main/java/org/apache/logging/log4j/Level.java
 // Please note that the enum name might not be the string, to keep compatible with previous definition.
+#ifdef SRS_LOG_LEVEL_V2
```
suzp1984: `SrsLogLevel` is SRS's internal log level enum.
`SRS_LOG_LEVEL_V2` controls the log print with the new standard, ref https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels/2031209#2031209
then the log analysis tools can compatible with those new log level keywords. 
So, redefine `SrsLogLevel` don't make sense here.

suzp1984 (commented):
I don't think this PR make sense.

trunk/src/app/srs_app_utility.cpp
xengine-qyt: .o,it different from the documentation
image

suzp1984: upper cases terms, TRACE, DEBUG, INFO, WARN, ERROR, are print to logs, not the configurable claims, use lower-cases in conf file.

trunk/src/kernel/srs_kernel_log.hpp
xengine-qyt: ok

trunk/src/kernel/srs_kernel_log.cpp
xengine-qyt: ok

duiniuluantanqin:
Do not replace spaces with tabs; it is uncomfortable to look at.

`TRANS_BY_GPT4`

winlinvip:
It seems that this pull request is not necessary.

PR #4301, opened 2025-03-14
Valgrind: Return error for unsupported check=new on Valgrind < 3.21. v7.0.52

duiniuluantanqin:
## Problem
The `valgrind?check=new` API parameter uses `VALGRIND_DO_NEW_LEAK_CHECK` which is only available in Valgrind 3.21+. On older versions like CentOS's default Valgrind 3.16, this causes undefined behavior since the macro is not defined.

## Solution
- Check for `VALGRIND_DO_NEW_LEAK_CHECK` availability before processing the request
- Return `ERROR_NOT_SUPPORTED` with version information when unsupported
- Move the version check before thread creation to avoid unnecessary resource allocation

## Changes
- Early validation of `check=new` parameter compatibility
- Proper error response with current Valgrind version details
- Prevents undefined behavior on older Valgrind installations

Fixes compatibility issues with older Valgrind versions commonly found in enterprise Linux distributions.

---------


trunk/src/app/srs_app_http_api.cpp
```diff
@@ -1185,7 +1185,14 @@ srs_error_t SrsGoApiValgrind::cycle()
         } else if (check == "changed") {
             VALGRIND_DO_CHANGED_LEAK_CHECK;
         } else if (check == "new") {
+#if defined(VALGRIND_DO_NEW_LEAK_CHECK)
             VALGRIND_DO_NEW_LEAK_CHECK;
+#else
+            srs_warn("valgrind?check=new is not supported in current Valgrind version,"
+                    "please upgrade to 3.22 or higher. "
+                    "Fallback to valgrind?check=full");
```
suzp1984: Seems good, but supplement current valgrind version information in this warning msg, maybe more perfect.
`__VALGRIND_MAJOR__` & `__VALGRIND_MINOR__`.

suzp1984 (approved):

trunk/src/app/srs_app_http_api.cpp
duiniuluantanqin: fixed

suzp1984 (approved):

winlinvip (approved):

PR #4309, opened 2025-03-21
Support custom deleter for SrsUniquePtr.

winlinvip:
SrsUniquePtr does not support array or object created by malloc, because we only use delete to dispose the resource. You can use a custom function to free the memory allocated by malloc or other allocators.
```cpp
      char* p = (char*)malloc(1024);
      SrsUniquePtr<char> ptr(p, your_free_chars);
```

This is used to replace the SrsAutoFreeH. For example:
```cpp
      addrinfo* r = NULL;
      SrsAutoFreeH(addrinfo, r, freeaddrinfo);
      getaddrinfo("127.0.0.1", NULL, &hints, &r);
```

Now, this can be replaced by:
```cpp
      addrinfo* r = NULL;
      getaddrinfo("127.0.0.1", NULL, &hints, &r);
      SrsUniquePtr<addrinfo> r2(r, freeaddrinfo);
```

Please aware that there is a slight difference between SrsAutoFreeH and SrsUniquePtr.  SrsAutoFreeH will track the address of pointer, while SrsUniquePtr will not.
```cpp
      addrinfo* r = NULL;
      SrsAutoFreeH(addrinfo, r, freeaddrinfo); // r will be freed even r is changed later.
      SrsUniquePtr<addrinfo> ptr(r, freeaddrinfo); // crash because r is an invalid pointer.
```


trunk/src/utest/srs_utest_core.cpp
```diff
@@ -489,3 +489,120 @@ VOID TEST(CoreSmartPtr, UniquePtrArray)
...
+        SrsUniquePtr<addrinfo> ptr(r, freeaddrinfo);
+    }
+}
+
+class MockSlice
+{
+public:
+    const char* bytes_;
+public:
+MockSlice(const char* bytes) {
```
suzp1984: code indent here.

trunk/src/utest/srs_utest_core.cpp
```diff
@@ -489,3 +489,120 @@ VOID TEST(CoreSmartPtr, UniquePtrArray)
...
+    }
+    virtual ~MockSpecialPacket() {
+        srs_freepa(bytes_);
+    }
+public:
+    static void deleter(vector<MockSpecialPacket*>* pkts) {
+        vector<MockSpecialPacket*>::iterator it;
+        for (it = pkts->begin(); it != pkts->end(); ++it) {
+            MockSpecialPacket* pkt = *it;
+            srs_freepa(pkt->bytes_);
```
suzp1984:
```
srs_freep(pkt);
```
I think `pkt` need to free here, instead of `pkt->bytes_`, which can be freed by `pkt`'s destructor, but `pkt` is not freed, so this is a memory leak.

trunk/auto/options.sh
```diff
@@ -360,6 +360,7 @@ function parse_user_option() {
         --gmp)                          SRS_GPERF_MP=$(switch2value $value) ;;
         --gcp)                          SRS_GPERF_CP=$(switch2value $value) ;;
         --gprof)                        SRS_GPROF=$(switch2value $value) ;;
+        --asan)                         SRS_SANITIZER=$(switch2value $value) ;;
```
suzp1984, line 363: duplicate with `--sanitizer`, or just an alias.

chundonglinlin: I believe it is an alias.

`TRANS_BY_GPT4`

trunk/src/utest/srs_utest_core.cpp
duiniuluantanqin: fixed

trunk/src/utest/srs_utest_core.cpp
duiniuluantanqin: fixed

duiniuluantanqin (approved):

suzp1984 (approved):

trunk/auto/options.sh
winlinvip: Yep. An alias better than sanitizer. 😆

chundonglinlin (approved):

trunk/src/core/srs_core_deprecated.hpp
```diff
@@ -11,44 +11,52 @@
...
+//      SrsUniquePtr<addrinfo> ptr(r, freeaddrinfo);
+//
+// Note: Please aware that there is a slight difference between SrsAutoFreeH and SrsUniquePtr. 
+// SrsAutoFreeH will track the address of pointer, while SrsUniquePtr will not.
+//      addrinfo* r = NULL;
+//      SrsAutoFreeH(addrinfo, r, freeaddrinfo); // r will be freed even r is changed later.
+//      SrsUniquePtr<addrinfo> ptr(r, freeaddrinfo); // crash because r is an invalid pointer.
+//
+// See https://github.com/ossrs/srs/discussions/3667#discussioncomment-8969107 for more details.
 #define SrsAutoFreeH(className, instance, hook) \
```
xiaozhihong: no used any more, delete it or not?

winlinvip: I agree, so let me remove all the unused macros.

duiniuluantanqin (approved):

xiaozhihong (approved):

PR #4322, opened 2025-04-10
feat(http): improved chunk reading

johzzy:
If read as 0, return ERROR_HTTP_RESPONSE_EOF as soon as possible


winlinvip:
Thanks for your PR. Just need to postpone this feature.

PR #4289, opened 2025-02-14
rtmp2rtc: Support RTMP-to-WebRTC conversion with HEVC. v7.0.33

duiniuluantanqin:
## Usage

Launch Chrome with H.265 enabled: 

Upgrade to the latest Chrome browser (version 136) fully enables this by default, so there's no need to launch it with any extra parameters.

Launch SRS with `rtmp2rtc.conf`

```bash
./objs/srs -c conf/rtmp2rtc.conf
```

Push RTMP with H.265 codec, without B frame:

```bash
ffmpeg -stream_loop -1 -re -i doc/source.flv \
  -c:v libx265 -profile:v main -preset fast -b:v 2000k -maxrate 2000k -bufsize 2000k -bf 0 \
  -c:a aac -b:a 48k -ar 44100 -ac 2 -f flv rtmp://localhost/live/livestream
```

Play with WebRTC: http://localhost:8080/players/whep.html?schema=http&&codec=hevc

image

---------

Co-authored-by: chundonglinlin <chundonglinlin@163.com>
Co-authored-by: winlin <winlinvip@gmail.com>
Co-authored-by: john <hondaxiao@tencent.com>

PR #4349, opened 2025-05-14
rtc2rtmp: Support RTC-to-RTMP remuxing with HEVC. v7.0.43

duiniuluantanqin:
**Introduce**

This pull request builds upon the foundation laid in  https://github.com/ossrs/srs/pull/4289 . While the previous work solely implemented unidirectional HEVC support from RTMP to RTC, this submission further enhances it by introducing support for the RTC to RTMP direction.

**Usage**

Launch SRS with `rtc2rtmp.conf`

```bash
./objs/srs -c conf/rtc2rtmp.conf
```

**Push with WebRTC**

Upgrade browser to Chrome(136+) or Safari(18+), then open [WHIP encoder](http://localhost:8080/players/whip.html?schema=http&&codec=hevc), push stream with URL that enables HEVC by query string `codec=hevc`:

```bash
http://localhost:1985/rtc/v1/whip/?app=live&stream=livestream&codec=hevc
```

This query string `codec=hevc` is used to select the video codec, and generate lines in the answer SDP.

```
m=video 9 UDP/TLS/RTP/SAVPF 49 123
a=rtpmap:49 H265/90000
```

The encoder log also show the codec:

```
Audio: opus, 48000HZ, channels: 2, pt: 111
Video: H265, 90000HZ, pt: 49
```

**Play with RTMP**

Play HEVC stream via RTMP.

```bash
ffplay -i rtmp://localhost/live/livestream
```

You will see the codec in logs:

```
  Stream #0:0: Audio: aac (LC), 48000 Hz, stereo, fltp
  Stream #0:1: Video: hevc (Main), yuv420p(tv, bt709), 320x240, 30 fps, 30 tbr, 1k tbn
```

You can also use [WHEP player](http://localhost:8080/players/whep.html?schema=http&&codec=hevc) to play the stream.

Important refactor with AI:

* [AI: Refactor packet cache for RTC frame builder.](https://github.com/ossrs/srs/pull/4349/commits/b8ffa1630e31d5f4505e349167d94eac2ce0cc45)
* [AI: Refactor the packet copy and free for SrsRtcFrameBuilder](https://github.com/ossrs/srs/pull/4349/commits/f3487b45d7804074bff92d54bcda91570682205c)
* [AI: Refactor the frame detector for SrsRtcFrameBuilder](https://github.com/ossrs/srs/pull/4349/commits/4ffc1526b9083143744a4e0aca735e02ef295b44)
* [AI: Refactor the packet_video_rtmp for SrsRtcFrameBuilder](https://github.com/ossrs/srs/pull/4349/commits/81f6aef4edcd8a052083e12929a419bf75d036bb)
* [AI: Add utests for SrsCodecPayload.codec](https://github.com/ossrs/srs/pull/4349/commits/61eb1c0bfc28ec53e4a9f67739c9b0038cebea7a)
* [AI: Add utests for VideoPacketCache in SrsRtcFrameBuilder.](https://github.com/ossrs/srs/pull/4349/commits/fd25480dfa2ee6a1edcbe9ba6d69fe796bd7ecfe)
* [AI: Add utests for VideoFrameDetector in SrsRtcFrameBuilder.](https://github.com/ossrs/srs/pull/4349/commits/b4aa977bbd4118af0ffe5b99ce9fc8ed7875866e)
* [AI: Add regression test for RTC2RTMP with HEVC.](https://github.com/ossrs/srs/pull/4349/commits/5259a2aac3bcf52fa99b89ed1d764fb747c82776)

---------


winlinvip:
When converting WebRTC to RTMP with HEVC, we should choose between eRTMP and RTMP(12), both of which support the HEVC codec. eRTMP is a new standard; you can find more information [here](https://github.com/veovera/enhanced-rtmp). RTMP(12) is a de facto standard in China, which was used before the era of eRTMP.

> Note: eRTMP uses a four-byte codecc to identify HEVC, while RTMP(12) uses codecID 12. Since codecID is 4 bits, the maximum number of codecs that can be supported is 16.

SRS used to use RTMP(12) because SRS needed to support HEVC before eRTMP was released. Therefore, SRT to RTMP and GB to RTMP both use RTMP(12) to support the HEVC codec. FFmpeg also supports RTMP(12) with HEVC. Interestingly, FFmpeg first supported eRTMP for HEVC, then sneakily added support for RTMP(12), which it had refused to support a decade ago.

Based on this context, eRTMP is actually a real standard, not RTMP(12). SRS should support eRTMP by default, rather than RTMP(12). Therefore, when converting WebRTC/SRT/GB to RTMP, eRTMP should be used. This is why this PR uses eRTMP instead of RTMP(12).

We should submit a PR to change the SRT/GB code to replace RTMP(12) with eRTMP. This will make the code consistent. So there may be one or two PRs to finish this.

trunk/src/kernel/srs_kernel_rtc_rtp.cpp
```diff
@@ -939,41 +939,44 @@ srs_error_t SrsRtpPacket::decode(SrsBuffer* buf)
...
+            if(NULL != stap_payload->get_sps() || NULL != stap_payload->get_pps()) {
+                return true;
+            }
+        } else if (nalu_type == kFuA) {
+            SrsRtpFUAPayload2* fua_payload = dynamic_cast<SrsRtpFUAPayload2*>(payload_);
+            if(SrsAvcNaluTypeIDR == fua_payload->nalu_type) {
+                return true;
+            }
+        } else {
+            if((SrsAvcNaluTypeIDR == nalu_type) || (SrsAvcNaluTypeSPS == nalu_type) || (SrsAvcNaluTypePPS == nalu_type)) {
```
suzp1984: not the scope of this commit, but `kStapA` and `SrsAvcNaluTypeIDR` should be in same enum `SrsAvcNaluType`.
https://datatracker.ietf.org/doc/html/rfc6184#section-5.2

trunk/src/kernel/srs_kernel_rtc_rtp.cpp
```diff
@@ -1543,6 +1546,7 @@ ISrsRtpPayloader* SrsRtpFUAPayload2::copy()
     return cp;
 }
 
+#ifdef SRS_H265
```
suzp1984, line 1571: still not sure, SRS_H265 already config as default feature, and when this macro switch can be remove totally?

trunk/src/protocol/srs_protocol_raw_avc.hpp
```diff
@@ -95,6 +95,13 @@ class SrsRawHEVCStream
     // @param flv output the muxed flv packet.
     // @param nb_flv output the muxed flv size.
     virtual srs_error_t mux_avc2flv(std::string video, int8_t frame_type, int8_t avc_packet_type, uint32_t dts, uint32_t pts, char **flv, int *nb_flv);
+    // Mux the hevc video packet to flv video packet, enhanced mode.
+    // @param packet_type, SrsVideoHEVCFrameTraitPacketTypeSequenceStart or SrsVideoHEVCFrameTraitPacketTypeCodedFrames.
+    // @param frame_type, SrsVideoAvcFrameTypeKeyFrame or SrsVideoAvcFrameTypeInterFrame.
+    // @param video the hevc raw data.
+    // @param flv output the muxed flv packet.
+    // @param nb_flv output the muxed flv size.
+    virtual srs_error_t mux_avc2flv_enhanced(std::string video, int8_t frame_type, int8_t packet_type, uint32_t dts, uint32_t pts, char **flv, int *nb_flv);
```
suzp1984, line 106: is avc an alias for H.264? then is this method name can be renamed to:
 ```suggestion
    virtual srs_error_t mux_hevc2flv_enhanced(std::string video, int8_t frame_type, int8_t packet_type, uint32_t dts, uint32_t pts, char **flv, int *nb_flv);
```

File: trunk/src/kernel/srs_kernel_rtc_rtp.cpp (hunk 2 above)
Comment (duiniuluantanqin) [reply trunk/src/kernel/srs_kernel_rtc_rtp.cpp:1571]: Yes, I also believe that this macro switch can be removed.

File: trunk/src/protocol/srs_protocol_raw_avc.hpp (hunk 3 above)
Comment (duiniuluantanqin) [reply trunk/src/protocol/srs_protocol_raw_avc.hpp:106]: That's right, but there's no need to address it in this PR; because it involves other modules, such as SRT and GB28181.

File: trunk/src/kernel/srs_kernel_rtc_rtp.cpp (hunk 1 above)
Comment (duiniuluantanqin) [reply trunk/src/kernel/srs_kernel_rtc_rtp.cpp]: I do not quite understand what you mean.

Comment (suzp1984) [reply trunk/src/kernel/srs_kernel_rtc_rtp.cpp]: https://github.com/ossrs/srs/blob/0c88ddbcdf5b4b596a190f46d9c84f942b261efc/trunk/src/kernel/srs_kernel_codec.hpp#L389-L429
Here is the AVC nalu type definition in an enum.
And `kStapA` and `kFuA` can be consider RTP extension for AVC NAL unit.
https://github.com/ossrs/srs/blob/0c88ddbcdf5b4b596a190f46d9c84f942b261efc/trunk/src/kernel/srs_kernel_rtc_rtp.hpp#L30-L33

So can we move `kStapA` and `kFuA` into `enum SrsAvcNaluType`?

File: trunk/src/app/srs_app_rtc_source.cpp [hunk 4]
```diff
@@ -1481,38 +1505,42 @@ SrsRtcFrameBuilder::SrsRtcFrameBuilder(ISrsStreamBridge* bridge)
...
+    srs_freep(audio_transcoder_);
+    audio_transcoder_ = new SrsAudioTranscoder();
 
     SrsAudioCodecId from = SrsAudioCodecIdOpus; // TODO: From SDP?
     SrsAudioCodecId to = SrsAudioCodecIdAAC; // The output audio codec.
     int channels = 2; // The output audio channels.
     int sample_rate = 48000; // The output audio sample rate in HZ.
     int bitrate = _srs_config->get_rtc_aac_bitrate(r->vhost); // The output audio bitrate in bps.
-    if ((err = codec_->initialize(from, to, channels, sample_rate, bitrate)) != srs_success) {
+    if ((err = audio_transcoder_->initialize(audio_codec, to, channels, sample_rate, bitrate)) != srs_success) {
```
suzp1984, line 1775:
```suggestion
    SrsAudioCodecId from = audio_codec;
    if ((err = audio_transcoder_->initialize(from, to, channels, sample_rate, bitrate)) != srs_success) {
```
The srs only support `opus` as audio codec, but I also suggest add a audio codec check, make sure `from != to` or from != AAC.

trunk/src/app/srs_app_rtc_source.hpp
```diff
@@ -356,7 +359,13 @@ class SrsRtcFrameBuilder
 private:
     srs_error_t packet_video(SrsRtpPacket* pkt);
     srs_error_t packet_video_key_frame(SrsRtpPacket* pkt);
+
+    srs_error_t packet_sps_pps(SrsRtpPacket* pkt);
     srs_error_t packet_sps_pps(SrsRtpPacket* pkt, SrsSample* sps, SrsSample* pps);
+#ifdef SRS_H265
+    srs_error_t packet_vps_sps_pps(SrsRtpPacket* pkt);
+    srs_error_t packet_vps_sps_pps(SrsRtpPacket* pkt, SrsSample* vps, SrsSample* sps, SrsSample* pps);
```
suzp1984:
```suggestion
    srs_error_t packet_hevc_sh(SrsRtpPacket* pkt);
    srs_error_t packet_hevc_sh(SrsRtpPacket* pkt, SrsSample* vps, SrsSample* sps, SrsSample* pps);
```
those 2 methods are used to packet sequence header rtmp from hevc rtp, so I think theirs name deserve a better one.
vps|sps|pps can consider the hevc sequence header.

winlinvip: Sequence header is a name from FLV. 
1. I think `sequence_header` is better than `sh`.
2. If want to change the name, please also change `packet_sps_pps` for H.264.

trunk/src/kernel/srs_kernel_rtc_rtp.cpp
duiniuluantanqin: These are two separate ideas: one is about the format found in H264, and the other relates to the format used in RTP.

`TRANS_BY_GPT4`

trunk/src/app/srs_app_rtc_source.cpp
duiniuluantanqin: You are right, but it is not imperative for now. Let's revisit this when I come to implement RTMP with OPUS payload in the future.

duiniuluantanqin: You are right, but it is not necessary at the moment. When I eventually implement RTMP with OPUS payload support, I will comprehensively refine this logic.

trunk/src/app/srs_app_rtc_source.hpp
duiniuluantanqin: ok

suzp1984 (approved):

trunk/src/kernel/srs_kernel_rtc_rtp.cpp
winlinvip: Users can disable HEVC by turning off this feature, which is useful if there are issues with HEVC. However, I think we should remove this option because no third-party libraries are introduced, so enabling this feature should never fail.

trunk/src/app/srs_app_rtc_source.cpp
winlinvip: It is advisable to include a comment line here for future reference, as it may be forgotten otherwise.

`TRANS_BY_GPT4`

trunk/src/kernel/srs_kernel_rtc_rtp.cpp
duiniuluantanqin: Yeah

trunk/src/app/srs_app_rtc_source.cpp
duiniuluantanqin: ok

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -1956,6 +2135,44 @@ srs_error_t SrsRtcFrameBuilder::packet_video_rtmp(const uint16_t start, const ui
...
+                nalu_len += fua_payload_hevc->size;
+                payload.write_bytes(fua_payload_hevc->payload, fua_payload_hevc->size);
+                if (fua_payload_hevc->end) {
+                    //write nalu_len back
+                    payload.skip(-(4 + nalu_len));
+                    payload.write_4bytes(nalu_len);
+                    payload.skip(nalu_len);
+                }
+            }
+            srs_freep(pkt);
```
winlinvip: Is it possible to use SrsUniquePtr to automatically free this object?

duiniuluantanqin: no problem

winlinvip:
bugbot run

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2853,25 +3065,44 @@ SrsRtcVideoRecvTrack::~SrsRtcVideoRecvTrack()
...
+            *ppayload = new SrsRtpSTAPPayloadHevc();
+            *ppt = SrsRtpPacketPayloadTypeSTAPHevc;
+        } else if (v == kFuHevc) {
+            *ppayload = new SrsRtpFUAPayloadHevc2();
+            *ppt = SrsRtpPacketPayloadTypeFUAHevc2;
+        } else {
+            *ppayload = new SrsRtpRawPayload();
+            *ppt = SrsRtpPacketPayloadTypeRaw;
+        }
+#endif
```
winlinvip, line 3105: If not AVC or HEVC, the payload and pt is undefined. You can:
1. Use assert to make sure no problem.
2. Use SrsRtpPacketPayloadTypeRaw by default.

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -626,7 +626,14 @@ srs_error_t SrsRtcSource::on_publish()
     // If bridge to other source, handle event and start timer to request PLI.
     if (bridge_) {
 #ifdef SRS_FFMPEG_FIT
-        if ((err = frame_builder_->initialize(req)) != srs_success) {
+        SrsAudioCodecId audio_codec = srs_audio_codec_str2id(stream_desc_->audio_track_desc_->media_->name_);
```
winlinvip: If no audio, the stream_desc_->audio_track_desc_ will be null.

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2853,25 +3065,44 @@ SrsRtcVideoRecvTrack::~SrsRtcVideoRecvTrack()
...
+void SrsRtcVideoRecvTrack::on_before_decode_payload(SrsRtpPacket* pkt, SrsBuffer* buf, ISrsRtpPayloader** ppayload, SrsRtpPacketPayloadType* ppt)
 {
     // No payload, ignore.
     if (buf->empty()) {
         return;
     }
 
-    uint8_t v = (uint8_t)(buf->head()[0] & kNalTypeMask);
-    pkt->nalu_type = v;
+    SrsVideoCodecId video_codec = srs_video_codec_str2id(track_desc_->media_->name_);
```
winlinvip: For each received RTP packet by publisher stream, it will call this function once. This function is very heavy, as it compare string and get the code.

```
SrsVideoCodecId srs_video_codec_str2id(const std::string &codec)
{
    std::string upper_codec = codec;
    std::transform(upper_codec.begin(), upper_codec.end(), upper_codec.begin(), ::toupper);

    if (upper_codec == "H264" || upper_codec == "AVC") {
        return SrsVideoCodecIdAVC;
    } else if (upper_codec == "H265" || upper_codec == "HEVC") {
        return SrsVideoCodecIdHEVC;
    } else if (upper_codec == "AV1") {
        return SrsVideoCodecIdAV1;
    } else if (upper_codec == "VP6") {
        return SrsVideoCodecIdOn2VP6;
    } else if (upper_codec == "VP6A") {
        return SrsVideoCodecIdOn2VP6WithAlphaChannel;
    }

    return SrsVideoCodecIdReserved;
}
```

Better to cache the codec id in `SrsCodecPayload`.

trunk/src/app/srs_app_rtc_source.hpp
```diff
@@ -356,7 +359,13 @@ class SrsRtcFrameBuilder
 private:
     srs_error_t packet_video(SrsRtpPacket* pkt);
     srs_error_t packet_video_key_frame(SrsRtpPacket* pkt);
-    srs_error_t packet_sps_pps(SrsRtpPacket* pkt, SrsSample* sps, SrsSample* pps);
+
+    srs_error_t packet_sequence_header_avc(SrsRtpPacket* pkt);
+    srs_error_t packet_sequence_header_avc(SrsRtpPacket* pkt, SrsSample* sps, SrsSample* pps);
```
winlinvip: Do not use the function overload, confusing search, rename to `do_packet_sequence_header_avc`, or `packet_sequence_header_avc_impl` or `packet_sequence_header_avc2`.

trunk/src/app/srs_app_rtc_source.cpp
duiniuluantanqin: ok

trunk/src/app/srs_app_rtc_source.cpp
duiniuluantanqin: fixed

trunk/src/app/srs_app_rtc_source.hpp
duiniuluantanqin: ok

trunk/src/protocol/srs_protocol_raw_avc.hpp
winlinvip: Better to add a TODO comment for this.

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2286,6 +2261,9 @@ SrsCodecPayload::SrsCodecPayload(uint8_t pt, std::string encode_name, int sample
     pt_of_publisher_ = pt_ = pt;
     name_ = encode_name;
     sample_ = sample;
+    if ((codec_ = srs_video_codec_str2id(name_)) == SrsVideoCodecIdReserved) {
+        codec_ = srs_audio_codec_str2id(name_);
+    }
```
winlinvip: It's incorrect, consider a unknown video codec H.266, it will be parsed to SrsVideoCodecIdReserved by srs_video_codec_str2id, but it's still a video codec.

The cause is it lacks of information to parse the name to video or audio codec, so the codec must be defined as a function:

```
int8_t SrsCodecPayload::codec(bool video) 
```

trunk/src/app/srs_app_rtc_source.cpp
```diff
@@ -2254,22 +2255,40 @@ SrsCodecPayload::SrsCodecPayload()
...
+{
+    // Return cached value if already initialized
+    if (codec_ != -1) {
+        return codec_;
+    }
+
+    // Parse codec based on context (video or audio)
+    if (video) {
+        codec_ = srs_video_codec_str2id(name_);
+        // For unknown video codecs like H.266, still return SrsVideoCodecIdReserved
```
suzp1984, line 2449: I thought this comments is for line L2282, the video codec, so they should be placed above line L2282.

trunk/src/app/srs_app_rtc_conn.cpp
```diff
@@ -3132,7 +3132,8 @@ srs_error_t SrsRtcConnection::negotiate_play_capability(SrsRtcUserConfig* ruc, s
                 // Get the source codec if not specified.
                 std::vector<SrsRtcTrackDescription*> track_descs = source->get_track_desc("video", "");
                 if (!track_descs.empty()) {
-                    prefer_codec = srs_video_codec_str2id(track_descs.at(0)->media_->name_);
+                    SrsRtcTrackDescription* first_track = track_descs.at(0);
+                    prefer_codec = srs_video_codec_str2id(first_track->media_->name_);
```
suzp1984, line 3136: this commit do nothing? nullable check?

winlinvip: Never directly use `track_desc.at(0)->media_`, instead, should define a variable with class, so AI and human can read and review the code.

trunk/src/app/srs_app_rtc_source.cpp
duiniuluantanqin: Yeap

winlinvip (approved):

PR #4356, opened 2025-05-20
RTMP: Use extended timestamp as delta when chunk fmt=1/2. v6.0.167 v7.0.37

pzx601917159:
1. When the chunk message header employs type 1 and type 2, the extended timestamp denotes the time delta.
2. When the DTS (Decoding Time Stamp) experiences a jump and exceeds 16777215, there can be errors in DTS calculation, and if the audio and video delta differs, it may result in audio-video synchronization issues.

---------

`TRANS_BY_GPT4`

PR #4365, opened 2025-05-30
RTMP: Refactor RTMP extended timestamp. v7.0.52

winlinvip:
Refactor the extended timestamp code for #4356, and follow the below steps:

1. Read the timestamp from the buffer.
2. Read the extended timestamp if the timestamp indicates that an extended timestamp is present.
3. Skip back 4 bytes if following the RTMP v1 2009 version, which specifies that there is no extended timestamp for Type 3 chunks.
4. Generate the chunk timestamp, treating it as an absolute timestamp for Type 0, or as a delta timestamp for Type 1 or 2.

This improvement will be merged to develop branch only.


winlinvip:
Already merged, piggyback in https://github.com/ossrs/srs/commit/cc115afc1da6347b008b25903e56dd0b508807eb

PR #4366, opened 2025-05-31
Script: Use clang-format to unify the coding style. v7.0.38

chen-guanghua:
1. add clang-format config file
2. add clang_format.sh file, use to format cpp code before pr merged.


trunk/scripts/clang_format.sh
```diff
@@ -0,0 +1,27 @@
...
+# and format them using clang-format with the style defined in .clang-format file
+if [ ! -f ".clang-format" ]; then
+    echo ".clang-format file does not exist, please create one in the project root."
+    exit 1
+fi
+
+echo "Formatting source files in trunk directory..."
+# Exclude the 3rdparty directory and format all .cpp, .hpp, and .h files
+# Use -i to edit files in place
+find trunk/ -path trunk/3rdparty -prune -o -iname "*.cpp" -o -iname "*.hpp" -o -iname "*.h" | xargs clang-format -style=file -i
```
winlinvip: Maybe we only need to `find trunk/src` for source code of SRS.

.clang-format
```diff
@@ -0,0 +1,296 @@
...
+  Maximum:         -1
+SpacesInParens:  Never
+SpacesInParensOptions:
+  ExceptDoubleParentheses: false
+  InCStyleCasts:   false
+  InConditionalStatements: false
+  InEmptyParentheses: false
+  Other:           false
+SpacesInSquareBrackets: false
+Standard:        Latest
```
winlinvip: After clang format, it will change the code:

```
std::vector< std::vector<uint8_t> > sub_layer_profile_compatibility_flag;
```

To the below one:

```
std::vector<std::vector<uint8_t>> sub_layer_profile_compatibility_flag;
```

This will fail in C++98, for there must be space when template in another template. Need to change the clang format config to process this correctly.

AI suggests this:

```
Standard: Cpp03
```

chen-guanghua: modify .clang-format, set "Standard:        c++03",  compile ok

PR #4367, opened 2025-05-31
format cpp/h/hpp file use clang-format based llvm style

chen-guanghua:
use clang-format cpp/h/hpp file based llvm style


winlinvip:
After clang format, it will change the code:

```
std::vector< std::vector<uint8_t> > sub_layer_profile_compatibility_flag;
```

To the below one:

```
std::vector<std::vector<uint8_t>> sub_layer_profile_compatibility_flag;
```

This will fail in C++98, for there must be space when template in another template, so need a typedef like this:

```
typedef std::vector<uint8_t> U8Vec;
std::vector<U8Vec> sub_layer_profile_compatibility_flag;
```

winlinvip:
See https://github.com/ossrs/srs/pull/4433

PR #4372, opened 2025-06-04
issue#4364: try to fix memory leak for daemon run.

suzp1984:
This PR is try to workaround #4364. 
The memory continues increase problem didn't appear in the PR, but the root cause is not yet founded, by moving global initialize code into srs_pre_global_initialize(), It seems those 2 instructions may trigger this memory issue.

```
    _srs_log = new SrsFileLog();
    _srs_context = new SrsThreadContext();
```


winlinvip:
It doesn't make sense because they are two variables with fixed memory; they should not cause the memory to increase over time.

PR #4446, opened 2025-08-20
SRT2RTMP: fix srt bridge hevc to rtmp error. v7.0.84

### LINKED ISSUE #4428: Problems when converting H265 data from srt to rtmp
When using the srt protocol to push h265 data, an error occurs when using the rtmp protocol to pull h265 data, and the decoder cannot be found.

[flv @ 0x7fb949116440] Video codec (c) is not implemented. Update your FFmpeg version to the newest one from Git. If the problem still occurs, it means that your file has a feature which has not been implemented.

suzp1984:
try to fix #4428.

## Cause

rtmp do not support hevc, rtmp enhanced do.

## How to reproduce

1. start srs.
   `./objs/srs -c conf/srt.conf`
2. publish hevc (h.265) stream to srs by srt.
    `ffmpeg -re -i ./doc/source.flv   -c:v libx265 -crf 28 -preset medium   -c:a copy   -pes_payload_size 0 -f mpegts   'srt://127.0.0.1:10080?streamid=#!::r=live/livestream,m=publish'`
3. probe the rtmp stream
   `ffprobe rtmp://localhost/live/livestream`

## About the Failed BlackBox test
The failed blackbox test: `TestSlow_SrtPublish_RtmpPlay_HEVC_Basic` `TestSlow_SrtPublish_HttpFlvPlay_HEVC_Basic`

### Cause: 

The ffmpeg 5 is used to record a piece of video (DRV), the ffmpeg will transcode the enhanced flv format to TS format, but ffmpeg 5 don't support enhanced rtmp (or flv) in this case.

The solution is to replace the ffmpeg to version 7 in those 2 test cases.

### why not upgrade ffmpeg to version 7?

The black tests dependency on ffmpeg 5 will fail, and there are a few of them are not easy to resolve in ffmpeg 7.

### DIFF
--- a/trunk/3rdparty/srs-bench/blackbox/hevc_test.go
+++ b/trunk/3rdparty/srs-bench/blackbox/hevc_test.go
@@ -709,6 +709,8 @@ func TestSlow_SrtPublish_RtmpPlay_HEVC_Basic(t *testing.T) {
 		v.dvrFile = path.Join(svr.WorkDir(), "objs", fmt.Sprintf("srs-ffprobe-%v.ts", streamID))
 		v.streamURL = fmt.Sprintf("rtmp://localhost:%v/live/%v", svr.RTMPPort(), streamID)
 		v.duration, v.timeout = duration, time.Duration(*srsFFprobeTimeout)*time.Millisecond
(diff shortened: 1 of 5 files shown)


winlinvip:
Feel free to upgrade [dev-docker](https://github.com/ossrs/dev-docker) to FFmpeg 7 and Ubuntu 22/24.

PR #4454, opened 2025-08-27
AI: Config: Move RTMP configs to rtmp{} section. v7.0.64

winlinvip:
This PR reorganizes SRS configuration structure by moving RTMP-specific configurations from global scope to a dedicated `rtmp {}` section, and includes various cleanups.

**Before (SRS 6.x):**

```nginx
listen 1935;
chunk_size 60000;
max_connections 1000;
```

**After (SRS 7.0+):**

```nginx
max_connections 1000;
rtmp {
    listen 1935;
    chunk_size 60000;
}
```

Cleanup:

* Removed unused threads_interval configuration and related code
* Cleaned up reload handlers and removed obsolete functionality


HolgerHuo:
Hi! In this PR, many functions like DVR and recording were marked as not recommended. Is it because that forking FFMPEG processes would be more efficient or something else? When dealing with live streams, when some incorrect packets were sent, ffmpeg might easily exit on error, are there any approaches to solve this problem?

PR #4466, opened 2025-09-02
AI: GB28181: Remove embedded SIP server and enforce external SIP usage

winlinvip:
This PR removes the embedded GB28181 SIP server implementation from SRS and enforces the use of external SIP servers for production deployments.

The embedded SIP server depended on the deprecated `http-parser` library. With the planned migration to `llhttp` (which doesn't support SIP parsing), maintaining the embedded SIP server would require significant additional work. Since external SIP servers are already the recommended approach for production, removing the embedded implementation simplifies the codebase and eliminates this dependency.

Eliminated `srs_gb28181_test` from CI workflow.

Removed SIP configuration validation tests.


trunk/conf/gb28181.conf
```diff
@@ -14,10 +14,7 @@ stream_caster {
     output rtmp://127.0.0.1/live/[stream];
     listen 9000;
     sip {
```
duiniuluantanqin: This `sip` configuration can be deleted now, right?

`TRANS_BY_GPT4`

trunk/src/app/srs_app_gb28181.cpp
```diff
@@ -31,7 +31,6 @@ using namespace std;
 
 // See https://www.ietf.org/rfc/rfc3261.html#section-8.1.1.7
 #define SRS_GB_BRANCH_MAGIC "z9hG4bK"
```
duiniuluantanqin: SRS_GB_BRANCH_MAGIC unused.

duiniuluantanqin:
The tests were successful in conjunction with the https://github.com/ossrs/srs-sip.

`TRANS_BY_GPT4`

duiniuluantanqin (approved):

PR #4475, opened 2025-09-06
AI: Support anonymous coroutine with code block. v7.0.80

winlinvip:
This PR introduces anonymous coroutine macros for easier coroutine creation and improves the State Threads (ST) mutex and condition variable handling in SRS.

- **Added coroutine macros**: `SRS_COROUTINE_GO`, `SRS_COROUTINE_GO2`, `SRS_COROUTINE_GO_CTX`, `SRS_COROUTINE_GO_CTX2`
- **Added `SrsCoroutineChan`**: Channel for sharing data between coroutines with coroutine-safe operations
- **Simplified coroutine creation**: Go-like syntax for creating anonymous coroutines with code blocks

---------

### DIFF
--- a/trunk/src/protocol/srs_protocol_st.hpp
+++ b/trunk/src/protocol/srs_protocol_st.hpp
@@ -84,6 +84,45 @@ extern int srs_mutex_destroy(srs_mutex_t mutex);
 extern int srs_mutex_lock(srs_mutex_t mutex);
 extern int srs_mutex_unlock(srs_mutex_t mutex);
 
+// Wrap as ptr, so you can use SrsUniquePtr and SrsSharedPtr to manage it.
+// For example:
+//      SrsUniquePtr<SrsCond> cond(new SrsCond());
+//      cond->signal();
+class SrsCond
+{
+private:
+    srs_cond_t cond_;
+
+public:
+    SrsCond();
+    virtual ~SrsCond();
+
+public:
+    int wait();
+    int timedwait(srs_utime_t timeout);
+    int signal();
+    int broadcast();
+};
+
+// Wrap as ptr, so you can use SrsUniquePtr and SrsSharedPtr to manage it.
+// For example:
+//      SrsUniquePtr<SrsMutex> mutex(new SrsMutex());
+//      SrsLocker(mutex->get());
+class SrsMutex
+{
+private:
+    srs_mutex_t mutex_;
+
+public:
+    SrsMutex();
+    virtual ~SrsMutex();
+
+public:
+    int lock();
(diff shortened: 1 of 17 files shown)


trunk/src/utest/srs_utest.hpp
```diff
@@ -131,4 +131,147 @@ class MockProtectedBuffer
...
+    SrsSTCoroutine st##id("anonymous", &handler##id);                             \
+    srs_error_t err_coroutine##id = st##id.start();                               \
+    srs_assert(err_coroutine##id == srs_success)
+
+// A helper to create a anonymous coroutine like goroutine in Go.
+// For example:
+//        SRS_COROUTINE_GO({
+//            srs_usleep(1 * SRS_UTIME_MILLISECONDS);
+//        });
+#define SRS_COROUTINE_GO(code_block) \
```
suzp1984, line 232: is the Macro input var `id` missing?
```suggestion
#define SRS_COROUTINE_GO(id, code_block) \
```

That's a duplicate of SRS_COROUTINE_GO2.

trunk/src/protocol/srs_protocol_st.hpp
```diff
@@ -109,7 +148,7 @@ extern bool srs_is_never_timeout(srs_utime_t tm);
 
 // The mutex locker.
 #define SrsLocker(instance) \
-    impl__SrsLocker _SRS_free_##instance(&instance)
+    impl__SrsLocker _SRS_free_instance(instance)
```
suzp1984, line 151: if this case exist:
```
{
    SrsLocker(mutex1);
    SrsLocker(mutex2);
}
```
I mean is the token pasting still needed?
```suggestion
    impl__SrsLocker _SRS_free_##instance(instance)
```

winlinvip: If change to:
```
impl__SrsLocker _SRS_free_##instance(instance)
```

You can not use unique ptr like this:
```
SrsLocker(p.get())
```

The only way is to define a variable:
```
srs_mutex_t lock = *p.get();
SrsLocker(lock);
```

I think if we need to support two lockers, we can define a new macro:
```
#define SrsLocker(lock1, lock2) \
    impl__SrsLocker _SRS_free_instance(lock1); \
    impl__SrsLocker _SRS_free_instance(lock2)
```

Currently I don't see any use case for two locks.

trunk/src/utest/srs_utest.hpp
winlinvip: This is defined for usecase for just one coroutine.

PR #4520, opened 2025-10-06
RTC2RTMP: fix illegal memory access. v7.0.97

JackLau1222:
Regression since 20f6cd595cc8be63a824d29d1332abd4488685be

The early code might meet bridge is empty when
there is no bridge(e.x. rtc to rtc). Then srs_freep will free the brige.

Remove this code that seems redundant.

---------


suzp1984:
This PR make sense.

suzp1984 (approved):

PR #4521, opened 2025-10-06
srs_app_rtc_api: support non BUNDLE of SDP

JackLau1222:
The SDP might use non BUNDLE when there is only
one stream


suzp1984:
I thank this PR do not make sense, SRS only used one tcp or udp connect peer to transfer multi media tracks, that's the reason limit the SDP to Bundle policy. 
You can't just remove this restriction because of this case: when there is only one media track, the SDP policy may be none bundle anymore.

JackLau1222:
> I thank this PR do not make sense, SRS only used one tcp or udp connect peer to transfer multi media tracks, that's the reason limit the SDP to Bundle policy. You can't just remove this restriction because of this case: when there is only one media track, the SDP policy may be none bundle anymore.

Agree, closing this PR

PR #4524, opened 2025-10-09
optimize: inefficient srs_avc_startswith_annexb

### LINKED ISSUE #4522: [Cannot Fix] [SRT]: Server hang when publishing with QSV/OBS encoder
**Describe the bug**

```nginx
srt_server {
    enabled on;
    listen 10080;
}

vhost __defaultVhost__ {
    srt {
        enabled on;
        srt_to_rtmp on;
    }
}
```

When SRT to RTMP is on and incoming streams isn't very protocol compliant ( e.g., QSV H264 on OBS, the produced stream may not contain SPS and IDR on first few packets, also VAAPI/QSV HEVC on OBS), SRS will refuse to process those packets with logs complaining SPS or PPS empty. This might not be an issue but SRS will also hang for other requests. From my observation, SRS will eat up 3 cores (AMD EPYC 7453) and all other requests (SRT/RTMP playback hangs, HTTP requests hang and delay)

**Version**
SRS/6.0.177(Hang) in Docker

**To Reproduce**
1. Setup SRT2RTMP
2. Publish stream using QSV encoder on OBS
3. Check SRS CPU Usage
4. Test HTTP connectivity to SRS HTTP Endpoints

**Expected behavior**
Server should return error to those non-compliant streams and not block other requests, or, when I setup an SRT listener using FFPlay (`ffprobe -mode listener -v debug -show_packets -i srt://:12345 `), although FFMPEG also complains for no sps on first few packets, the stream is also playable after a few seconds. Maybe srs should also try to handle these cases, at least not hang and block other requests. 

**Screenshots**

SRS Log

SRS CPU Usage

**Additional context**
From my testing, QSV H264/H265 on OBS, VA-API H265 on OBS has these issues. QSV/VA-API on FFMPEG doesn't. My guess is that OBS may be lagging due to some reasons and not producing expected results, I'm working on fix this bug in OBS or also open an issue their side.

HolgerHuo:
Partially fixes #4522 .

I profiled SRS server during hang and found that `srs_avc_startswith_annexb` took up most of the CPU time:

image

After examining the code, I found that the current implementation of `srs_avc_startswith_annexb` seems inefficient. Annex B specifies that bitstream begins with 00 00 00 01 or 00 00 01. There is no need to check for N x 00 + 01, which would require the buffer much more time. 

One thing yet to be explored is that this solved problem with hardware encoding of HEVC with OBS, but not QSV H264 with OBS (although it wont cause the server to hang anymore ).  When playing with ffplay (`ffplay -mode listener -i srt://:12345`), video will eventually start to play when there is SPS in stream; however, with SRS, SPS will never come in. I don't think OBS will change its behavior with different behaviors but maybe there's some special logic in ffmpeg that handles this. 

I developed on R6 branch. In develop branch, this function was moved to another file. Should I update this pr, please let me know.

### DIFF
--- a/trunk/src/kernel/srs_kernel_utility.cpp
+++ b/trunk/src/kernel/srs_kernel_utility.cpp
@@ -725,26 +725,32 @@ bool srs_avc_startswith_annexb(SrsBuffer* stream, int* pnb_start_code)
     
     char* bytes = stream->data() + stream->pos();
     char* p = bytes;
+
+    // h264 annexb start codes are either 00 00 01 or 00 00 00 01
+    if (!stream->require(3)) {
+        return false;
+    }
     
-    for (;;) {
-        if (!stream->require((int)(p - bytes + 3))) {
-            return false;
+    // first check if starts with 00 00
+    if (p[0] != (char)0x00 || p[1] != (char)0x00) {
+        return false;
+    }
+    
+    // check for 00 00 01
+    if (p[2] == (char)0x01) {
+        if (pnb_start_code) {
+            *pnb_start_code = 3;
         }
-        
-        // not match
-        if (p[0] != (char)0x00 || p[1] != (char)0x00) {
-            return false;
+        return true;
+    }
+    
+    // check for 00 00 00 01
+    // only check when we have enough bytes
+    if (stream->require(4) && p[2] == (char)0x00 && p[3] == (char)0x01) {
+        if (pnb_start_code) {
+            *pnb_start_code = 4;
         }
-        
-        // match N[00] 00 00 01, where N>=0
-        if (p[2] == (char)0x01) {
-            if (pnb_start_code) {
-                *pnb_start_code = (int)(p - bytes) + 3;
-            }
-            return true;
-        }
-        
-        p++;
+        return true;
     }
     
     return false;


winlinvip:
Thank you for investigating the performance issue and submitting this PR! However, after carefully reviewing the H.264 Annex B specification, I believe the current implementation is actually **correct** and this PR would break specification compliance.

## H.264 Annex B Specification

According to the H.264 standard (ISO/IEC 14496-10) Annex B, the byte stream NAL unit syntax is defined as:

```
byte_stream_nal_unit( NumBytesInNALunit ) {
    while( next_bits( 24 ) != 0x000001 && next_bits( 32 ) != 0x00000001 )
        leading_zero_8bits /* equal to 0x00 */
    if( next_bits( 24 ) != 0x000001 )
        zero_byte /* equal to 0x00 */
    start_code_prefix_one_3bytes /* equal to 0x000001 */
    nal_unit( NumBytesInNALunit )
    ...
}
```

**Key points from the specification:**

1. **`leading_zero_8bits`**: The specification explicitly allows **any number of leading 0x00 bytes** before the start code prefix
2. **`zero_byte`**: An additional optional 0x00 byte before the 3-byte start code prefix
3. **Valid start codes include:**
   - `00 00 01` (3 bytes)
   - `00 00 00 01` (4 bytes)
   - `00 00 00 00 01` (5 bytes)
   - `00 00 00 00 00 01` (6 bytes)
   - And so on... (N×0x00 + 00 01, where N≥0)

**Reference**: You can find this syntax definition in:
- ISO/IEC 14496-10 Annex B (Byte stream format)

## Why the Current Implementation is Correct

The current SRS implementation correctly handles the specification:

```cpp
for (;;) {
    if (!stream->require((int)(p - bytes + 3))) {
        return false;
    }
    
    // not match
    if (p[0] != (char)0x00 || p[1] != (char)0x00) {
        return false;
    }
    
    // match N[00] 00 00 01, where N>=0
    if (p[2] == (char)0x01) {
        if (pnb_start_code) {
            *pnb_start_code = (int)(p - bytes) + 3;
        }
        return true;
    }
    
    p++;  // Skip one 0x00 and continue looking
}
```

This correctly accepts:
- `00 00 01` ✅
- `00 00 00 01` ✅
- `00 00 00 00 01` ✅
- Any number of leading zeros ✅

## Why This PR Would Break Compliance

Your proposed change only accepts exactly `00 00 01` or `00 00 00 01`, which would **reject valid H.264 streams** that have more than one leading zero byte. This would break compatibility with:
- Encoders that add multiple leading zeros for alignment or other purposes
- Valid H.264/HEVC streams that follow the Annex B specification
- Certain broadcast and professional video equipment

HolgerHuo:
You are right. I overlooked the usage of this function. 

Do you think the issue in #4522 is caused by this function? seems the original implementation should be working...

PR #4545, opened 2025-10-29
Increase the file descriptor limit at startup to support high concurrency configurations

Jason-JP-Yang:
Solve the leftover problem in /etc/init.d/srs
- [X] # TODO: FIXME: set limit by, for instance, "ulimit -HSn 10000"


Jason-JP-Yang:
Without this fix, when I start srs through systemctl `sudo systemctl start srs`, error occurs:
``` logs
[2025-10-29 13:52:44.066][ERROR][45124][4870w333][0] max_connections=1000, system limit to 1024, please run: ulimit -HSn 10000
[2025-10-29 13:52:44.066][ERROR][45124][4870w333][0] Failed, code=1023(ConfigInvalid)(Configuration is invalid) : check config : check connections: 1127 exceed max open files=1024
```

Issue #4593, opened 2025-11-25
[Proposal] SRT: short streamid format


artem-smorodin-dacast:
I would like to suggest making it possible to use a short format for streamid, for example `srt://127.0.0.1:10808?streamid=testkey`, which would correspond to `srt://127.0.0.1:10808?streamid=#!::r=DEFAULT_APP/testkey,m=publish`

This would allow SRS users to use a shorter, simpler, and more attractive streamid format, which would ultimately be easier for end users to understand and remember.

To make this possible, we could add a default_mode option (with default = publish) and modify streamid parser so that if streamid does not start with `#!::`, we could consider its content as a stream key.

@winlinvip  Please share your thoughts. If you agree, I will send PR with these changes.

winlinvip:
Does it follow any standard of SRT? Is it compatible with FFmpeg and SRT tools?

artem-smorodin-dacast:
Yes, it is compatible because you can actually pass any string there. It is also compatible with ffmpeg and srt tools for the same reason. Each media server actually implements its own format. For example

mediamtx - `?streamid=publish:{streamkey}`
ovenmediaengine - `?streamid={vhost}/{app}/{streamkey}`
ant media server - `?streamid={app}/{streamkey}`
srt live server - `?streamid={vhost}/{app}/{streamkey}`
mist server - `?streamid={streamkey}`

winlinvip:
First, all these samples are implemented from other SRT servers, rather than something defined by the SRT standards. Second, if it's defined by SRT standard, please show me the exact text from the standard. Finally, it will be helpful to see that FFmpeg can handle the short SID format, by trying to connect to these SRT servers and verify whether it works.

artem-smorodin-dacast:
> Second, if it's defined by SRT standard, please show me the exact text from the standard

Based on [Latest IETF RFC Draft](https://tools.ietf.org/html/draft-sharabayko-srt-00#section-3.2.1.3)

> The Stream ID handshake extension message can be used to identify the stream content.  The Stream ID **value can be free-form**, but there is also a recommended convention that can be used to achieve interoperability.

SRS follows the recommendations with the full format, but the wording "value can be free-form" definitely gives us the option to use the short format in addition.

> Finally, it will be helpful to see that FFmpeg can handle the short SID format, by trying to connect to these SRT servers and verify whether it works.

[...]

PR #4597, opened 2025-11-28
SRT: Short streamid format support

artem-smorodin-dacast:
PR for #4593 

Based on [Latest IETF RFC Draft](https://tools.ietf.org/html/draft-sharabayko-srt-00#section-3.2.1.3)

> The Stream ID handshake extension message can be used to identify the stream content. The Stream ID value can be free-form, but there is also a recommended convention that can be used to achieve interoperability.

Unit tests updated, documentation updated


winlinvip:
Thank you, but just let AI and me to do the task.

winlinvip:
Closed by #4598

Issue #865, opened 2017-04-30
HLS: 服务器退出或重启时不应该清除已生成TS切片


wuxianlijiang:
版本：2.0.239 发行版 
环境：centos 6
配置：
```
vhost dvr.host.com{
    hls {
        enabled         on;
        hls_path        /home/video/rec/live;
		hls_ts_file   [stream]/[stream]-[seq].ts;
		hls_m3u8_file  [stream].m3u8;
		hls_fragment    10;
		hls_window      86400;
		#hls_window      60;
		hls_wait_keyframe       on;
		hls_dispose     0;
		hls_cleanup     off;
    }
}
```
情况：SRS服务被重启后，已经录制好的m3u8文件和ts文件都会被清除，只会留下一个空的ts目录。
预期：无论停止推流，还是重启了SRS，已录制的数据都不会被清除。谢谢！

wuxianlijiang:
Reproduction of the situation:
1. Use Adobe Flash Media Live Encoder (or other encoders) to stream.
url: rtmp://domain.com/live   stream:live0501AM
Later, you can find live0501AM.m3u8 and the live0501AM folder with ts files inside at /home/video/rec/live.

2. Attempting to stop and disconnect the live streaming will not delete the files generated above.

3. Enter `service srs restart` in the shell, 
you can see that the m3u8 files generated in the above directory have been cleared, and the ts files under live0501AM have also been cleared.

4. If you terminate the service with `killall srs`, the situation will be the same as in step 3.

`TRANS_BY_GPT3`

winlinvip:
Already reproduced:
```
[2017-05-01 16:28:12.156][trace][22819][104] user terminate program, gracefully quit.
[2017-05-01 16:28:13.029][trace][22819][100] cleanup for gracefully terminate.
[2017-05-01 16:28:13.029][warn][22819][100][35] main cycle terminated, system quit normally.
[2017-05-01 16:28:13.031][trace][22819][100] gracefully dispose hls /live/livestream
[2017-05-01 16:28:13.031][trace][22819][100] srs terminated
```

It should be that Dispose was not disabled, resulting in cleanup.

`TRANS_BY_GPT3`

winlinvip:
When `killall srs` or `/etc/init.d/srs stop` or `/etc/init.d/srs restart` is executed, it will attempt to clean the Server. At this time, it will unpublish the stream, clean temporary files, and then clean HLS (which means deleting all TS files).

[...]

PR #4618, opened 2025-12-31
HLS/DASH: Fix dispose() to cleanup files after unpublish

ppeeou:
# HLS/DASH: Fix dispose() to cleanup files after unpublish

## Summary
Fixes a bug where HLS/DASH files are not deleted after the configured `hls_dispose`/`dash_dispose` timeout.

## Problem
When a stream is unpublished:
1. `on_unpublish()` is called and sets `enabled_ = false`
2. After the dispose timeout, `cycle()` calls `dispose()`
3. `dispose()` immediately returns due to `if (!enabled_)` check at line 2722 (HLS) and line 891 (DASH)
4. `controller_->dispose()` is never called
5. Files remain on disk indefinitely

**Observed behavior**:
- Stream stopped at 11:32:42
- `dispose()` called at 11:33:14 (after 30s timeout)
- Log shows "hls cycle to dispose hls" but no "gracefully dispose hls" message
- Files remain on disk

## Root Cause
Commit 550760f2d introduced an early return in `dispose()` when `!enabled_`, which prevents file cleanup after `on_unpublish()` has already been called and set `enabled_` to false.

## Solution
Reorder the logic in `dispose()` to:
1. Check if dispose is enabled (hls_dispose/dash_dispose > 0) first
2. Call `on_unpublish()` only if `enabled_` is still true (prevents duplicate calls)
3. Always call `controller_->dispose()` to cleanup files when dispose timeout occurs

This ensures files are properly cleaned up while still preventing duplicate `on_unpublish()` calls.

## Changes Made
- **trunk/src/app/srs_app_hls.cpp** (lines 2718-2734): Reordered dispose() logic
- **trunk/src/app/srs_app_dash.cpp** (lines 887-902): Reordered dispose() logic
- **trunk/doc/CHANGELOG.md**: Added v7.0.137 entry

## Testing Recommendation

To verify the fix:

1. Start RTMP stream to `/live/test`:
   ```bash
   ffmpeg -re -i test.mp4 -c copy -f flv rtmp://localhost:1935/live/test
   ```

2. Wait for HLS segments to be created:
   ```bash
   ls -la /path/to/hls/live/test/
   ```

3. Stop the stream (Ctrl+C)

4. Wait for `hls_dispose` timeout (default 120s, or 30s with your config):
   ```bash
   # Watch logs for "hls cycle to dispose hls" and "gracefully dispose hls"
   tail -f srs.log
   ```

5. Verify files are deleted:
   ```bash
   ls -la /path/to/hls/live/test/
   # Should be empty or directory removed
   ```

**Expected results**:
- Before fix: Files remain on disk
- After fix: Files are deleted, logs show "gracefully dispose hls"

## Impact
- **Risk**: Low - minimal logic change, only reordering of checks
- **Breaking changes**: None
- **Performance**: No impact
- **Compatibility**: Fixes existing bug, improves expected behavior

## Checklist
- [x] Code follows project style
- [x] Both HLS and DASH are fixed
- [x] CHANGELOG updated
- [x] Tested locally (recommended before merge)
- [x] No breaking changes

## Related Issues
- Regression introduced in: 550760f2d
- Related to: #865 (hls_dispose feature)

### DIFF
--- a/trunk/src/utest/srs_utest_ai17.cpp
+++ b/trunk/src/utest/srs_utest_ai17.cpp
@@ -1389,6 +1389,17 @@ VOID TEST(DashTest, LifecycleInitializeDisposeCleanupDelay)
     mock_controller->on_unpublish_called_ = false;
     mock_controller->dispose_called_ = false;
 
+    // Test dispose() when not enabled but dash_dispose is non-zero - should call controller->dispose() but not on_unpublish
+    dash->enabled_ = false;
+    mock_config->dash_dispose_ = 120 * SRS_UTIME_SECONDS;
+    dash->dispose();
+    EXPECT_FALSE(mock_controller->on_unpublish_called_); // enabled is false, so on_unpublish should not be called
+    EXPECT_TRUE(mock_controller->dispose_called_); // dash_dispose is non-zero, so controller->dispose() should be called
+
+    // Reset flags for next test
+    mock_controller->on_unpublish_called_ = false;
+    mock_controller->dispose_called_ = false;
+
     // Test dispose() when enabled and dash_dispose is non-zero - should call both on_unpublish and controller->dispose()
     dash->enabled_ = true;
     mock_config->dash_dispose_ = 120 * SRS_UTIME_SECONDS;
--- a/trunk/src/utest/srs_utest_ai17.cpp
+++ b/trunk/src/utest/srs_utest_ai17.cpp
@@ -1533,6 +1533,76 @@ VOID TEST(DashTest, PublishLifecycleWithAudioVideo)
     srs_freep(mock_hub);
 }
 
+// Test SrsHls dispose behavior after disabled
+// This test covers dispose() logic when enabled_ is false but hls_dispose is configured
+VOID TEST(HlsTest, LifecycleDisposeAfterDisabled)
+{
+    srs_error_t err;
+
+    // Create SrsHls object
+    SrsUniquePtr<SrsHls> hls(new SrsHls());
+
+    // Create mock dependencies
+    SrsUniquePtr<MockAppConfig> mock_config(new MockAppConfig());
+    SrsUniquePtr<MockSrsRequest> mock_req(new MockSrsRequest("test.vhost", "live", "livestream"));
+    MockHlsController *mock_controller = new MockHlsController();
+    MockOriginHub *mock_hub = new MockOriginHub();
+
+    // Inject mock config into SrsHls
+    hls->config_ = mock_config.get();
+
+    // Inject mock controller into SrsHls
+    srs_freep(hls->controller_);
+    hls->controller_ = mock_controller;
+
+    // Inject mock request
+    hls->req_ = mock_req.get();
+
+    // Test dispose() when not enabled and hls_dispose is 0 - should not call on_unpublish or controller->dispose()
+    mock_config->hls_dispose_ = 0;
+    hls->dispose();
+    EXPECT_EQ(0, mock_controller->on_unpublish_count_);
+    EXPECT_EQ(0, mock_controller->dispose_count_); // hls_dispose is 0, so controller->dispose() should not be called
+
+    // Test dispose() when enabled but hls_dispose is 0 - should call on_unpublish but not controller->dispose()
+    hls->enabled_ = true;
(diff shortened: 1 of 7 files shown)


suzp1984:
I believe this PR make sense.

suzp1984:
One more thing, Let the utest to cover this case.

e.g.

https://github.com/ossrs/srs/blob/93c5d7225b954c1b9013e52003d6c09935bcf3d5/trunk/src/utest/srs_utest_ai17.cpp#L1381-L1387

To cover when dash/hls is `enalbed_ == false`, the dispose also called.

ppeeou:
Thanks for the review feedback!

I've added the test case for DASH to cover `enabled_ == false` with `dash_dispose > 0`
The test now covers all 4 combinations for DASH dispose behavior and passes successfully.

However, I noticed there's no equivalent lifecycle/dispose test for HLS like `DashTest.LifecycleInitializeDisposeCleanupDelay`. 

**Question**: Should I create a new `HlsTest.LifecycleInitializeDisposeCleanupDelay` test with the same coverage, or is the DASH test sufficient since both DASH and HLS have identical dispose logic?

suzp1984:
> Thanks for the review feedback!
> 
> I've added the test case for DASH to cover `enabled_ == false` with `dash_dispose > 0` The test now covers all 4 combinations for DASH dispose behavior and passes successfully.
> 
> However, I noticed there's no equivalent lifecycle/dispose test for HLS like `DashTest.LifecycleInitializeDisposeCleanupDelay`.
> 
> **Question**: Should I create a new `HlsTest.LifecycleInitializeDisposeCleanupDelay` test with the same coverage, or is the DASH test sufficient since both DASH and HLS have identical dispose logic?

I think the independent `HLS` related utest is necessary, because they are independent codes.
Just cover what you commit is OK.

My suggest put the code in the `srs_utest_ai17.cpp` is OK.
The Test Case name maybe: `HlsTest.LifecycleDisposeAfterDisabled`.

trunk/src/utest/srs_utest_manual_mock.hpp
```diff
@@ -562,7 +564,7 @@ class MockAppConfig : public ISrsAppConfig
     virtual srs_utime_t get_hls_window(std::string vhost) { return 60 * SRS_UTIME_SECONDS; }
     virtual std::string get_hls_on_error(std::string vhost) { return "continue"; }
     virtual bool get_hls_cleanup(std::string vhost) { return true; }
-    virtual srs_utime_t get_hls_dispose(std::string vhost) { return 120 * SRS_UTIME_SECONDS; }
+    virtual srs_utime_t get_hls_dispose(std::string vhost) { return hls_dispose_; }
```
suzp1984, line 567: the codes in `srs_utest_manual_mock.hpp` is irrelevent? is it?

trunk/src/utest/srs_utest_ai17.cpp
```diff
@@ -1522,6 +1533,76 @@ VOID TEST(DashTest, PublishLifecycleWithAudioVideo)
     srs_freep(mock_hub);
 }
 
+// Test SrsHls dispose behavior after disabled
+// This test covers dispose() logic when enabled_ is false but hls_dispose is configured
+VOID TEST(HlsTest, LifecycleDisposeAfterDisabled)
+{
+    srs_error_t err;
```
suzp1984: unused variable, can be deleted.

trunk/src/utest/srs_utest_manual_mock.hpp
suzp1984: Oh, sorry, this patch make sense, the hls_dispose config mock is used in HlsTest.LifecycleDisposeAfterDisabled.

trunk/src/utest/srs_utest_ai17.cpp
suzp1984: One more thing, could you grant the edit permissions for the project maintainers for this PR? So I can push commits, e.g. format the CHANGELOGS.

ppeeou: @suzp1984 
Thank you for checking the PR

I'm sorry for the delay due to work-related issues.
I've added the permissions!

image

suzp1984 (approved):
Approved and ready to merge.

PR #4635, opened 2026-02-11
WebRTC: counting video frames for stats instead of packets

artem-smorodin-dacast:
This PR allows counting webrtc video frames instead of RTP packets. This would be very useful, as the number of RTP packets is not a very useful metric on its own, especially due to NACK.

Technically, this is possible because the marker bit is set for the last packet of the frame.

The count is only done for the primary ssrc, ignoring RTX and FEC.


winlinvip:
Thank you for your idea, it's a value improvement. I appreciate your contribution. Now SRS is maintained by AI, so AI will implement this idea with comprehensive tests. See https://github.com/ossrs/srs/pull/4717

PR #4640, opened 2026-02-28
RTC: Fix missing CRLF for SSRC group in SDP

thinker132203:
Hello SRS team,

During my recent experimental development to support downlink RTX/FEC (which requires negotiating a=ssrc-group:FID/FEC to players), I found a formatting bug in the SDP generation logic.

In  trunk/src/protocol/srs_protocol_sdp.cpp, the SrsSSRCGroup::encode function lacks a trailing kCRLF.
If ssrc_groups_ is not empty during downlink SDP generation, it will generate concatenated SDP lines like:
a=ssrc-group:FID 111 222a=ssrc:111 cname:...
This causes WebRTC clients (e.g., Chrome) to fail to parse the remote answer SDP.


duiniuluantanqin (approved):

winlinvip:
Already fixed.

PR #4649, opened 2026-03-18
Proxy: Fix memory load balancer server expiration handling

liukun4515:
## Summary

Fixes a critical issue in memory-based load balancer where streams could not recover when origin servers become unhealthy.

## Changes

- **Fixed `MemoryLoadBalancer.Pick()`**: Now automatically switches to healthy servers when current server's `UpdatedAt` exceeds `ServerAliveDuration` (300s)
- **Added double-checked locking**: Ensures thread safety during concurrent server selection
- **Added comprehensive unit tests**: Verifies server expiration handling and concurrent access safety

## Problem

Previously, once a stream was assigned to a server, it would remain bound to that server forever, even if the server became unhealthy (no heartbeat updates for >300s). This prevented streams from recovering when origin servers failed.

## Solution

The `Pick()` method now:
1. Checks if previously picked server is still healthy (fast path, no lock)
2. If unhealthy, acquires lock and re-checks (double-checked locking pattern)
3. Selects a new healthy server if needed

## Testing

- ✅ All unit tests pass
- ✅ Go race detector shows no race conditions
- ✅ Concurrent access tests verify thread safety (100 goroutines)
- ✅ Server expiration tests verify automatic failover

## Files Changed

- `proxy/internal/lb/mem.go`: Added health checking and mutex-based locking
- `proxy/internal/lb/mem_test.go`: Added 5 comprehensive test cases (286 lines)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/proxy/internal/lb/mem.go
+++ b/proxy/internal/lb/mem.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	stdsync "sync"
 	"time"
 
 	"proxy/internal/env"
@@ -23,6 +24,8 @@ type MemoryLoadBalancer struct {
 	servers sync.Map[string, *SRSServer]
 	// The picked server to service client by specified stream URL, key is stream url.
 	picked sync.Map[string, *SRSServer]
+	// Mutex to protect the Pick operation when reselecting servers.
+	pickMutex stdsync.Mutex
 	// The HLS streaming, key is stream URL.
 	hlsStreamURL sync.Map[string, HLSPlayStream]
 	// The HLS streaming, key is SPBHID.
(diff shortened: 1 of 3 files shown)


liukun4515:
@winlinvip  PTAL

winlinvip:
Thanks for the PR. This behavior is already configurable in current `develop` through `PROXY_ORIGIN_SERVER_TTL`, which controls the origin heartbeat expiration TTL and defaults to `300s`.

The proxy does not actively probe/check the origin server before picking it; it relies on the origin registration/heartbeat timestamp. For deployments that need faster or slower failover, please configure it, for example:

```bash
PROXY_ORIGIN_SERVER_TTL=60s
```

Because this is already supported, we don’t need to merge this PR. Closing.

PR #4658, opened 2026-04-11
Proxy: Fix HLS proxy response header loss and m3u8 URL query parameter corruption

sevico:
## Problem

Two bugs in the HLS proxy handler (`internal/protocol/http.go`):

### 1. Response headers lost during HLS proxying

`WriteHeader()` is called **before** copying backend response headers. In Go's `http.ResponseWriter`, headers set after `WriteHeader()` are silently ignored. This causes all backend headers (`Content-Type`, `Cache-Control`, etc.) to be dropped.

### 2. Double ampersand (`&&`) in m3u8 ts URL rewriting

When the original `.ts` URL already contains query parameters (e.g., `.ts?token=abc`), the proxy generates malformed URLs:

```
# Before (broken)
segment.ts?spbhid=xxx&&token=abc

# After (correct)  
segment.ts?spbhid=xxx&token=abc
```

## Fix

1. Move `WriteHeader()` after setting response headers (both streaming and HLS handlers).
2. Change `&&` to `&` in the m3u8 URL rewriting logic.

## Impact

- HLS playback may fail or behave unexpectedly when backend response headers are needed by the player.
- HLS playback fails when `.ts` segments have existing query parameters, because `&&` is not a valid URL separator.


suzp1984 (approved):
Approve

I added the UT to cover the committed code.

winlinvip:
Thank you for your contribution, this feedback is very valuable. Now AI manages this project, so AI will do the coding job, see #4718

PR #4659, opened 2026-04-11
Proxy: Fix resource leak in ParseBody. v8.0.26

sevico:
## Problem

`ParseBody` in `internal/utils/utils.go` defers `r.Close()` only after `io.ReadAll(r)` succeeds. If `io.ReadAll` returns an error, the function returns before registering the deferred close, so the request/body reader is left open.

```go
func ParseBody(r io.ReadCloser, v interface{}) error {
    b, err := io.ReadAll(r)
    if err != nil {
        return errors.Wrapf(err, "read body") // r is not closed on this path
    }
    defer r.Close()
```

## Solution

Move `defer r.Close()` to the top of `ParseBody`, before reading the body, so all return paths close the reader.

```go
func ParseBody(r io.ReadCloser, v interface{}) error {
    defer r.Close()

    b, err := io.ReadAll(r)
```


suzp1984:
I rebased your branch to develop branch. 
And this commit make sense.

suzp1984 (approved):

suzp1984:
One more thing, I added UT to cover this PR.

PR #4679, opened 2026-05-20
StateThreads: Set Non-Executable Stack. v8.0.21

atupone:
Inform the assembler that we don't need an executable stack


winlinvip:
Thanks for this PR. SRS works normally without this change, so this is not a functional bug fix. It is a Linux security-hardening improvement.

SRS includes hand-written State Threads assembly. Without a .note.GNU-stack declaration, the linker may assume that these assembly objects require an executable process stack. This can produce linker or packaging warnings and may cause the final SRS binary to have an executable stack.

The assembly code does not execute code from the stack, so marking it as non-executable is correct. This restores an important defense-in-depth mitigation: if SRS ever has a memory-corruption vulnerability, an attacker should not be allowed to execute injected code directly from the stack.

Therefore, this PR is worth merging because it:
* improves the security properties of the Linux binary;
* avoids executable-stack warnings from modern toolchains and distributions;
* accurately describes the requirements of the assembly code;
* has no expected effect on normal SRS behavior or performance.

This PR does not fix a known exploitable vulnerability, but it removes an unnecessary security weakness with a small and low-risk change.

Before merging, there is a warning when link SRS in Linux. It's fixed after this PR:

```bash
/usr/bin/ld: warning: md_linux.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
```

Before merging, the .note.GNU-stack directives should be moved outside the MD_ST_NO_ASM guards so no-ASM builds are also covered. It would also be useful to verify the final binary with:

```bash
readelf -W -l objs/srs | grep GNU_STACK
# Before
# GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RWE 0x10
# After
# GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
```

The expected result after the fix is RW, not RWE.

PR #4683, opened 2026-07-02
RTMP: Fix incorrect chunk stream id (fix for strict clients like Larix Broadcaster)

artem-smorodin-dacast:
## Problem

Larix Broadcaster for Android does not work with SRS v7.

I captured traffic dumps for comparison and found that the CIDs differ between v6 and v7 ([v6.txt](https://github.com/user-attachments/files/29591401/v6.txt) and [v7.txt](https://github.com/user-attachments/files/29591409/v7.txt))

Here is a comparison of CSIDs for v6 and v7

|Message                                 |v6 |v7 |
|----------------------------------------|---|---|
|Window Acknowledgement Size             |2  |3  |
|Set Peer Bandwidth                      |2  |3  |
|Set Chunk Size                          |2  |3  |
|_result('NetConnection.Connect.Success')|2  |3  |
|onBWDone()                              |3  |5  |

Based on [RTMP Specification](https://ossrs.net/lts/en-us/assets/files/rtmp_specification_1.0-25a467618b92a3115bc97d4b0038b0ff.pdf)

> 5. Protocol Control Messages
RTMP reserves message type IDs 1-7 for protocol control messages.
These messages contain information needed by the RTM Chunk Stream
protocol or RTMP itself. Protocol messages with IDs 1 & 2 are
reserved for usage with RTM Chunk Stream protocol. Protocol messages
with IDs 3-6 are reserved for usage of RTMP. Protocol message with ID
7 is used between edge server and origin server.
**Protocol control messages MUST have message stream ID 0 (called as
control stream) and chunk stream ID 2**, and are sent with highest
priority.
Each protocol control message type has a fixed-size payload. 

It seems that most clients tolerated this, but strict clients like Larix do not.

I think it happens after replacing `SrsSharedPtrMessage` with `SrsMediaPacket`

## Changes

- Restored `prefer_cid` propagation through RTMP message/header/media packet paths.
- Restored per-packet preferred CSID mapping for RTMP commands and protocol control packets.
- Updated fallback RTMP message-type to CSID mapping.
- Added/updated unit tests to verify expected chunk stream IDs.


winlinvip:
Thanks for your PR. See AI changes https://github.com/ossrs/srs/pull/4723

PR #4696, opened 2026-08-03
RTC2RTMP: Only emit video sequence header when parameter sets change. #4695

### LINKED ISSUE #4695: RTC2RTMP: HLS discontinuity before every segment
**Describe the bug**

For any WHIP-published stream consumed as HLS, the media playlist carries
`#EXT-X-DISCONTINUITY` before **every** segment — a decoder reset every `hls_fragment`
seconds. HLS players stutter or repeatedly drop the stream. Illustrative playlist shape
(every segment, not just the first):

```
#EXT-X-DISCONTINUITY
#EXTINF:2.000, no desc
livestream-0.ts
#EXT-X-DISCONTINUITY
#EXTINF:2.000, no desc
livestream-1.ts
#EXT-X-DISCONTINUITY
...
```

An RTMP publish (OBS or ffmpeg) of equivalent content through the same config shows one
initial discontinuity, then a clean playlist — the defect is specific to the RTC bridge.
In the SRS log the trigger is visible as a `34B video sh` write once per keyframe
interval (~1/s in our config) instead of once per publish.

Cause (traced in source, details under Additional context): libwebrtc repeats SPS/PPS
with every IDR, and the bridge re-emits an RTMP sequence header on every arrival; each
one makes the HLS muxer mark the open segment and stamp a discontinuity at close.

**Version**

- SRS 5.0.213 (tag v5.0-r3), runtime-verified.
- Also confirmed by code inspection on 7.0-develop (7.0.148, commit 16a24fd): the emit
  is still unconditional there (`SrsRtcFrameBuilder::do_packet_sequence_header_avc` /
  `_hevc`), so the defect is present on develop as well. (Develop was not run.)
- Config: `rtc_server` with WHIP ingest, `rtc_to_rtmp on;`, HLS enabled
  (`hls_fragment 2; hls_window 30;`), `mix_correct on;`.
- Publisher: libwebrtc via WHIP (flutter_webrtc 1.4.x, libwebrtc m137-era), H.264 + Opus.
  Any libwebrtc-based WHIP publisher is expected to reproduce (OBS 30+ WHIP, GStreamer
  `whipclientsink`), because the trigger is libwebrtc's standard STAP-A SPS/PPS repetition.

**To Reproduce**

1. Configure `rtc_server` + WHIP + `rtc_to_rtmp on;` + HLS as in Version above.
2. Publish from any libwebrtc WHIP client (OBS 30+ WHIP works).
3. After ~10 s, read the on-disk media playlist.
4. Every segment is preceded by `#EXT-X-DISCONTINUITY`. Repeat with an RTMP publish of
   equivalent content: only the initial discontinuity appears.

**Expected behavior**

The bridge should match the contract RTMP encoders follow: emit the sequence header once
per publish, and again only when the parameter sets actually change. The HLS playlist for
a WHIP publish should then look like the RTMP baseline — one initial discontinuity, none
per segment; a genuine parameter-set change (e.g. resolution switch) should still stamp
exactly one.

**Screenshots**

N/A — playlist excerpt and log signature inline above; full playlists/logs available on
request.

**Additional context**

*Root cause.* libwebrtc repeats SPS/PPS with every IDR (STAP-A). On every arrival the
bridge re-muxes and re-emits an RTMP AVC sequence header:

- 5.0: `SrsRtmpFromRtcBridge::packet_video_key_frame` (`srs_app_rtc_source.cpp`)
- develop: `SrsRtcFrameBuilder::do_packet_sequence_header_avc` (and the HEVC twin
  `do_packet_sequence_header_hevc`) — same shape after the refactor.

Each sequence header reaches `SrsHls`/`SrsHlsMuxer::on_sequence_header()`, which marks
the open segment, and `segment_close()` then stamps `#EXT-X-DISCONTINUITY` — once per
segment, forever. `reduce_sequence_header` only gates RTMP/FLV consumers; the hub→HLS
path has no dedup anywhere.

*Fix (PR attached).* Emit the sequence header only when the muxed avcC/hvcC bytes
actually changed: one `std::string last_sh_` member on the frame builder, compared before
emit, recorded only after a successful emit (so a failed emit is retried on the next
keyframe rather than remembered as delivered). The PR targets develop and patches both
the AVC and HEVC paths; the HEVC hunk is the same logic but untested end-to-end (no HEVC
WHIP publisher at hand).

*Verification.* Backport of the same fix onto 5.0.213, run on-device: WHIP camera publish
→ playlist with only the initial stream-start discontinuity, identical to the RTMP
baseline, confirmed in two independent soak runs (30-minute class, with mid-stream
mute/unmute phases). Regression: ffmpeg RTMP publish through the same config unchanged,
clean. Negative case: a mid-stream resolution change still emits a new sequence header
and stamps exactly one discontinuity. The develop branch of the fix applies cleanly and
compiles (ubuntu:focal, `./configure && make`).

*Existing-issue check (2026-07-20, GitHub search on ossrs/srs).* Queries:
`EXT-X-DISCONTINUITY`, `rtc_to_rtmp sequence header`, `WHIP HLS`, `discontinuity webrtc`,
`hls discontinuity`, `SPS PPS every`. No existing report of this defect. Nearest
neighbors, all different defects: #4590 (WHIP→HLS sometimes no video at all), #3191 (HLS
slice longer than configured, RTMP path), #4076 / #4418 (RTC2RTMP audio missing at
stream start — see our separate DTX/avsync report, linked below).

*AI-first note.* Per the getting-started-ai doc, this was diagnosed using an AI coding
agent (Claude Code) working directly against the SRS source tree; the analysis and fix
were then human-verified on-device as described above.

Related: #4697.

Varg3n:
Fixes #4695 — the WebRTC-to-RTMP bridge re-emits the video sequence header on every keyframe, so HLS stamps `#EXT-X-DISCONTINUITY` before every segment.

Emit the sequence header only when the muxed avcC/hvcC bytes actually change, and record it only after a successful emit so a failed emit is retried on the next keyframe. Covers both the AVC and HEVC paths of `SrsRtcFrameBuilder`.

Verified: a backport of the same fix onto 5.0.213 removes the per-segment discontinuity on-device (libwebrtc WHIP publish, two soak runs) with an RTMP publish through the same config unchanged as baseline; a mid-stream resolution change still stamps exactly one discontinuity. This branch applies cleanly to develop and compiles (ubuntu:focal, `./configure && make`). The HEVC hunk is the same logic but not runtime-tested (no HEVC WHIP publisher at hand).

### DIFF
--- a/trunk/src/app/srs_app_rtc_source.cpp
+++ b/trunk/src/app/srs_app_rtc_source.cpp
@@ -2131,6 +2131,15 @@ srs_error_t SrsRtcFrameBuilder::do_packet_sequence_header_avc(SrsRtpPacket *pkt,
         return srs_error_wrap(err, "mux sequence header");
     }
 
+    // WebRTC publishers (libwebrtc) repeat SPS/PPS with every IDR, so this is reached
+    // once per keyframe interval. Re-emitting an identical sequence header each time
+    // makes HLS mark the open segment (SrsHlsMuxer::on_sequence_header) and stamp
+    // #EXT-X-DISCONTINUITY before every segment. RTMP encoders send the sequence
+    // header once; match that contract and only emit when the bytes change.
+    if (sh == last_sh_) {
+        return err;
+    }
+
     // h264 packet to flv packet.
     char *flv = NULL;
     int nb_flv = 0;
@@ -2153,6 +2162,10 @@ srs_error_t SrsRtcFrameBuilder::do_packet_sequence_header_avc(SrsRtpPacket *pkt,
     if ((err = frame_target_->on_frame(&msg)) != srs_success) {
         return err;
     }
+
+    // Record only after a successful emit, so a failed emit is retried on the
+    // next keyframe instead of being remembered as delivered.
+    last_sh_ = sh;
 
     return err;
 }
@@ -2230,6 +2243,13 @@ srs_error_t SrsRtcFrameBuilder::do_packet_sequence_header_hevc(SrsRtpPacket *pkt
         return srs_error_wrap(err, "mux sequence header");
     }
 
+    // Same dedup as the AVC path: WebRTC publishers repeat parameter sets with
+    // every IDR; only emit the sequence header when the bytes change, else HLS
+    // stamps #EXT-X-DISCONTINUITY before every segment.
+    if (sh == last_sh_) {
+        return err;
+    }
+
     char *flv = NULL;
     int nb_flv = 0;
     if ((err = hevc->mux_hevc2flv_enhanced(sh, SrsVideoAvcFrameTypeKeyFrame, SrsVideoHEVCFrameTraitPacketTypeSequenceStart, pkt->get_avsync_time(),
@@ -2250,6 +2270,10 @@ srs_error_t SrsRtcFrameBuilder::do_packet_sequence_header_hevc(SrsRtpPacket *pkt
     if ((err = frame_target_->on_frame(&msg)) != srs_success) {
         return err;
     }
+
+    // Record only after a successful emit, so a failed emit is retried on the
+    // next keyframe instead of being remembered as delivered.
+    last_sh_ = sh;
 
     return err;
 }
--- a/trunk/src/app/srs_app_rtc_source.hpp
+++ b/trunk/src/app/srs_app_rtc_source.hpp
@@ -600,6 +600,14 @@ SRS_DECLARE_PRIVATE: // clang-format on
     SrsRtpPacket *obs_whip_sps_;
     SrsRtpPacket *obs_whip_pps_;
 
+// clang-format off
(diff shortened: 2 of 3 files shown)


Varg3n:
Pushed a fix for the CI utest failure: `RtcFrameBuilderTest.PacketSequenceHeaderAvc_ComprehensiveCoverage` asserted the pre-fix behavior (re-emitting the sequence header for every repeated SPS/PPS arrival), which is exactly the defect this PR removes (#4695).

The test now pins the new contract instead: identical repeated parameter sets are deduped (no re-emit), changed parameter sets emit again, and a failed emit is retried on the next arrival, because the header is only recorded as delivered after a successful emit.

Full utest suite is green locally (2352 tests, CI-mirrored build with sanitizer on). The coverage job failed from the same test, so it should recover as well.

Issue #4036, opened 2024-04-22
HTTPS: SSL errors in safari and chrome.


suzp1984:
**Describe the bug**
There are a similar issue #3497, but I think it just fix part of problem.

start srs: `./objs/srs -c conf/https.srs.conf`
access: `https://localhost:8088`

1. Safari browser access SRS https server.
SRS error logs
```
[2024-04-22 21:41:49.784][ERROR][9458][907848vt][0] serve error code=4043(HttpsRead)(Failed to read data from HTTPS stream) : parse message : parse message : grow buffer : read bytes : SSL_read r0=0, r1=6, r2=0, r3=1
thread [9458][907848vt]: process_requests() [./src/app/srs_app_http_conn.cpp:185][errno=0]
thread [9458][907848vt]: parse_message() [./src/protocol/srs_protocol_http_conn.cpp:103][errno=0]
thread [9458][907848vt]: parse_message_imp() [./src/protocol/srs_protocol_http_conn.cpp:153][errno=0]
thread [9458][907848vt]: grow() [./src/protocol/srs_protocol_stream.cpp:162][errno=0]
thread [9458][907848vt]: read() [./src/app/srs_app_conn.cpp:947][errno=0]
```

2. Chrome browser access SRS https server.
SRS error logs
```
[2024-04-22 21:43:25.057][INFO][9458][m46ggt66] TCP: before dispose resource(HttpsConn)(0x600002220000), conns=2, zombies=0, ign=0, inz=0, ind=0
[2024-04-22 21:43:25.057][ERROR][9458][m46ggt66][0] serve error code=4042(HttpsHandshake)(Failed to do handshake for HTTPS) : start : handshake : handshake r0=-1, r1=1
thread [9458][m46ggt66]: do_cycle() [./src/app/srs_app_http_conn.cpp:155][errno=0]
thread [9458][m46ggt66]: on_start() [./src/app/srs_app_http_conn.cpp:388][errno=0]
thread [9458][m46ggt66]: handshake() [./src/app/srs_app_conn.cpp:849][errno=0]
[2024-04-22 21:43:25.057][INFO][9458][c0k5sp99] TCP: clear zombies=1 resources, conns=2, removing=0, unsubs=0
[2024-04-22 21:43:25.057][INFO][9458][m46ggt66] TCP: disposing #0 resource(HttpsConn)(0x600002220000), conns=2, disposing=1, zombies=0
[...]

PR #4720, opened 2026-08-13
Backport: Support RTMPS server and graceful TLS shutdown. v6.0.192

luiscovelo:
## Summary

- Backport the server-side RTMPS support from #4443 to `6.0release`.
- Add the `rtmps` configuration section and the `SRS_RTMPS_*` environment overrides.
- Add a dedicated RTMPS listener and a TLS transport that completes the TLS handshake before the existing RTMP handshake.
- Fix `SrsSslConnection::read_fully()` to consume decrypted plaintext through the TLS read path.
- Include the follow-up fix from #4036 so a peer TLS `close_notify` is classified as a graceful client disconnect.
- Preserve plain RTMP behavior and add focused configuration and TLS read regression tests.

## Motivation

Native RTMPS server support was introduced in SRS 7.0 by #4443, but it is not available in the maintained `6.0release` branch. Deployments that must remain on SRS 6.0 therefore need external TLS termination to securely publish or play RTMP streams.

This backport provides the same inbound RTMPS capability on a dedicated port while avoiding unrelated SRS 7.0 refactors.

The original server implementation also exposed an existing TLS read behavior when an RTMPS peer closed the connection cleanly. OpenSSL reports an orderly TLS shutdown as `SSL_read()` returning zero with `SSL_ERROR_ZERO_RETURN`, but the SRS 6.0 TLS path classified that result as `ERROR_HTTPS_READ` (`4043`). As a result, a normal RTMPS publisher shutdown was logged as a server read error after the stream had already been unpublished and cleaned up.

This PR therefore also includes the focused follow-up from #4036 so the same peer-initiated TLS shutdown follows the existing graceful client-disconnect lifecycle.

## Implementation

- Introduce `SrsRtmpTransport` for plain RTMP and `SrsRtmpsTransport` for RTMP over TLS.
- Route both transports into the existing `SrsRtmpConn` and `SrsRtmpServer` lifecycle.
- Add `rtmps { enabled; listen; key; cert; }` configuration and the corresponding environment variables.
- Read key and certificate configuration when establishing the TLS connection.
- Make `SrsSslConnection::read_fully()` repeatedly call the virtual TLS `read()` method so the RTMP handshake receives decrypted bytes, including fragmented TLS records.
- Map `SSL_ERROR_ZERO_RETURN` to `ERROR_SOCKET_READ`, matching the current SRS 7.0/develop behavior while preserving other TLS read failures.
- Add `conf/rtmps.conf`, update `full.conf`, and bump SRS 6.0 to `v6.0.192`.

## Graceful TLS shutdown

When a peer sends the TLS `close_notify` alert, OpenSSL returns `SSL_ERROR_ZERO_RETURN`. The TLS read path now maps only that condition to `ERROR_SOCKET_READ`, which is already recognized by SRS as a graceful client disconnect.

This changes the final classification and log from an `ERROR_HTTPS_READ` (`4043`) service error to the normal `client disconnect peer. ret=1007` path. The existing unpublish, resource disposal, and connection cleanup lifecycle remains unchanged.

Other TLS failures continue to return `ERROR_HTTPS_READ`; the change does not hide handshake failures, malformed TLS traffic, transport failures, or unexpected read errors. It also does not add a server-initiated `SSL_shutdown()` sequence—the behavior applies specifically to an orderly shutdown initiated by the peer.

## Scope

This PR is intentionally limited to inbound RTMPS server support on a dedicated listener. It does not include:

- RTMP and RTMPS auto-detection on the same port;
- outbound `rtmps://` support for forward, edge, or ingest clients;
- per-vhost certificates, SNI, or mTLS;
- the global `ERROR_HTTPS_*` to `ERROR_TLS_*` rename from SRS 7.0;
- unrelated HTTP TLS client or IPv6 refactors.

## Verification

- `make -j2`
- `make utest -j2`
- `./objs/srs_utest --gtest_filter='ConfigRtmpsTest.*:TlsConnectionTest.*'` — 4 tests passed
- `./objs/srs_utest --gtest_filter='-TCPServerTest.TCPListen:TCPServerTest.TCPClientServer'` — 591 tests passed
- `./objs/srs -v` — `6.0.192`
- OpenSSL TLS 1.2 handshake against the RTMPS listener passed.
- FFmpeg publish/play passed over RTMPS, including H.264/AAC decoding.
- Plain RTMP publish/play passed as a regression check.
- Plain RTMP sent to the RTMPS port and TLS sent to the RTMP port were rejected without affecting server health.
- A missing TLS private key failed the connection cleanly and left the server healthy.
- FFmpeg RTMPS shutdown produced `client disconnect peer. ret=1007` without `serve error code=4043`.
- `git diff --check`

## AI co-authorship

This backport was co-authored with OpenAI Codex.

- Model: `gpt-5.6-sol`
- Reasoning effort: `xhigh`
- Scope of assistance: upstream implementation analysis, SRS 6.0 adaptation, TLS read-path correction, tests, end-to-end validation, and PR preparation.

The contributor reviewed the resulting diff and executed the validation listed above.

## Links

- Original RTMPS PR: https://github.com/ossrs/srs/pull/4443
- Original implementation commit: https://github.com/ossrs/srs/commit/ebcaef43c6162753453e4d74cf1eb73ceda3b26f
- Graceful TLS shutdown issue: https://github.com/ossrs/srs/issues/4036
- Graceful TLS shutdown fix: https://github.com/ossrs/srs/commit/19b603a0d722cb84af95e82f9dfb0b9c23437d73


winlinvip:
No feature for SRS 6, for it's released version.

luiscovelo:
@winlinvip No problems. 

A question, for released version just fixes will be accepted?

PR #4736, opened 2026-09-09
HTTP: Return HTTP status for rejected playback hooks

nguyencobap:
Currently, SRS does not propagate an `on_play` hook's error to the HTTP status sent to the HTTP-FLV client. For example, a hook returning HTTP 200 with `{"code":401}` does not result in HTTP 401 for the client.

This change makes SRS return the corresponding HTTP error status to the client: hook HTTP 4xx/5xx statuses are preserved, and JSON `code` values in the 400-599 range from HTTP 200/201 hook responses become the client's HTTP status. Other hook failures return HTTP 500. In the example above, the client now receives HTTP 401.

Validation: all 2,381 C++ unit tests and real HTTP-FLV playback checks pass. The edge test fails on FFmpeg's log format; WHIP media testing is blocked by a missing muxer.

### DIFF
--- a/skills/internal-docs-for-srs/references/cpp-docs/doc/http-callback.md
+++ b/skills/internal-docs-for-srs/references/cpp-docs/doc/http-callback.md
@@ -425,6 +425,22 @@ Content-Length: 11
 {"code": 0}
 ```
 
+### HTTP Live Playback Errors
+
+When an `on_play` callback rejects HTTP live playback (HTTP-FLV, HTTP-TS, AAC, or
+MP3 through `http_remux`), SRS sends an HTTP error response to the player:
+
+* A callback HTTP status in the `400`-`599` range is passed through.
+* For callback HTTP `200` or `201`, an integer JSON `code` in the `400`-`599`
+  range becomes the playback HTTP status. For example, `{"code":401}` returns
+  HTTP `401 Unauthorized` to the player.
+* Other application error codes, invalid or empty responses, and transport
+  failures return HTTP `500 Internal Server Error`.
+
+The response body contains only the standard HTTP status text. Callback URLs,
+request parameters, and backend error messages are not included. Successful
+callbacks and rejection handling for other playback protocols are unchanged.
+
 You could run the example HTTP callback server by:
 
 ```
--- a/trunk/src/app/srs_app_http_hooks.cpp
+++ b/trunk/src/app/srs_app_http_hooks.cpp
@@ -45,6 +45,14 @@ ISrsHttpHooks::~ISrsHttpHooks()
 {
 }
 
+srs_error_t ISrsHttpHooks::on_play(string url, ISrsRequest *req, int *http_status)
+{
+    if (http_status) {
+        *http_status = SRS_CONSTS_HTTP_InternalServerError;
+    }
+    return on_play(url, req);
+}
+
 SrsHttpHooks::SrsHttpHooks()
 {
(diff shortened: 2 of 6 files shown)


winlinvip:
## Cross-protocol note: `on_play` hook rejection status is not consistently surfaced to the player

**Background**

All SRS playback protocols call the same `on_play` HTTP hook to authorize a play request (`ISrsHttpHooks::on_play` in `trunk/src/app/srs_app_http_hooks.cpp`). Six call sites exist in `trunk/src/app/`: RTMP (`srs_app_rtmp_conn.cpp:604`), HTTP-FLV/TS/AAC/MP3 via `http_remux` (`srs_app_http_stream.cpp:714`), HLS (`srs_app_http_static.cpp:181`), WebRTC WHEP/legacy play API (`srs_app_rtc_api.cpp:269`), RTSP `DESCRIBE` (`srs_app_rtsp_conn.cpp:759`), and SRT/MPEG-TS play (`srs_app_srt_conn.cpp:443`). GB28181 is publish-only and never calls it; MPEG-DASH has no `on_play` call site in this codebase.

The hook rejection itself is universal — a non-zero `code` always blocks playback on every protocol. This PR fixes one specific gap for HTTP-FLV: today, a rejected hook just drops the TCP connection with no HTTP response at all; the PR makes `serve_http_impl` call `srs_go_http_error(w, http_status)` so the client gets a real status line (e.g. `401 Unauthorized`) derived from the hook's HTTP status or JSON `code`.

**Current state**

| Protocol | Current behavior on hook rejection |
|---|---|
| RTMP | Silent TCP close — worse, `start_play()` already sent `NetStream.Play.Start` (success) before the hook ran (`srs_protocol_rtmp_stack.cpp:2588-2601`). |
| HLS (`.m3u8`) | Silent connection close, no HTTP response — same bug this PR fixes for HTTP-FLV, but not touched here. |
| SRT (MPEG-TS play) | Silent connection close; SRT/MPEG-TS has no application-level status channel to carry a reason at all. |
| WebRTC WHEP / `/rtc/v1/play/` | `SrsGoApiRtcPlay::serve_http` (`srs_app_rtc_api.cpp:64-77`) responds, but always wire-level `HTTP 200` with JSON body `{"code":400}` (`srs_app_http_api.cpp:108-115` never calls `write_header`) — the hook's actual status/code (401, 403, ...) is discarded and collapsed to a fixed generic value. |
| RTSP `DESCRIBE` | Real RTSP status is returned, but the error-code switch (`srs_app_rtsp_conn.cpp:618-627`) only recognizes `ERROR_RTSP_NO_TRACK`→404 and `ERROR_SYSTEM_SECURITY_DENY`→403; a hook rejection matches neither, so it always becomes RTSP `500`, discarding the hook's actual code. |
| HTTP-FLV/TS/AAC/MP3 | Today: silent close, same as HLS. After this PR merges: faithful HTTP status forwarded. |

**Conclusion**

This is a real, confirmed gap, not a misunderstanding of intended behavior: five of six playback protocols either give the client no reason at all, or a coarse generic one, when `on_play` rejects them — nothing in the codebase suggests this was an intentional per-protocol design choice; each protocol's hook integration appears to have been implemented independently over time, without a shared "propagate hook status" mechanism. This PR establishes the pattern for HTTP-FLV only; HLS is the closest sibling and would be the cheapest follow-up if the maintainers want to extend this fix further.

winlinvip:
Fixed by AI, see #4737

